PR #12, opened 2021-07-13
Add function to calculate per cell QC with mitochondrial fraction

allyhawkins:
This PR adds in a function to use scater's `addPerCellQC()` function, adding in the calculation of mitochondrial fraction as part of the function and allowing for the option to pass other arguments to the function (such as the threshold option used in benchmarking). 

In testing this, `scater::addPerCellQC()` will still output a table with QC metrics in `colData` even if the subset that is input is either empty or not found in the input `SingleCellExperiment`. If mito is null, then it assumes that all reads are in the subset. I have it failing without the mito input list to avoid inaccurate mito outputs and without the mito list of genes you  wouldn't need to use this function. 

If mito genes are input, but not found in the rownames, then it shows 0's for all data in the mito columns. I also have implemented a warning message to state that the genes have not been found, but it should still calculate QC metrics.


jashapiro (approved):
LGTM!

PR #13, opened 2021-07-14
Add tests for import functions

### LINKED ISSUE #8: Add tests for import functions
Right now, the functions involved in importing quants data do not have associated test functions. So far only `collapse_intron_counts.R` has a test function. 

We should add some test functions for `import_quant_data.R` and related functions (`read_alevin.R`, `read_kallisto.R`, `read_cellranger.R` and `filter_counts.R`).

jashapiro:
This PR adds tests for each of the single cell processor format read functions. I did not include a test for `import_quant_data()` partly because that test would be largely redundant with these tests, which would double the time it takes to run the tests. In general, I think that function will fail if given bad arguments, but we might want to look into it more deeply at some point (I do not feel that every single function needs to be tested every possible way). 

To run these tests, you must have the `scpcaData` package **installed** (it does not need to be loaded), which is most easily done by checking out [the repo](https://github.com/AlexsLemonade/scpcaData), opening the project within it in RStudio and running `Build> Install and Restart`. Until https://github.com/AlexsLemonade/scpcaData/pull/2 is merged, you will need to do the latter step from the [jashapiro/add-bc-data branch](https://github.com/AlexsLemonade/scpcaData/tree/jashapiro/add-bc-data). Once we make the `scpcaData` package public, we should be able to install it with `remotes::install_github("AlexsLemonade/scpcaData")` (Technically, we could install from a private repo, but that is a pain.)

Closes #8


tests/testthat/test-read_alevin.R
```diff
@@ -0,0 +1,30 @@
...
+
+test_that("Check that reading intron mode works", {
+  sce <- read_alevin(intron_dir,
+                     intron_mode = TRUE,
+                     which_counts = "unspliced")
+  expect_s4_class(sce, "SingleCellExperiment")
+  expect_equal(dim(sce), sce_size)
+  # check that column names are barcodes
+  col_barcode <- stringr::str_detect(colnames(sce), "^[ACGT]+$")
+  expect_true(all(col_barcode))
```
allyhawkins, line 28:
```suggestion
  expect_true(all(col_barcode))
  gene_names <- str_detect(rownames(sce), "-[IUA]$")
  expect_false(all(gene_names))
```
If you like the check above, you could add it in here too.

tests/testthat/test-read_cellranger.R
```diff
@@ -0,0 +1,14 @@
...
+
+# expected matrix size
+sce_size <- c(19970, 687)
+
+test_that("Check that cellranger import works", {
+  sce <- read_cellranger(cellranger_dir)
+  expect_s4_class(sce, "SingleCellExperiment")
+  expect_equal(dim(sce), sce_size)
+  # check that column names are barcodes
+  col_barcode <- stringr::str_detect(colnames(sce), "^[ACGT]+$")
```
allyhawkins, line 12:
```suggestion
  col_barcode <- str_detect(colnames(sce), "^[ACGT]+$")
```

tests/testthat/test-read_kallisto.R
```diff
@@ -0,0 +1,14 @@
...
+
+# expected matrix size
+sce_size <- c(60666, 30339)
+
+test_that("Check that kallisto import works", {
+  sce <- read_kallisto(kallisto_dir)
+  expect_s4_class(sce, "SingleCellExperiment")
+  expect_equal(dim(sce), sce_size)
+  # check that column names are barcodes
+  col_barcode <- stringr::str_detect(colnames(sce), "^[ACGT]+$")
```
allyhawkins, line 12:
```suggestion
  col_barcode <- str_detect(colnames(sce), "^[ACGT]+$")
```

allyhawkins (approved):
Overall I think these look great! I added in one suggestion to the tests for `read_alevin` since you are also testing the use of `usa_mode` and `intron_mode`, it might be worth just adding a check to make sure the gene names don't contain the additional -U,A,I anymore. I know we have the separate `collapse_intron_counts` test, so it's not 100 percent necessary, but it wouldn't add much time to the test. 

The other minor thing was just about removing the `stringr::` from using `stringr::str_detect` since we use `@import` for that package. 

A thought for a future issue and outside the scope of this PR is do we want to include a test for the filtering function? That might be one reason we want to either add a test for `import_quant_data` to test use of that function, or we can create a test specifically for that function.

tests/testthat/test-read_alevin.R
```diff
@@ -0,0 +1,30 @@
...
+
+test_that("Check that reading USA mode works", {
+  sce <- read_alevin(usa_dir,
+                     usa_mode = TRUE,
+                    which_counts = "spliced")
+  expect_s4_class(sce, "SingleCellExperiment")
+  expect_equal(dim(sce), sce_size)
+  # check that column names are barcodes
+  col_barcode <- stringr::str_detect(colnames(sce), "^[ACGT]+$")
+  expect_true(all(col_barcode))
```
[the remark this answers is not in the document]
jashapiro: This is a good idea, but I am going to implement it slightly differently (for variety?).
```suggestion
  expect_true(all(col_barcode))
  unmerged_genes <- str_subset(rownames(sce), "-[IUA]$")
  expect_length(unmerged_genes, 0)
```

tests/testthat/test-read_alevin.R
jashapiro: Same idea: 
```suggestion
  expect_true(all(col_barcode))
  unmerged_genes <- str_subset(rownames(sce), "-[IUA]$")
  expect_length(unmerged_genes, 0)
```

jashapiro:
> A thought for a future issue and outside the scope of this PR is do we want to include a test for the filtering function? That might be one reason we want to either add a test for `import_quant_data` to test use of that function, or we can create a test specifically for that function.

We definitely want a test for that function! That is part of the motivation I have for #15. I don't want that test to be dependent on reading real data.

Testing `import_quant_data` is probably a good idea too, but it gets a bit redundant with the import functions, so it was not a priority for me.

PR #14, opened 2021-07-15
Add tests for add_cell_mito_qc

jashapiro:
I chose to make these tests mostly to test out making a random `SingleCellExperiment` for testing. In making it though, I decided to make a little change to the `add_cell_mito_qc()` function, change the column names that `addPerCellQC()` for its defaults. So now the function create columns named `mito_*` rather than `subsets_mito_*` which just seems nicer.

One question for reviewers: Should I make the chunk of code that generates a random SCE its own function? I imagine we might use that a few places, so it probably makes sense. Maybe in the next test PR though?


allyhawkins (approved):
All this looks good to me! I definitely think there is a case for pushing the piece that generates a random sce out to its own function, but we can save that for the next time it needs to be used.

Issue #7, opened 2021-07-02
Add function to generate QC report


allyhawkins:
Along the same lines as https://github.com/AlexsLemonade/alsf-scpca/issues/43, we should add in a function here to generate the QC report. Exactly what tools will be used to generate QC metrics still need to be figured out (ie. are we going to use miQC or not). Writing this down now since this is something we have talked about and I think would be useful to integrate with the current `import_quant_data` functions so we can get a QC report (and maybe even a dataframe with metrics?) for each sample.

jashapiro:
I am working on a basic version of this now, which will do the basic part of being a function that turns a template notebook into an html qc report with a SingleCellExperiment (and other info?) as input. After that is done, we will want to think about what components should go into the report. 

Thoughts at this point (partially inspired by AlevinQC & CR's `web_output.html`):
- Processing info 
  - Software, versions, etc.
  - can we attach this to SCE's as metadata? If so, we might well want to make that part of scpcaTools read functions, etc.
- Number of cells
- UMIs per cell stats (mean, median, sd?)
- Genes per cell
- Genes detected
- Knee plot with included/excluded cell highlighted (this might mean require 2 sce inputs: filtered & unfiltered, or at least unfiltered + the filtered cell list)
- Distribution of mito %
- Scatter plot of cells with UMI & expressed genes like we use in training?
- Maybe a UMAP plot & clusters (filtered cells only)?

Most of these parts can be broken out into separate issues, if we agree that they should be included. @jaclyn-taroni suggested that many of the plots might be made interactive. I am not opposed to that idea, but we will have to pick which interactive tool we use! I just learned about [`plotly::ggplotly()`](https://plotly.com/ggplot2/), which might make this easy (and easy to add later if we decide not to use it for initial versions!).

allyhawkins:
I think all of this looks good in terms of what to include. For the stats for UMI/cell and genes per cell, I like the idea of  making both a table and a plot for this. The scatter plot we make in training could be sufficient, but it also could be helpful to show a trio of plots with the distribution of UMI/cell, genes/cell, and mito content post filtering. I think it will depend on which one looks better visually. 

My only concern with the knee plot is for the single-nuclei samples. We know that they don't perform well with the knee filtering so I'm not sure how this would look and do we want to include this if we aren't using a knee plot for filtering?

PR #16, opened 2021-07-16
Add a markdown QC report function & skeleton

jashapiro:
Related to https://github.com/AlexsLemonade/scpcaTools/issues/7

This PR adds a function and template file to generate a QC report about a SingleCellExperiment, placing that report wherever you like. 

The template lives at `inst/rmd/qc_report.Rmd` and is a parameterized rmarkdown file. For now it is very basic (it only reports the number of cells and genes), but it should serve as a proof of concept. 

The `generate_qc_report()` function takes a `SingleCellExperiment` and a sample name, passes it to the report and renders to the specified location. 

### Future thoughts: 
I suspect we will want to do most of the QC processing outside of the rmarkdown render, so that things like the mito percent can just be plotted, rather than recalculated.

For future expansion, we may well want to take 2 SCE objects, pre and post filtering, but I figure that can come as we expand the ideas that are now present in https://github.com/AlexsLemonade/scpcaTools/issues/7


R/generate_qc_report.R
```diff
@@ -0,0 +1,44 @@
+#' Generate a QC report from a SingleCellExperiment object
+#'
+#' @param sce A SingleCellExperiment object that the report will describe
+#' @param sample_name The name of the sample
+#' @param outfile The output file path that will be created.
+#'   If the file name does not include an extension, ".html" will be added automatically.
+#'   Any directories in the path will be created as needed.
```
jaclyn-taroni, line 7: Assuming that this is `rmarkdown::render()` behavior?

inst/rmd/qc_report.rmd
```diff
@@ -0,0 +1,32 @@
+---
+params:
+  sample: Example
+  sce: !r SingleCellExperiment::SingleCellExperiment()
+  date: !r Sys.Date()
+
+title: "`r glue::glue('ScPCA QC report for {params$sample}')`"
+author: "CCDL"
+date: "`r params$date`"
```
jaclyn-taroni, line 9: Why not `Sys.Date()` here? I can see the argument for what you've implemented but just wondering what you are potentially future-proofing for...

inst/rmd/qc_report.rmd
```diff
@@ -0,0 +1,32 @@
+---
+params:
+  sample: Example
+  sce: !r SingleCellExperiment::SingleCellExperiment()
+  date: !r Sys.Date()
```
jaclyn-taroni, line 5: So `!r` means these are R expressions I take it.

jaclyn-taroni (approved):
Looks like a great start! Had a couple questions for my understanding.

R/generate_qc_report.R
jashapiro: It is. The directory stuff is strange, and what I was trying to solve for. `render()` creates directories iff you provide the `output_dir` argument. If you provide only an `output_file` argument, then it will fail if the path does not exist. But if you provide `output_file` without a path (just a file), it puts the output where the `.rmd` file was, which would be bad here, since that is wherever the package was installed. 

The file extension stuff I thought was neat to just leave alone, in the off chance we want to create pdfs or word docs. You can make it make a pdf by just using `"foo.pdf"` as your `output_file` even if the `.rmd` indicates it is for an html file.

inst/rmd/qc_report.rmd
jashapiro: I think that was me playing around a bit. Also solving for the case when we want to forge report dates.

inst/rmd/qc_report.rmd
jashapiro: Yep. Seems to be required for params, which is different from the other YAML headers. Maybe. I actually don't remember if I tested all combinations. One thing that is true is that `::` notation is required for params, as they are executed before any libraries are loaded.

PR #22, opened 2021-07-20
Add Function to create a test SCE object

jashapiro:
This PR adds a (non-exported) `sim_sce()` function that creates a `SingleCellExperiment`. I was inspired a bit by `DropletUtils::simCounts()`, but went a bit simpler (less realistic?) with the simulation. However, I added a different feature, which is that the cells should cluster into the specified number of groups, which may be nice for some tests down the road. 

Other than that, it should be pretty straightforward. I converted the test of `add_cell_mito_qc()` to use the function.

In checking I also found a documentation mismatch in generate_qc_report, which I fixed.


R/sim_sce.R
```diff
@@ -0,0 +1,58 @@
...
+    warning("Did you really want to simulate more than a million droplets?")
+  }
+
+  # generate a random distribution of gene expression for each group
+  gene_expr <- rexp(n_genes * n_groups, 0.2)
+  # expression for each gene x cell
+  cell_expr <- matrix(rpois(n_genes * n_cells, lambda = gene_expr), nrow = n_genes)
+
+  # empty drops will have expression calculated from the overall mean
+  empty_expr <- rowMeans(cell_expr) * 0.01
```
allyhawkins, line 40: This is just so I understand, but why can't we just make a vector of zeros here?

allyhawkins (approved):
I think this looks good to me! I just had one question about generating the random gene expression for the empty droplets for my own understanding.

R/sim_sce.R
jashapiro: I wanted it to have a whiff of realism? The idea is that there are droplets with a few reads from the soup, so the expression counts will be based on the average expression of the "real" cells, just much lower. I actually checked whether EmptyDrops would toss these cells, and it does... probably mostly because they have lower expression, but it keeps the "real" cells, so that seemed good enough for now.

allyhawkins: Ah that makes a lot of sense. I had forgotten that empty drops actually do have some reads in them sometimes.

PR #23, opened 2021-07-21
Add alevin (no fry) test

jashapiro:
For completeness, we should have a test for base alevin. Now that such data is in the updated scpcaData, this test should run.

Also fixed a small documentation error that was throwing a check warning.


allyhawkins (approved):
LGTM!

PR #25, opened 2021-07-23
Support new alevin json files

jashapiro:
This PR adds support for alevin-fry 0.4.1, which (at my request) moved the json file for `alevin-fry quant` to `quant.json`. I am maintaining support for the old version for now, so we look for `meta_info.json` as backup, but we may be able to remove that in the future.

### DIFF
--- a/R/read_alevin.R
+++ b/R/read_alevin.R
@@ -115,14 +115,18 @@ read_usa_mode <- function(quant_dir,
     stop(paste0("Missing Alevin output file(s): ", missing_files))
   }
 
-  meta_json_path <- file.path(quant_dir, "meta_info.json")
-  if(!file.exists(meta_json_path)){
-    stop("Missing meta_info.json in Alevin output directory")
+  quant_json_path <- file.path(quant_dir, "quant.json")
+  if(!file.exists(quant_json_path)){
+    # file for alevin-fry < 0.4.1
+    quant_json_path <- file.path(quant_dir, "meta_info.json")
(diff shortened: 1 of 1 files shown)


R/read_alevin.R
```diff
@@ -115,14 +115,18 @@ read_usa_mode <- function(quant_dir,
...
+    if(!file.exists(quant_json_path)){
+      stop("Missing quant.json (or meta_info.json) in Alevin output directory")
+    }
   }
 
   # check that USA mode is true in JSON file
-  meta_json <- jsonlite::fromJSON(meta_json_path)
-  if(meta_json$usa_mode != "TRUE"){
+  quant_json <- jsonlite::fromJSON(quant_json_path)
+  if(quant_json$usa_mode != "TRUE"){
```
jaclyn-taroni, line 129: Checking my understanding: before alevin 0.4.1 the `meta_info.json` indicated that something was run in USA mode and now that information is in `quant.json`.

jaclyn-taroni (approved):
LGTM 🚀  I am assuming that my understanding is correct with this approval!

R/read_alevin.R
jashapiro: yup! They changed the filename to avoid potential conflict (salmon already had a `meta_info.json` file in a subdirectory) and to make it the name match the pattern of the  `.json` files the other `alevin-fry` commands used. They also added some fields, but that's it.

Issue #17, opened 2021-07-16
Add processing information to QC report 


allyhawkins:
As part of the QC report, we should include information about processing including software information and versions. This may involve modifying the scpcaTools read functions and attaching this information to the SCE metadata and then grabbing it from the SCE to output during the report.

jashapiro:
With the current version of `salmon` (1.5.1) some of the expected files are not being produced (see https://github.com/COMBINE-lab/salmon/issues/688), which may limit our ability to use the processor file output for tracking this information. We will likely want to have some additional way of tracking the command invocation for steps of our workflow with the output files. This may involve additions to https://github.com/AlexsLemonade/scpca-nf, but we should decide here what fields we would like to include. 

We should make this flexible so missing information does not result in failure. 

Some initial thoughts:
-`pipeline`: probably a reference to https://github.com/AlexsLemonade/scpca-nf most of the time, with either a release tag or commit id
- `reference` Ensembl 103, etc.
- `index`
- `filter` (emptydrops etc.)

What else would people want to see?

PR #26, opened 2021-07-28
Add reading alevin metadata (and refactor)

jashapiro:
The main point of this PR was to add reading some metadata from alevin and alein-fry outputs and incorporate that data into the SingleCellExperiment `metadata` slot. While I was waiting for the output files to be made more consistent, I ended up doing a bit of refactoring as well. 

The biggest change on the refactoring front was moving the `tximport` reads to a separate function in order to simplify the main  `read_alevin()` function. I also ended up renaming `read_usa_mode()` to reflect that we might use it in the future for other tasks (namely CITE-seq).

The metadata reading happens in a separate function as well, for cleanliness.

### Questions for reviewers
I'd appreciate a close look at the metadata fields I chose to include. Are there other fields that seem worth including?

We may want to rename our index files to make the origin of the data (ensembl release) a bit more clear, but that is outside the scope of this PR/issue.

There is one choice I made that I am not fully comfortable with, which was to add the `transcript_type` only upon collapsing calls (so that lives in the `read_alevin()` function instead of `read_alevin_metadata()`). I can see that we might want to handle that differently, but I wasn't sure the best way.

related to #17


R/read_alevin.R
```diff
@@ -115,29 +106,101 @@ read_usa_mode <- function(quant_dir,
...
+
+#' Read alevin metadata from json files
+#'
+#' @param quant_dir Path alevin output directory.
+#'
+#' @return A list containing alevin run metadata,
+#'   with NULL values for missing elements.
+#'
+#' @noRd
+read_alevin_metadata <- function(quant_dir){
```
allyhawkins, line 152:
```suggestion
read_alevin_metadata <- function(quant_dir, which_counts = c("spliced", "unspliced"){
```
If we want to add transcript type during this part why don't we include it as an argument here and then append it later?

R/read_alevin.R
```diff
@@ -115,29 +106,101 @@ read_usa_mode <- function(quant_dir,
...
+  } else {
+    meta$mapping_tool <- "alevin-fry"
+  }
+
+  # Add other metadata
+  # assume all alevin-fry tool versions are the same
+  meta$alevinfry_version <- permit_info[['version_str']]
+  meta$af_permit_type <- permit_info[['permit-list-type']]
+  meta$af_resolution <- quant_info[['resolution_strategy']]
+  meta$usa_mode <- quant_info[['usa_mode']]
```
allyhawkins, line 202:
```suggestion
  meta$usa_mode <- quant_info[['usa_mode']]
  meta$transcript_type <- which_counts
```
If you like my earlier suggestion of including the transcript type in this function.

tests/testthat/test-read_alevin.R
```diff
@@ -15,17 +15,23 @@ test_that("reading salmon alevin data works", {
...
                      usa_mode = TRUE,
-                    which_counts = "spliced")
+                     which_counts = "spliced")
   expect_s4_class(sce, "SingleCellExperiment")
   expect_equal(dim(sce), sce_af_size)
   # check that column names are barcodes
   col_barcode <- str_detect(colnames(sce), "^[ACGT]+$")
   expect_true(all(col_barcode))
+  # check metadata
+  expect_equal(sce@metadata$mapping_tool, "alevin-fry")
```
allyhawkins, line 32:
```suggestion
  expect_false(is.na(sce@metadata$salmon_version))
  expect_false(is.na(sce@metadata$reference_index))
  expect_equal(sce@metadata$mapping_tool, "alevin-fry")
```
Probably a good idea to add checks that the salmon version and reference index aren't empty to ensure the `cmd_info.json` was read in properly.

R/read_alevin.R
```diff
@@ -115,29 +106,101 @@ read_usa_mode <- function(quant_dir,
...
+  # Read other info files if they exist. Otherwise, create dummy values
+  if (file.exists(permit_json_path)){
+    permit_info <- jsonlite::read_json(permit_json_path)
+  } else {
+    permit_info <- list()
+  }
+  if (file.exists(quant_json_path)){
+    collate_info <- jsonlite::read_json(collate_json_path)
+  } else {
+    collate_info <- list()
```
allyhawkins, line 178:
```suggestion

```

Based on my comment above, we could remove this.

allyhawkins (approved):
I think these are generally good changes. To answer your question about what should we include or not include in the metadata, I think you got most of everything we would want to capture. I added a suggestion about adding in the tx2gene file name for if we are using alevin since that contains the version information and also could be beneficial to have. The other thing I thought of that I'm going back and forth on is the number of cells along with general metrics like mean/ median umis per cell and genes per cell. Both alevin and alevin-fry output the number of cells that are considered and alevin has a `alevin_meta_info.json` file that outputs the mean umis per cell and mean genes per cell. I know we might calculate some of these in the QC report and I'm assuming that we will add them on to the sce at that point? But is it worth grabbing this information from the alevin report here? I could see an argument for both sides here. 

The other question you had was about adding the `transcript_type` within the `read_alevin` function. This may only be a half solution, but one possibility is to use `which_counts` as input to the `read_alevin_metadata` function and then add it in the function. I'm not sure it makes a huge difference but if you want to add it there instead? Right now, `which_counts` is a required input and I think it would be helpful to have that in all metadata where you currently have it for only those that are in usa or intron mode.

R/read_alevin.R
```diff
@@ -115,29 +106,101 @@ read_usa_mode <- function(quant_dir,
...
+#' @param quant_dir Path alevin output directory.
+#'
+#' @return A list containing alevin run metadata,
+#'   with NULL values for missing elements.
+#'
+#' @noRd
+read_alevin_metadata <- function(quant_dir){
+  cmd_info_path <- file.path(quant_dir, "cmd_info.json")
+  permit_json_path <- file.path(quant_dir, "generate_permit_list.json")
+  collate_json_path <- file.path(quant_dir, "collate.json")
```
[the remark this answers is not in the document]
jashapiro: Yeah, I guess we never use it...

R/read_alevin.R
jashapiro: I thought about doing this (it was there in a pervious version) but I think my logic was that this function is (by its name) about reading metadata, so anything not read in could be added separately.

R/read_alevin.R
```diff
@@ -115,29 +106,101 @@ read_usa_mode <- function(quant_dir,
...
+    meta$mapping_tool <- "alevin-fry"
+  }
+
+  # Add other metadata
+  # assume all alevin-fry tool versions are the same
+  meta$alevinfry_version <- permit_info[['version_str']]
+  meta$af_permit_type <- permit_info[['permit-list-type']]
+  meta$af_resolution <- quant_info[['resolution_strategy']]
+  meta$usa_mode <- quant_info[['usa_mode']]
+
```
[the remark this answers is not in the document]
jashapiro: I like the idea of this, but for `--rad` mode this might not actually be there (the argument is no longer required, as the file is unused). There should be an equivalent in the `quant.json` file, but there isn't a specific field, so we would have to parse it out of the command line call stored in `"cmd"`. I don't love that.

I think we probably want to explore other ways of passing along this information along the workflow... not sure the best way as of now.

tests/testthat/test-read_alevin.R
jashapiro: yup, good idea... made me realize the line after is wrong though: should be `is.null`!

jashapiro: Oh, actually, I know why I wasn't testing the reference index: it would fail with some versions where that data wasn't reported properly... so null values there are possible. (this was also why I wasn't testing for salmon version probably. But that is there (if sometimes wrong) in all cases.

PR #29, opened 2021-08-05
Add processing information to QC report

### LINKED ISSUE #17: Add processing information to QC report

allyhawkins:
Closes #17. This PR adds in basic processing information to the beginning of the QC report using `generate_qc_report()`. As it is right now, I grab any metadata available from the sce as a data frame, remove any empty rows, and then print it out in a basic table. Because we have some types of tools that currently do not have metadata in them, I have also allowed for the possibility of empty metadata and it will print out a statement that no processing information is available. 

What I have not done here is modify any of the row names from the original names in the metadata table or any of the values that are grabbed directly from the metadata which originates from the json files in alevin-fry. I wasn't sure if we also wanted to include all of the information that is present in the metadata. I was hoping to get some initial feedback first in terms of how much we should change from the original metadata table and if we want to do the changing here or in the original `read_alevin_metadata` function instead to allow this to be more flexible with multiple tools. Although I am assuming at this point we will for the most part be using Alevin-fry. 

Some specific pieces of metadata I was thinking we should change was the `af permit type` and `af resolution` to just be `filter method` and `resolution`. I also wasn't sure if we wanted to change the index names at this point for external users or just keep them as they are. Please let me know if there is also other processing information not in the metadata that you think we should be including here as well.


inst/rmd/qc_report.rmd
```diff
@@ -20,7 +20,27 @@ output:
 library(SingleCellExperiment)
 ```
 
-# `r params$sample` Experiment Information
+# Processing Information for `r params$sample`
+
+```{r echo=FALSE}
+# extract sce metadata containing processing information as table
+processing_info <- sce@metadata %>%
```
jashapiro, line 27: This probably works most of the time, but it is dangerous, as `sce` is not defined in this notebook.

I also switched to use the `metadata()` function, as I think this is a bit more standard than directly accessing the slot name. 

```suggestion
processing_info <- metadata(params$sce) %>%
```

inst/rmd/qc_report.rmd
```diff
@@ -20,7 +20,27 @@ output:
...
+  as.data.frame() %>%
```
jashapiro, line 28: This works for the current structure of the metadata, but I am a bit worried about it, as it is entirely possible that we could end up with a metadata element that is not a single value. We also might have some metadata values we don't want to display.

jashapiro (commented):
I think that the approach you have taken here is okay, but I am a bit worried about its opacity. That is, it isn't clear from the transformations and table creation what metadata we are reporting. I think I would generally favor a more explicit approach, where we decide exactly which values we want to display, give them good names (probably more than removing underscores), and then create the table for display. This could also make it a bit easier to pass in other data (via `params`, or other sources) that might be worth reporting but isn't captured in the `sce` metadata.

allyhawkins:
> I think I would generally favor a more explicit approach, where we decide exactly which values we want to display, give them good names (probably more than removing underscores), and then create the table for display. This could also make it a bit easier to pass in other data (via params, or other sources) that might be worth reporting but isn't captured in the sce metadata.

I can definitely make this more explicit based on what we agree needs to go into this table. I am going to operate on the assumption that this is only being used for samples being processed with Alevin-fry right now and based on looking at both the QC reports from Alevin and Cellranger it looks like we might want a table that gives some version and parameter information. This would include the following: 

- Salmon Version
- Alevin-fry Version 
- Index
- Filtering Method (knee vs. unfiltered-pl)
- Resolution 
- Transcripts Counted (spliced vs. unspliced or do we want to refer to this as a different name?) 

It also looks like there might be usefulness in having a sample specific table with information about each specific sample that doesn't quite fall under the processing category including the following: 

- Sample ID
- 10X Kit
- Number of Reads
- Number of Mapped Reads
- Number of Cells Reported by Alevin-fry

If you could let me know what categories you like or don't like/ if you have any additional ideas that would be great and then I can go ahead and create the explicit tables that show these.

jashapiro:
> If you could let me know what categories you like or don't like/ if you have any additional ideas that would be great and then I can go ahead and create the explicit tables that show these.

These look good to me. As I said, nothing we can't add or subtract later!

A question I have is about the read counts that you listed in the second section: where are the read counts coming from? Are they as reported by alevin? If so, I think having something about the mapping rate might be worthwhile as well. Some of this is bleeding into QC, but if we are just using the data that comes out of Alevin and clearly state that, I think it should be fine. We can report the more limited data. 

I do think "transcripts counted" is a bit of an odd phrase, but I'm having trouble with a better compact phrase. Maybe just something boolean, like "Unspliced transcripts included" with a value of true or false/yes or no? I don't necessarily like that better.

allyhawkins:
@jashapiro, I adjusted these to make things more explicit so we are labeling them how we want and grabbing the fields that we want. I also added in the addition of a sample summary table that has information on the number of reads and mapped reads as I said in my previous comment. To get these values, I am pulling them directly from the Alevin output. In the `aux_info/meta_info.json` file it stores the number of reads processed and the number of reads mapped. This file is present for both alevin and alevin-fry. I adjusted the `read_alevin_metadata` function accordingly to grab that metadata that is being used to build the table. I didn't do much in terms of formatting or making things pretty but figured that could come later.

jashapiro (commented):
This looks good. Most of my suggestions are pretty minor, but some should streamline things a bit. 

One thing that is not in my line comments is that (for the second review in a row) I am going to suggest sentence case for the table row labels; I find it a bit easier to read. But I can be overruled as long as we are consistent.

R/read_alevin.R
```diff
@@ -199,6 +208,8 @@ read_alevin_metadata <- function(quant_dir){
   meta$af_resolution <- quant_info[['resolution_strategy']]
   meta$af_tx2gene <- cmd_info[['tgMap']]
   meta$usa_mode <- quant_info[['usa_mode']]
+  meta$num_cells <- quant_info[['num_quantified_cells']]
+  meta$version_10X <- names(cmd_info)[[3]]
```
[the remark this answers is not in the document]
allyhawkins: So this is something I was struggling with. I forgot to mention this in my previous comment, but I agree that this isn't the most stable method of getting the 10X version moving forward. This was the only way I could get it using the alevin output which is all we have currently as input to the `read_alevin` function where the metadata is added to the sce. Moving forward, I do think the best place to get it will actually be from the `scpca-library-metadata.tsv` sheet that we have that becomes the metadata map in the nextflow workflow. Because we plan on creating the QC report from the workflow, another option is to actually use the 10X version as input to the function along with the sample ID? Unless you have other ideas on how to grab this information I am open to suggestions.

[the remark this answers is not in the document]
jashapiro: We could make it an optional argument to `read_alevin`... I think there are a number of places that it could go, but that probably makes the most sense, so it ends up in the SCE metadata if desired.

R/read_alevin.R
```diff
@@ -39,7 +40,8 @@
 read_alevin <- function(quant_dir,
                         intron_mode = FALSE,
                         usa_mode = FALSE,
-                        which_counts = c("spliced", "unspliced")){
+                        which_counts = c("spliced", "unspliced"),
+                        version_10x = NULL){
```
jashapiro, line 44: I know everything has been 10X so far, but let's make this variable name more generic? Maybe something like `kit_version`, or `tech_version`? The values we are using do encode that it is 10X, so I don't think it should be ambiguous at all.

inst/rmd/qc_report.rmd
```diff
@@ -53,21 +48,16 @@ if(!is.null(processing_info)){
...
+  "Number of reads sequenced" = format(metadata_list$total_reads, big.mark = ',', scientific = FALSE),
+  "Number of mapped reads" = metadata_list$mapped_reads,
+  "Number of cells reported by alevin-fry" = metadata_list$af_num_cells
 ) %>%
-  t() %>%
-  # remove any empty slots before making table
-  na.omit()
+  t()
 
 if(!is.null(sample_information)){
```
jashapiro, line 60: After removing the `na.omit` step, I think this `if` is no longer needed.

R/read_alevin.R
allyhawkins: That makes sense. I updated it to be `tech_version`. Do we want the name in the table to also be `Tech version` or would we prefer something like `Technology`, `Technology version`, or`Technology used`?

DESCRIPTION
```diff
@@ -18,7 +18,7 @@ Description: Tools for processing single cell data associated with the
 License: BSD_3_clause + file LICENSE
 Encoding: UTF-8
 LazyData: true
-RoxygenNote: 7.1.1
+RoxygenNote: 7.1.1.9001
```
jashapiro, line 21: Trying to figure out why you have these extra digits for roxygen here. What OS are you using? This is the "full" version number, but whenever I install (on my mac or on the rstudio server), I get just a plain old `7.1.1`. The only reason this matters is that when I try to update docs etc. later, Roxygen complains that my version is older. But try as I might, I can't get this "newer" version (which is exactly the same, by all means I can tell)! 

So I am suggesting removing those extra digits, but I really am curious about the deeper explanation here.

```suggestion
RoxygenNote: 7.1.1
```

jashapiro (approved):
This looks good to me. I have a number of minor comments, but the overall content seems good to me. We will certainly be iterating on the report a bunch, so I don't think we need to get everything perfect at this stage, but getting the extra info into the metadata is going to be important for other steps later.

DESCRIPTION
allyhawkins: Hmm I'm on macOS Catalina v10.15.7? I have no idea how that would have happened but I'll remove those numbers...

PR #41, opened 2021-09-02
Create scpcaTools docker image 

### LINKED ISSUE #32: Create docker image
We have a docker image in https://github.com/AlexsLemonade/alsf-scpca that includes scpcaTools, but that image is big, and includes Rstudio, which won't be needed for workflows. We will create a new version of the docker image here, based on `r-vers`, with all dependencies (and related tools, including miQC). 

One note is that we will need to add this to `.Rbuildignore` so as not to cause problems with R package building.

jashapiro:
This PR adds a Dockerfile to create an image containing scpcaTools and all of its dependencies. 

closes #32

This is similar to the one that was in `alsf-scpca`, but it is more directly tied to this repository, which will make it easier to automate building the image.  (On that front, there is a `LABEL` field that ties this image to the repo: cool!)

The main difference is that this image is derived from `r-vers`, so it does not include RStudio. Since the main use of this will be in workflows, not interactive sessions, that seemed like a reasonable overhead to remove. It does mean we have to install `tidyverse` manually, but we can leave out some of the packages that are part of the rocker tidyverse image that we don't need at all. 

The other big changes are that I moved most of the installation out to a separate script. This is to make it easier to make sure intermediate files are deleted and don't become part of the docker image (keeping it as small as we can, which is actually not that small still, unfortunately). I could split up the layers a bit to potentially speed intermediate builds, but it isn't that bad as is. Most of the time is in installing Bioconductor packages from source; everything before that is binary installs anyway.

One question is if we want to bump the R version to 4.1; I don't expect any regressions (I have tested the package locally with R 4.1, and I think the automated testing is with latest R), but nor do I expect much benefit.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,13 @@
+FROM rocker/r-ver:4.0.5
+LABEL maintainer="ccdl@alexslemonade.org"
+
+COPY scripts/install_scpca_deps.sh .
+
+RUN bash ./install_scpca_deps.sh
+
+##########################
+# Install scpcaTools package
+RUN Rscript -e "remotes::install_github('AlexsLemonade/scpcaTools')"
+
+# set final workdir for commands
+WORKDIR /home
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,6 @@
 FROM rocker/r-ver:4.0.5
 LABEL maintainer="ccdl@alexslemonade.org"
+LABEL org.opencontainers.image.source https://github.com/AlexsLemonade/scpcaTools
 
 COPY scripts/install_scpca_deps.sh .
 
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,12 @@
+This folder contains a Dockerfile to build an image with R, scpcaTools and all of its dependencies.
+
+The image is built from a versioned Rocker image, with additional R packages and other utilities added as needed.
+
+It can be built with the following command run in this directory:
+
+```
+docker build . -t ghcr.io/alexslemonade/scpca-tools
+```
+
+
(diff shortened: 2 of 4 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM rocker/r-ver:4.0.5
+LABEL maintainer="ccdl@alexslemonade.org"
+LABEL org.opencontainers.image.source https://github.com/AlexsLemonade/scpcaTools
```
jaclyn-taroni, line 3: This _is_ cool.

docker/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM rocker/r-ver:4.0.5
```
jaclyn-taroni, line 1: Yea, could do `4.1.0` for the time being because it is the latest that will have a stable MRAN snapshot.

docker/README.md
```diff
@@ -0,0 +1,12 @@
...
+The image is built from a versioned Rocker image, with additional R packages and other utilities added as needed.
+
+It can be built with the following command run in this directory:
+
+```
+docker build . -t ghcr.io/alexslemonade/scpca-tools
+```
+
+
+Note that this image does not include Rstudio, in attempt to make the image smaller.
```
jaclyn-taroni, line 12: Sorry 🦅 👁️ 

```suggestion
Note that this image does not include RStudio, in attempt to make the image smaller. 
```

jaclyn-taroni (approved):
Looks good to me!

docker/Dockerfile
jashapiro: gonna leave it for now.. might change later.

PR #43, opened 2021-09-03
Automate docker builds

jashapiro:
This PR adds a workflow to build and push new docker images on commits to main and when we tag a new versioned release.

Releases are tagged with their version number, and as `latest`.
Commits to main are tagged with `edge`.
All commits are tagged with the commit hash and the date.

I made some updates to the license in an attempt to get github to recognize the license file and display some fancy chrome when people look at it. I don't know if it will work. If it does, this will also automatically label the image with the license type, which seemed like neat thing.


jaclyn-taroni (approved):
👍🏻  seems to do what it says based on the `docker/metadata-action` docs

PR #45, opened 2021-09-14
Use filtered and unfiltered data for QC report

### LINKED ISSUE #44: Support filtered and unfiltered SCE objects in generate_qc_report function
WE currently take only a single SCE object as input to the QC report. Now that we may have filtered and unfiltered versions, we should take both, with the filtered version optional.

After some discussion, the easiest way to handle the absence of a filtered SCE may be to simply make the filtered SCE object  a copy of the unfiltered, so that we do not have to have a bunch of conditional statements in the body of the report itself.

jashapiro:
Closes #44 

This PR updates the function for generating a QC report to (optionally) take two SCE objects, and makes a report from those two objects.

The report itself is still mostly a proof of concept, but a few of things are worth noting. 
- The assumption in the report is that there will be two SCE objects passed in with one a subset of the other (this is checked in the outer function). If there is only a single SCE object we will use a copy of the `unfiltered_sce` for the `filtered_sce` as well. Currently that copying happens in the `Rmd` processing, which is more due to happenstance than a conscious decision.
- I added a knee plot, because it seemed a reasonable way to test/demonstrate how the two SCEs might be used.
- There is also a section where I demonstrate one way of having text conditional on the presence or absence of a filtered SCE. Doing this too much will probably get to be a pain, so maybe we should assume the presence of both and just output the results, even when it makes less sense (i.e. the filtered and unfiltered numbers are the same).  I like having this as a demo for now though... We will likely have to do similar things when there is or is not `CITE-seq` data, for example.

### DIFF
--- a/R/generate_qc_report.R
+++ b/R/generate_qc_report.R
@@ -50,8 +50,8 @@ generate_qc_report <- function(sce,
     output_file = output_file,
     output_dir = output_dir,
     params = list(
-      sce = sce,
       sample = sample_name,
+      unfiltered_sce = sce,
       filtered_sce = filtered_sce
     ),
     envir = new.env()
--- a/R/generate_qc_report.R
+++ b/R/generate_qc_report.R
@@ -1,7 +1,7 @@
 #' Generate a QC report from a SingleCellExperiment object
 #'
-#' @param sce A SingleCellExperiment object that the report will describe
-#' @param sample_name The name of the sample
+#' @param sample_name The name of the sample for report headers
+#' @param unfiltered_sce A SingleCellExperiment object that the report will describe
 #' @param filtered_sce An optional filtered single cell experiment derived from first
 #' @param output The output file path that will be created.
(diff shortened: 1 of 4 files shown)


R/generate_qc_report.R
```diff
@@ -17,10 +18,23 @@
 #'
 generate_qc_report <- function(sce,
```
allyhawkins, line 19:
```suggestion
generate_qc_report <- function(unfiltered_sce,
```
To avoid confusion do we want to keep the naming convention consistent here with how it is in the actual qc report?

allyhawkins (approved):
I think this generally looks good to go. I just added in a comment about a small name change to keep it as `unfiltered_sce` and `filtered_sce` throughout rather than `sce` and `filtered_sce`.

R/generate_qc_report.R
jashapiro: We could, though I think my reasoning was that since only one sce is required, I would leave it as it was. Not sure we will ever use it that way, but that was my thought.

PR #46, opened 2021-09-14
Add option for `mtx_format` to import_quant_data

allyhawkins:
When working on the benchmarking for the 5' samples, I noticed an error in the `import_quant_data` function that the benchmarking functions currently use. I was getting errors when trying to run it and noticed that when the `mtx_format` option was added into `read_alevin` it was never incorporated into the `import_quant_data` function which directly uses `read_alevin`. Here, I am adding in that option so that you can now use `import_quant_data` again to read in alevin-fry data.


jashapiro (approved):
👍🏼

PR #48, opened 2021-09-15
Fix Docker build

### LINKED ISSUE #47: Docker build is broken
Apparently miQC requires Bioconductor 3.13, which requires R 4.1. 

Updating now, but ran into other bioconductor problems, because bioconductor 3.13 was updated after the release of R 4.1.1...

Also adding some checks to docker builds will fail rather than succeed & push without actually installing everything.

jashapiro:
Closes #47

With the addition of `miQC`, we needed to update to Bioconductor 3.13. Unfortunately, that broke a few things in the docker build. Worse, it broke them silently.

This PR fixes the failure by updating the base R image to 4.1.0, which updates Bioconductor to 3.13. Unfortunately, this also means that `MatrixGenerics`  has been updated (post release!), and now requires a version of `matrixStats` that came out _after_ R4.1.1 (which froze the R repo). So now I added a repo frozen to a later date.

The real solution here may be to use Renv, but I wanted a quick solution, and this did the job.

To prevent this from causing trouble in the future, I also added a few checks that should cause the docker build to fail if packages fail to install. I think it is fine to have such failures only after PRs are merged (i.e. not check docker build with every PR) as the result of a failure will just be not pushing the failing build to the package repo, which seems fine to me. We might add a badge or something to indicate that the docker image may not be up to date, but I don't think it is a terrible failure mode.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -7,8 +7,9 @@ COPY scripts/install_scpca_deps.sh .
 RUN bash ./install_scpca_deps.sh
 
 ##########################
-# Install scpcaTools package
-RUN Rscript -e "remotes::install_github('AlexsLemonade/scpcaTools')"
+# Install scpcaTools package (& test loading)
+RUN Rscript -e "remotes::install_github('AlexsLemonade/scpcaTools'); \
+  require(scpcaTools)"
 
 # set final workdir for commands
 WORKDIR /home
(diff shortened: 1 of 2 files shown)


jaclyn-taroni (approved):
Ugh. Agree with a potential move to `renv` later!

PR #50, opened 2021-09-15
Fix references to sce in QC function

jashapiro:
I thought I had fixed the references to `sce` in the QC report function, but apparently I missed some. 

This also updates docs to match, which didn't happen for whatever reason.


allyhawkins (approved):
LGTM!

PR #51, opened 2021-09-15
Add pandoc to docker

jashapiro:
To actually run the QC reports within the docker container, we need `pandoc`, so here I am adding it! I verified that this is sufficent to allow html files to knit. If we later want to make pdfs, we may need to add some of the suggested tex packages, but I don't think that is in our plans.


allyhawkins (approved):
LGTM!

PR #52, opened 2021-09-16
Add testing to qc report

jashapiro:
After yet another little bug that my ad-hoc testing missed, this PR adds minimal tests to the qc_report generation script (and fixes the bug).

The test is pretty minimal, but it does clean up after itself!
I also made the qc rendering quieter, because it was bugging me.


allyhawkins (approved):
Looks pretty straight forward to me, and does the job since it caught another bug 😄

PR #53, opened 2021-09-27
Function to add gene symbols to rowData

### LINKED ISSUE #35: Add gene symbols to rowData
We had a request to add gene symbols into the `SingleCellExperiment`. Our initial thoughts on how to do this were:

- Use the GTF (e.g., don't use `AnnotationDbi`)
- The GTF could be passed as a reference file argument to `read_alevin()` and the identifiers could be put in `rowData`
- If the GTF contains other identifiers, such as Entrez IDs, we might consider adding them, too

jashapiro:
Closes #35

Here I am adding a function that takes an SCE and a data.frame-like object (could be a GRanges object, and most of the time will be) and uses it to add a gene_symbol column to the row data of an SCE object.

Originally, I was going to make this use a GTF file as input, but I decided that was a bit too inflexible: I will add the GTF reading as part of  https://github.com/AlexsLemonade/scpca-nf/issues/38. 

There are a few steps here in the dplyr section to make sure that we only have a single match (if any) for each `gene_id`, and some checks to make sure the columns we need are present, but I don't check for content/reasonableness. So it is possible we will end up with no matches, but I figure that is an acceptable risk (which a user will hopefully notice).


R/add_gene_symbols.R
```diff
@@ -0,0 +1,35 @@
...
+#' @import SummarizedExperiment
+#' @importFrom rlang .data
+
+add_gene_symbols <- function(sce, gene_info){
+  if(is.null(gene_info$gene_id)){
+    stop("`gene_info` must contain a `gene_id` column.")
+  }
+  if (is.null(gene_info$gene_name)){
+    stop("`gene_info` must contain a `gene_name` column.")
+  }
```
allyhawkins, line 19: We probably want to add in a check for sce being a SingleCellExperiment here?

allyhawkins (commented):
I think this looks good to go, the only minor thing is that maybe we want to use the `magrittr` pipe to stay consistent with the rest of the functions in `scpcaTools`? I feel like it doesn't matter a whole lot but since the other functions use `%>%`, was there a specific reason you switched here?

jashapiro:
> I think this looks good to go, the only minor thing is that maybe we want to use the `magrittr` pipe to stay consistent with the rest of the functions in `scpcaTools`? I feel like it doesn't matter a whole lot but since the other functions use `%>%`, was there a specific reason you switched here?

I was basically thinking that we could start to switch and remove a (minor) dependency...but I am reconsidering. With the inclusion of miQC, I was thinking that we were effectively requiring R 4.1, but on second thought, that isn't really true (somebody could theoretically install miQC on an earlier version via github), so I will change it back.

jashapiro:
I made one other change... while working on the next step, I noticed that `scater::addPerCellQC` is sort of deprecated with the preference to `scuttle::addPerCellQCMetrics` (and the same for features) so I converted to use those here.

> `addPerCellQC` and `addPerFeatureQC` are exactly the same functions, sans the `Metrics` at the end of their names. They were added in the tempestuous youth of this package when naming was fast and loose. These can be considered to be soft-deprecated in favor of the longer forms.

allyhawkins (approved):
LGTM!

PR #56, opened 2021-09-27
add scater back to image

jashapiro:
I realized that I may have been too aggressive in removing `scater` from the docker image, as that might break things down the line, so I am adding it back here. I don't know that it will break things, but I don't want things broken tomorrow while I am out!


allyhawkins (approved):
Looks good.

Issue #54, opened 2021-09-27
Implement fenced divs to make a two-column html output.


jashapiro:
I looked into making two column output via markdown. It looks like we can do that with "fenced divs" https://bookdown.org/yihui/rmarkdown-cookbook/multi-column.html

Basically we add a bunch of `::: {}` elements where we want to create divs, and enclose ones we want side by side with `::::{style="display: flex;"}` 

We can try it at least!

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpcaTools/issues/37#issuecomment-928209017_

jashapiro:
I played around with this a bit today, and while it is easy to do from a structural point of view, it does introduce a lot of visual tweaking that would probably require some time with css to make look good. I don't know if we want to devote the design time to this at this stage, but we can circle back to it.

jaclyn-taroni:
I'm going to take this out of the current Epic.

sjspielman:
Noting I'm sending this issue over to `scpca-nf`, which is where the QC report now lives.

PR #57, opened 2021-09-30
Add plots to QC report

### LINKED ISSUE #19: Add knee plot to QC report
We may want to include a knee plot in the QC report for each sample where we highlight cells that are included and excluded. This will require us to have information about both the unfiltered and filtered cells which could be read in as two separate SCE's, or we could use the unfiltered sce and highlight just the cell barcodes that have passed filtering.

### LINKED ISSUE #39: Add plot to QC: Total count vs. gene counts by mito content
We have the R code to make this plot _somewhere_ already. This issue is to track having this plot make it into the QC report we provide to users.

### LINKED ISSUE #40: Collapsible session info section in QC report, hidden by default
We do have session info in the template Rmd file currently: https://github.com/AlexsLemonade/scpcaTools/blob/a8e7f416241040aa1034591ddf3f5e80581a1072/inst/rmd/qc_report.rmd#L27

We can imagine a situation where that is long and not helpful unless a user is especially motivated to look at it. So perhaps we can make the section collapsible and hidden by default.

### LINKED ISSUE #55: Add miQC plot to QC report
As discussed in #37, we want to show a plot of the miQC model in the QC report. This would mean a plot similar to what is produced using `miQC::plot_model()` where genes detected is on the x-axis, % mito is on the y-axis, cells are colored by probability compromised, and the linear mixture model lines are overlaid on the plot. Because we are storing the model as metadata of the filtered sce, we should be able to grab the model from there and then create the plot that way rather than rerunning miQC.

jashapiro:
Here I am mostly adding the Total vs. gene count plot to the QC report. I also fixed the knee plot and made the session info hidden by default, so closing all three of those issues here.

I was also playing around with #54, which I ended up abandoning, but in that process I added a bit of formatting to the tables, including making them not full-width, and striping the rows.

A few things to focus on that I was thinking about as I did them: 
- Please also look at all axis labels and legend titles! (Should I make the knee plot use non-scientific notation?)

- For consistency among reports, I made the mitochondrial percentage go from 0-100% for all reports. This may make lower numbers harder to see in mostly-good plots (where no cells are above 50%, say). I thought the consistency was more important, but we could tweak the color scale if we think it matters.

- The other potentially controversial aesthetic decision I made was to move the figure legends into the figure panels, at least for the knee & UMI/gene plots. In both of these plots, the location of the data is predictable, so in the vast majority of cases the positions I chose will not overlap data. Still, there is a chance they will, so if there is worry, I can move them back out.  (when we add the miQC plot, the top right ought to work as a clear space)

I am attaching an [example report](https://github.com/AlexsLemonade/scpcaTools/files/7262155/SCPCL000001_qc.html.zip) (zipped, b/c github), and the two main figures I worked on below, for reference:
image
image

closes #39
closes #40
closes #19
closes #55

### DIFF
--- a/inst/rmd/qc_report.rmd
+++ b/inst/rmd/qc_report.rmd
@@ -13,7 +13,7 @@ output:
     toc: true
     toc_depth: 2
     toc_float: true
-    number_sections: true
+    number_sections: false
     code_download: true
 ---
 
@@ -27,13 +27,13 @@ library(SingleCellExperiment)
 library(dplyr)
 library(ggplot2)
 
-# Set standard ggplot theme
+# Set default ggplot theme
 theme_set(theme_bw())
 ```
 
 ```{r sce_setup}
 # save some typing later
-sample <- params$sample
+sample_id <- params$sample
 unfiltered_sce <- params$unfiltered_sce
 filtered_sce <- params$filtered_sce
 
@@ -51,7 +51,7 @@ if (is.null(unfiltered_sce$sum)){
 
 # Introduction
 
-# Processing Information for `r sample`
+# Processing Information for `r sample_id`
 
 ## Raw Sample Metrics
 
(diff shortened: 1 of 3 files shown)


jashapiro:
Adding in the `miQC` plot as well, which looks like this (not included in the example report above):

image

Same story here where I moved the legend inside the plot. In most cases the upper right should have no genes, so it ought to be safe, but I can move back outside if that is preferred.

allyhawkins (commented):
Overall this is coming together really nicely! I just wanted to comment on a few minor things: 

1. I like the placement of the legends and am not overly concerned about if they overlap data. I think for the majority of data where you've placed them should be outside of where the data lie. 
2. We should be consistent with the type of case we choose to use for the axis titles. I think we could use sentence case as we do most everywhere else, and I think you are mostly using that now so I suggested a few changes to make that consistent throughout. 
3. You mentioned in the comment that you made the legend go from 0-100% for mito content. I think we should do the same thing with plotting the miQC plot. Since we always anticipate that axis to be on the same scale.

jashapiro:
I updated legends and titles as suggested, using sentence case everywhere, and unifying labels.

I moved what I had put as figure captions to alt text because it was laregely redundant, as noted, but I thought we should have some kind of figure label beyond the title. 

For the knee plot, I chose a more contrasting color scheme, but I didn't want to go to red, because that would imply (to me) that those points failed some QC. So I went with a dark green, and lightened up the grey. I also went with smaller points for the filtered cells. Let me know what you think. 
image

Here is a full report file...
[SCPCL000001_qc.html.zip](https://github.com/AlexsLemonade/scpcaTools/files/7280358/SCPCL000001_qc.html.zip)

inst/rmd/qc_report.rmd
```diff
@@ -103,32 +123,85 @@ if(has_filtered){
...
+
+```{r, fig.alt="miQC model diagnostics plot", results='asis', warning=FALSE}
+if(skip_miQC){
+  cat("miQC model not created, skipping miQC plot. Usually this is because mitochondrial gene data was not available.")
+}else{
+  # remove prob_compromised if it exists, as this will cause errors with plotModel
+  filtered_sce$prob_compromised <- NULL
+  miQC::plotModel(filtered_sce, model = metadata(filtered_sce)$miQC_model) + 
+    coord_cartesian(ylim = c(0,100)) +
+    labs(x = "Number of genes expressed",
```
allyhawkins, line 189:
```suggestion
    labs(x = "Number of genes detected",
```
Just noticing that we have detected in the other plot so either we should change it here or change it to expressed in the other plot, I'm fine with either.

allyhawkins (approved):
I like the changes and think the green in the knee plot looks a lot better! I don't necessarily think we need the change in size and have previously been taught not to double encode variables where we have both size and color showing if a cell is considered `passed` or `excluded` but I'll leave that final call up to you. I do think the changes highlight the passed cells which is what we want so I can see an argument for either one.

I had one other minor axis wording change to keep the labels consistent between the `Cell read metrics` and `miQC` plot, but other than that I think this looks good!

jashapiro:
> I don't necessarily think we need the change in size and have previously been taught not to double encode variables where we have both size and color showing if a cell is considered `passed` or `excluded` but I'll leave that final call up to you.

Yeah, I don't usually do this, but I thought it was worth it in this case, as I think it solves a problem where the two kinds of points overlap.

allyhawkins:
> Yeah, I don't usually do this, but I thought it was worth it in this case, as I think it solves a problem where the two kinds of points overlap.

I would agree, it does help show the separation a lot more. I think for the purposes of this plot its okay to keep it in.

Issue #37, opened 2021-09-02
[Placeholder / Discussion] Draw a picture of what the QC report will contain for MVP


jaclyn-taroni:
A good way to make sure we're all on the same page about the contents and layout of the QC report is to sketch out what the overall report will look like. 

From there, we can 1. disagree a little bit initially, probably 2. ensure we have issues for all the relevant sections 🚀

allyhawkins:
In an effort to get this started I went ahead and created a very very rough drawing of what I think we are looking for in the QC report. The first two tables, `sample_information` and `processing_information` we already have a version added to the report so I just filled in some examples of data that is in there but it is not complete so as not to crowd the picture. 

I also tried my best to draw out the plots but in general, I think we would have the following plots: 
- our usual plot we show during training with UMI/cell on the x-axis, genes/cell on the y-axis and colored by mito content 
- knee plot where we are coloring cells based on if they are included in the filtered sce object or not (filtering here referring to `emptyDrops` filtering) 
- plot showing the miQC model with unique genes on the x-axis and mito on the y-axis and showing the model. Would we rather change this and maybe color the cells based on if they are keep or throw based on miQC? 

I also added in a section for ADT/ CITE-seq information and noticed on the [issue](https://github.com/AlexsLemonade/scpcaTools/issues/38) that there was a comment to add in the list of antibodies. I wasn't sure how lengthy that could get or how we would want to portray that so I thought we might want to have that separate from the statistics? 

I just wanted to get a general idea of the order of things and the blocks that we need to put in place so this is pretty rough but would love to hear @jaclyn-taroni and @jashapiro thoughts/ opinions/ disagreements/ suggestions. I also wanted to keep in mind that this is for MVP so I did not choose to include any UMAPs or clustering at this point. 

<img width="397" alt="Screen Shot 2021-09-24 at 4 32 20 PM" src="https://user-images.githubusercontent.com/54039191/134741919-276d5d8f-c28f-48f3-8493-cd1482807c6c.png">

jaclyn-taroni:
This seems perfectly reasonable to me for MVP, which is to say I have no disagreements. I'm uploading a version of that sketch with lettered panels just in the interest of making the next steps easier to talk about.

<img width="397" alt="134741919-276d5d8f-c28f-48f3-8493-cd1482807c6c" src="https://user-images.githubusercontent.com/19534205/134955140-cf93b960-3491-4261-8b8b-f7c9e82fa57d.png">


[...]

PR #58, opened 2021-10-04
Add table with basic cell statistics to QC report

### LINKED ISSUE #18: Attach basic statistics of per cell QC metrics in QC report
In the QC report we should calculate some metrics on a per cell level. This can be done using `scpcaTools::add_cell_mito_qc()`, which will add `colData` to the SCE. We will need to extract that `colData` and output information like: 

- UMI/cell
- Genes/cell
- % mito/cell 

In doing this, we should think about including both a table with the median and mean value of these numbers (and maybe standard dev?) and a plot showing the distribution of these values.

allyhawkins:
Closes #18. Based on discussions in both #37 and #18, I have added in a table that includes some basic cell statistics to the qc report. These are statistics that are calculated after filtering using `emptyDrops`, so in addition to including the median UMI/cell, genes detected/ cell and % mito/cell, I also added in the number of cells post filtering at the top of the table. I thought this table fit in best in the `Experiment Summary` section, but before the knee plot. 

A few notes/ questions for the changes implemented here:

- Do we want to include both median and mean? I included median for now but I know Cell Ranger includes both so we could add that in if we think it's necessary. 
- While working on this I noticed that the majority of the section headers were title case other than a few in sentence case, so I chose to convert them all to title case. Let me know if you would prefer it the other way around - I just went with the majority on this one. 
- I also removed the `Introduction` header, since there wasn't anything that seemed to be really underneath it and it seemed out of place, but let me know if there was an intended use there. 
- If there are any other basic statistics on a per cell level that you think we should be including here, happy to discuss putting them in!

I have included a copy of an [example report](https://github.com/AlexsLemonade/scpcaTools/files/7281347/SCPCL000006_example.html.zip) for help in reviewing. 

Here is a screenshot of the table that has been added as well: 
Screen Shot 2021-10-04 at 4 15 44 PM

### DIFF
--- a/inst/rmd/qc_report.rmd
+++ b/inst/rmd/qc_report.rmd
@@ -12,8 +12,7 @@ output:
   html_document:
     toc: true
     toc_depth: 2
-    toc_float: 
-      collapsed: false
+    toc_float: true
     number_sections: false
     code_download: true
 ---
--- a/inst/rmd/qc_report.rmd
+++ b/inst/rmd/qc_report.rmd
@@ -120,21 +120,21 @@ if(has_filtered){
 }
 ```
 
-## Basic Cell Statistics 
+## Cell Statistics 
 
 ```{r}
 basic_statistics <- tibble::tibble(
   "Number of cells post filtering"              = format(ncol(filtered_sce), big.mark = ","),
   "Median UMI count per cell"                   = format(median(filtered_sce$sum), big.mark = ","),
   "Median genes detected per cell"              = format(median(filtered_sce$detected), big.mark = ","),
-  "Median percent reads mitochondrial per cell" = paste0(round(median(filtered_sce$subsets_mito_percent), 2), " %")
+  "Median percent reads mitochondrial per cell" = paste0(round(median(filtered_sce$subsets_mito_percent), 2), "%")
   ) %>%
   mutate(across(.fns = ~ifelse(.x == "NULL", "N/A", .x))) %>%
   t()
 
 
 # make table with basic statistics
-knitr::kable(basic_statistics) %>%
+knitr::kable(basic_statistics, align = "lr") %>%
   kableExtra::kable_styling(bootstrap_options = "striped",
                             full_width = FALSE,
                             position = "left")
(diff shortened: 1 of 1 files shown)


inst/rmd/qc_report.rmd
```diff
@@ -122,7 +120,30 @@ if(has_filtered){
...
+  "Median UMI count per cell"                   = format(median(filtered_sce$sum), big.mark = ","),
+  "Median genes detected per cell"              = format(median(filtered_sce$detected), big.mark = ","),
+  "Median percent reads mitochondrial per cell" = paste0(round(median(filtered_sce$subsets_mito_percent), 2), " %")
+  ) %>%
+  mutate(across(.fns = ~ifelse(.x == "NULL", "N/A", .x))) %>%
+  t()
+
+
+# make table with basic statistics
+knitr::kable(basic_statistics) %>%
```
jashapiro, line 137: Right justifying the numbers (we should probably do this everywhere).
```suggestion
knitr::kable(basic_statistics, align = "lr") %>%
```

jashapiro (commented):
This looks good. 

I can go either way on case, so title case is fine until somebody tells us differently.

I was looking back at the stats that Cell Ranger reports... Many I don't think are too useful, but maybe we should add the % reads in cells? I think this would just be the total reads in filtered/ total in unfiltered, so something like:
```
sum(filtered_sce$sum)/sum(unfiltered_sce$sum)
```

We could also maybe change the # of reads mapped to % reads mapped in the sample metrics? I think that will be easier for people to interpret. 


As for the introduction, I think that is fine to remove. We might add some notes at the bottom with a link to the docs, once we have them. We may well want to add a bit more explanatory text for each section before we call this all done.

allyhawkins:
> We could also maybe change the # of reads mapped to % reads mapped in the sample metrics? I think that will be easier for people to interpret.

I would definitely agree with this, I'll go ahead and make this change. I'll also go ahead and add in the % reads in cells, I think that could be useful to know for samples.

allyhawkins:
@jashapiro I added in the % reads mapped, and % reads in cells and made the label change for the mitochondrial reads. I think we will need to go back through and add text for all the plots, but I haven't added any in here yet. I've attached the updated version of the report for reference. 
[SCPCL000006_test.html.zip](https://github.com/AlexsLemonade/scpcaTools/files/7286397/SCPCL000006_test.html.zip)

jashapiro (approved):
LGTM!

PR #62, opened 2021-10-05
59 Add explanatory captions to QC report plots

jashapiro:
Here I am adding explanatory captions to the QC report for each plot. These captions may or may not be ideal, so editing suggestions are encouraged!

I have for now included both versions of the knee plot, but I am definitely leaning toward the smoothed version, which is what I refer to in the caption. I think it better shows where the bulk of cells are coming from vs. being filtered out. On the other hand, there are a lot of droplets with lower counts that make it through, because there are a **lot** of droplets with lower counts! Anyhow, compare for yourself in the attached file:

[SCPCL000001_qc.html.zip](https://github.com/AlexsLemonade/scpcaTools/files/7288209/SCPCL000001_qc.html.zip)

I also made a few small aesthetic changes: 
- All parameter/statistic tables now use monospace fonts and right justification for the "values" column.
- Table of contents starts expanded.

Are there other places where we should add explanatory text?

Filing as a draft until we resolve the knee plot question.

### DIFF
--- a/inst/rmd/qc_report.rmd
+++ b/inst/rmd/qc_report.rmd
@@ -218,7 +218,7 @@ ggplot(grouped_celldata, aes(x = med_rank, y = med_sum, color = pct_passed)) +
 ```
 
 The total UMI count of each droplet (barcode) plotted against the rank of that droplet allows visualization of the distribution of sequencing depth across droplets. 
-The droplets that are expected to contain cells were identified with [`DropletUtils::emptyDrops()`](https://bioconductor.org/packages/release/bioc/html/DropletUtils.html) , which uses both the total UMI counts and expressed gene content ([Lun  _et al._ 2019](https://doi.org/10.1186/s13059-019-1662-y)).
+The droplets that are expected to contain cells were identified with [`DropletUtils::emptyDrops()`](https://bioconductor.org/packages/release/bioc/html/DropletUtils.html), which uses both the total UMI counts and expressed gene content ([Lun  _et al._ 2019](https://doi.org/10.1186/s13059-019-1662-y)).
 As the boundary between droplets passing and failing this filter does not correspond only to the axes of this plot, some regions contain droplets in both categories, so we here color the plot by the percentage passing the filter.
 
 ## Cell Read Metrics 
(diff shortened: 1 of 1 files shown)


jaclyn-taroni:
Sorry to be annoying (read: off topic) but can we change the author to `Childhood Cancer Data Lab` from `CCDL` please? (I can not comment on the line itself and it seems we might as well do that here.)

jashapiro:
> Sorry to be annoying (read: off topic) but can we change the author to `Childhood Cancer Data Lab` from `CCDL` please? (I can not comment on the line itself and it seems we might as well do that here.)

Done!

inst/rmd/qc_report.rmd
```diff
@@ -157,25 +161,65 @@ unfiltered_celldata <- data.frame(colData(unfiltered_sce)) %>%
...
+  ) + 
+  theme(legend.position = c(0, 0),
         legend.justification = c(0,0),
         legend.background = element_rect(color = "grey20", size = 0.25),
         legend.box.margin = margin(rep(5, 4)))
 ```
 
+The total UMI count of each droplet (barcode) plotted against the rank of that droplet allows visualization of the distribution of sequencing depth across droplets. 
+The droplets that are expected to contain cells were identified with [`DropletUtils::emptyDrops()`](https://bioconductor.org/packages/release/bioc/html/DropletUtils.html) , which uses both the total UMI counts and expressed gene content ([Lun  _et al._ 2019](https://doi.org/10.1186/s13059-019-1662-y)).
+As the boundary between droplets passing and failing this filter does not correspond only to the axes of this plot, some regions contain droplets in both categories, so we here color the plot by the percentage passing the filter.
```
jaclyn-taroni, line 222: I might restate the fact that the expressed gene content is used for filtering in the second sentence here. Repetition is good.

jaclyn-taroni: When I read this (in the rendered form, where I can't see code), I do find myself asking how we get the `rank_group` exactly but I'm not our user.

jaclyn-taroni: Sorry I said rank group as a matter of convenience (I can see the code in the diff after all). I basically meant how do we get the averages.

allyhawkins (approved):
I like the changes, all looks good to me.

Issue #60, opened 2021-10-05
Tag release v0.1.0


jashapiro:
Once we are satisfied with the QC report & seed setting options, we should be ready to tag a release of the scpcaTools package. This should also prompt the docker image to build, so we can use the tagged version in a similarly versioned release of the [scpca-nf](https://github.com/AlexsLemonade/scpca-nf) workflow.

jashapiro:
Done!

PR #63, opened 2021-10-05
Add a CITE-seq section to the QC report

### LINKED ISSUE #38: Add a CITE-seq section to the QC report
Some initial ideas for contents, taken from meeting notes:

- ADT table normally in the CellRanger report (number of barcode reads per cell)
- List the antibodies in a table
- Number of filtered cells that have antibody counts

jashapiro:
Note: stacked on #60

Here I am adding a section with CITE-seq statistics for samples that have them. 
I did this partly by adding a separate document with the CITE-seq content, which is then included conditionally as a "child" document.

The content in that chunk includes a table that is a bit of a hybrid of the `sample_information` and `basic_statistics` tables from the main page. I thought that separating out into two tables was probably not required here, as I didn't want to add too much. I think this covers the highlights of the fields that Cell Ranger reports. 

The second table shows the % of cells with each ADT detected. I made this table more compact so it doesn't take up too much space. If we still think it is too big, we could make it scroll, though it looked a bit ugly when I did that.

You might notice some tortured logic in there to get the ADT sorting as I wanted it, which is basically alphabetical with natural number sorting (i.e. CD2 comes before CD15). I hope that will hold up to future data sets! An alternative would be to just rank by the detected value. 

Other random thoughts:
- I added a line for "modalities" in the main `sample_information` table. Do we like that? 
- Header names? Maybe we should change "SCPCL00000 Experiment Summary" to  RNA-seq experiment summary?
- Do we still need the  text under the experiment summary? https://github.com/AlexsLemonade/scpcaTools/blob/94bc15743bbdacf6200db7f68db8eecefbaaeb04/inst/rmd/qc_report.rmd#L115-L121

Example report: [SCPCL000050_qc.html.zip](https://github.com/AlexsLemonade/scpcaTools/files/7289174/SCPCL000050_qc.html.zip)

Closes #38

### DIFF
--- a/inst/rmd/cite_qc.rmd
+++ b/inst/rmd/cite_qc.rmd
@@ -1,6 +1,7 @@
 # CITE-seq Experiment Summary
 
 ## CITE-seq Statistics
+
 ```{r}
 cite_exp <- altExp(filtered_sce, "CITEseq")
 cite_meta <- metadata(cite_exp)
@@ -10,7 +11,7 @@ if (is.null(rowData(cite_exp)$detected)){
   cite_exp <- scuttle::addPerFeatureQCMetrics(cite_exp)
 }
 
-cell_adt_counts <- colSums(counts(cite_exp))
+cell_adt_counts <- Matrix::colSums(counts(cite_exp))
 
 cite_information <- tibble::tibble(
   "Number of ADTs assayed" = 
@@ -31,7 +32,7 @@ cite_information <- tibble::tibble(
 knitr::kable(cite_information, align = 'r') %>%
   kableExtra::kable_styling(bootstrap_options = "striped",
                             full_width = FALSE,
-                            position = "left")
+                            position = "left") %>%
   kableExtra::column_spec(2, monospace = TRUE)
 ```
 
@@ -42,8 +43,8 @@ antibody_tags <- as.data.frame(rowData(cite_exp)) %>%
   select("Antibody",
          "Percent of cells detected" = detected) %>%
   # sort by leading characters, then numeric portion
-  arrange(stringr::str_extract(`Antibody tag`, "^[A-Za-z]+"),
-          as.numeric(stringr::str_extract(`Antibody tag`, "\\d+")))
+  arrange(stringr::str_extract(Antibody, "^[A-Za-z]+"),
+          as.numeric(stringr::str_extract(Antibody, "\\d+")))
 
 knitr::kable(antibody_tags, digits = 2) %>%
   kableExtra::kable_styling(bootstrap_options = "striped",
--- a/inst/rmd/cite_qc.rmd
+++ b/inst/rmd/cite_qc.rmd
@@ -1,7 +1,8 @@
-## CITE-seq statistics
+# CITE-seq Experiment Summary
 
+## CITE-seq Statistics
 ```{r}
-cite_exp <- altExp(filtered_sce2, "CITEseq")
+cite_exp <- altExp(filtered_sce, "CITEseq")
 cite_meta <- metadata(cite_exp)
 
 # add rowData if missing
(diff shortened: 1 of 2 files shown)


inst/rmd/cite_qc.rmd
```diff
@@ -0,0 +1,57 @@
...
+
+## Antibody Derived Tag Statistics 
+```{r}
+antibody_tags <- as.data.frame(rowData(cite_exp)) %>%
+  tibble::rownames_to_column("Antibody") %>%
+  select("Antibody",
+         "Percent of cells detected" = detected) %>%
+  # sort by leading characters, then numeric portion
+  arrange(stringr::str_extract(Antibody, "^[A-Za-z]+"),
+          as.numeric(stringr::str_extract(Antibody, "\\d+")))
```
allyhawkins, line 47: I personally think the alphabetical order is not as important here and would rather see them listed by % of cells detected in.

inst/rmd/qc_report.rmd
```diff
@@ -73,10 +82,21 @@ unfiltered_meta <- metadata(unfiltered_sce)
...
   "Percent of reads mapped" = paste0(round((unfiltered_meta$mapped_reads/unfiltered_meta$total_reads) *100, 2), "%"),
   "Number of cells reported by alevin-fry" = format(unfiltered_meta$af_num_cells, big.mark = ',', scientific = FALSE)
-) %>%
+)
+if (has_cite){
+  cite_meta <- metadata(altExp(unfiltered_sce, "CITEseq"))
+  sample_information <- sample_information %>%
+    mutate(
+      "CITE-seq reads sequenced" = format(cite_meta$total_reads, big.mark = ',', scientific = FALSE),
+      "CITE-seq reads mapped" = format(cite_meta$mapped_reads, big.mark = ',', scientific = FALSE),
```
allyhawkins, line 95: I think this should also be percent reads for CITE-seq to be consistent with the RNA-seq reporting above?

allyhawkins (commented):
I think this looks really good! I like the addition of the modalities in the main table. I just had a few minor inline comments but also to answer your questions on the places that you didn't change/ I can't comment on: 

1. I do think we should change `SCPCL000 Experiment Summary` to `RNA-seq Experiment Summary` to be consistent with the CITE-seq part. The SCPCL000 is also at the top of the table of contents so I don't think we need it again. 
2. I also would agree that we could get rid of the text underneath the RNA-seq experiment summary before `Cell Statistics`, but if we do that we should add the total number of genes that are detected to the `Cell Statistics` table since the only place we have it currently is in that text. We could also add  the number of cells pre-filtering with `emptyDrops`, but I'm not sure that we really need it.

inst/rmd/qc_report.rmd
jashapiro: Yup! I thought I had done this, but apparently not!

jashapiro:
> 2\. I also would agree that we could get rid of the text underneath the RNA-seq experiment summary before `Cell Statistics`, but if we do that we should add the total number of genes that are detected to the `Cell Statistics` table

Should this be in the `Cell Statistics` or `Preprocessing stats`? Since we decided not to filter out zero count genes, the number is really set by the salmon index, I think, so it might make sense there. But I don't really have a strong feeling about where people will expect it.

inst/rmd/qc_report.rmd
jashapiro: Oh, I see the issue... I had actually meant to delete these lines, as they are redundant with the info in the CITE-seq section (where I do have it as % mapped%). Do you think we should have this info in both places, or should I delete it?

allyhawkins:
> Should this be in the Cell Statistics or Preprocessing stats? Since we decided not to filter out zero count genes, the number is really set by the salmon index, I think, so it might make sense there. But I don't really have a strong feeling about where people will expect it.

Hmm you are right, since it's technically the number of genes being assayed and related to the index and not the actual genes that are detected maybe this belongs in the `Preprocessing Information` instead.

allyhawkins:
> I also updated the sort of the ADT table (which only made me a little sad). I meant to ask earlier if you thought I should include the mean UMI counts for each tag. It's already calculated, so that is easy, and it may result in a somewhat more stable sort (I can imagine the 99%+ detected class would move around stochastically).

I do like this idea of adding in a third column with mean UMI/cell and then sorting on that rather than detected. Should we do median instead though? When doing this I would make the order of the columns appear as `Antibody`, `Median UMI per cell`, `Percent of cells detected`. Everything else looks good though so if you decide to make that change I'll wait until that's in to approve.

jashapiro:
> > I also updated the sort of the ADT table (which only made me a little sad). I meant to ask earlier if you thought I should include the mean UMI counts for each tag. It's already calculated, so that is easy, and it may result in a somewhat more stable sort (I can imagine the 99%+ detected class would move around stochastically).
> 
> I do like this idea of adding in a third column with mean UMI/cell and then sorting on that rather than detected. Should we do median instead though? When doing this I would make the order of the columns appear as `Antibody`, `Median UMI per cell`, `Percent of cells detected`. Everything else looks good though so if you decide to make that change I'll wait until that's in to approve.

I thought about median, but that will be 0 for anything found in < 50%, and probably 1 for lots more, so I think mean is a better metric in this case.

allyhawkins (approved):
LGTM 🚀

PR #67, opened 2021-10-08
66 mi qc errorhandling

### LINKED ISSUE #65: remove scientific notation from QC plots
In the knee plot for teh QC report, we use `scales::label_number` to force non-scientific notation for the axis tick labels.
See: https://github.com/AlexsLemonade/scpcaTools/blob/main/inst/rmd/qc_report.rmd#L203-L204 (Note that `accuracy` is used here when the numbers may not be integers to prevent `.0` labels. I think it may be an artifact of earlier versions of this plot and may not be needed anymore.)

In initial testing this wasn't needed for other plots, but there are datasets where the tick numbers get large enough to force ggplot to switch. We should therefore implement the same scale settings for the "Cell read metrics" plot and the miQC plot. The latter is probably not going to be an issue as gene counts can only get so large, but just to be safe we might as well.

### LINKED ISSUE #66: Handle miQC model fit failures
When the miQC model fit fails (fits a single component instead of two), we can get errors when calling `miQC::plotModel()` as we do in the QC report: https://github.com/AlexsLemonade/scpcaTools/blob/92c0b49c0202cb394a2ea7ee164eefd63d17f4e1/inst/rmd/qc_report.rmd#L257 

We will also get an error if we are adding the miQC `prob_compromised` via `miQC::filterCells()` using our `add_cell_mito_qc()` function. https://github.com/AlexsLemonade/scpcaTools/blob/92c0b49c0202cb394a2ea7ee164eefd63d17f4e1/R/add_cell_mito_qc.R#L58

A solution to these is to add a quick check that `length(model@components) < 2`. For the filtering, am going to propose adding a separate exported function `add_miQC()` that performs that check and fills the `prob_compromised` column with `NA_real_`. We can then use that function separately to reliably add the miQC model and results to a SCE, with NA in the case that the model fit is not as expected.

jashapiro:
Closes #66 

This PR adds an `add_miQC()` function to the package, partly factors out that functionality from the `add_cell_mito_qc()` function, but extends it to handle failures of model fit. In the case of model fit failure, I don't include the miQC model in the output metadata, but I do fill in `prob_compromised` with `NAs`.

I then use the results of this potential fit failure to update the QC document: if the model has failed to fit, I use `plotMetrics()` instead of `plotModel()`. I also gave a bit more information about the potential reasons for fit failure in the descriptive text.

Since I was editing one QC plot, I also fixed up the formatting on the other so this also
closes #65


allyhawkins (approved):
I think this looks good to go. I tested it a few different times and tried to get it to break and it seems to catch the failures that we would anticipate (at least the ones I could think of for now). I think this should at least solve this problem!

PR #68, opened 2021-10-08
Add seed to arguments for `filter_counts`

### LINKED ISSUE #61: Add argument to set seed to `filter_counts()`
Based on discussion had in https://github.com/AlexsLemonade/scpca-nf/issues/45 and in the scpca meeting, we will want to add an argument to the function to set the seed during the `filter_counts()` function. We should set the default of this to be `NULL` and then use the workflow in `scpca-nf` to set the seed in the global environment rather than set a default of a value inside this function.

allyhawkins:
Closes #61. As we had discussed previously it would be good to add in an option to set the seed within the `filter_counts` function before running `emptyDrops`. Here, I am adding in an argument for the seed and setting the default to `NULL` and then setting the seed at the beginning of the script. I believe we also talked about not needing error checks for this so I didn't add any in, but let me know if there is an error check we should have for this.

### DIFF
--- a/R/filter_counts.R
+++ b/R/filter_counts.R
@@ -3,6 +3,7 @@
 #' @param sce SingleCellExperiment with unfiltered gene x cell counts matrix.
 #' @param fdr_cutoff FDR cutoff to use for DropletUtils::emptyDrops.
 #'   Default is 0.01.
+#' @param seed A random seed for reproducibility.
 #' @param ... Any arguments to be passed into DropletUtils::emptyDrops.
 #'
 #' @return SingleCellExperiment with filtered gene x cell matrix.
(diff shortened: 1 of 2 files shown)


R/filter_counts.R
```diff
@@ -3,6 +3,7 @@
 #' @param sce SingleCellExperiment with unfiltered gene x cell counts matrix.
 #' @param fdr_cutoff FDR cutoff to use for DropletUtils::emptyDrops.
 #'   Default is 0.01.
+#' @param seed A random seed for reproducibility.
```
jashapiro, line 6:
```suggestion
#' @param seed An optional random seed for reproducibility.
```

PR #69, opened 2021-10-09
Keep fewer images on ghcr

jashapiro:
Now that we have some releases under our belt, I think that the practice of keeping docker images for every release may have outlived its usefulness. So here I am removing tagging images built at every push to main with the sha hash and date. The images are still built and pushed with the `edge` tag, but this will mean that we will not keep around every single version: just the ones tagged as releases and the most recent as `edge`.

My motivation here is based on a few thoughts: One is that I believe we pay for storage, and while the dependency layers (where most of the size is) should not change often, they may nonetheless accumulate. (I do not know where to find the current usage, so I don't know how to monitor this.) Maybe more importantly, the list of tagged is getting long, and most (all?) of those have never been used explicitly. We tend to use either `edge` or one of the release tagged versions.

Happy to discuss this in more depth, but since the change was so small, I thought a PR made as much sense as an issue.

### DIFF
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -34,8 +34,6 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
-            type=sha
-            type=raw,value={{date 'YYYY-MM-DD'}}
             type=semver,pattern={{raw}}
             type=edge,branch=main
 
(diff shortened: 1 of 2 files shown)


jashapiro:
> We do pay for Docker now, although we may not have added you to the team yet if we didn’t have an immediate use case (we pay per seat). So seems worth asking at this point: should we be switching to Dockerhub?

I don't have any problem with that. I assume the only real need here is to have the proper cred/secret access for this repo. I don't think I would necessarily need personal access.

jaclyn-taroni:
I wouldn't hold this up for now, but something to consider later.

PR #72, opened 2021-11-05
Use renv for docker dependencies & update Bioconductor version

### LINKED ISSUE #49: Use renv to build docker image
Since building the docker image can fail to build when bioconductor updates mid-release (ugh), we may want to use renv to build the docker image here. 

Alternatively, we might switch to the bioconductor docker images? https://www.bioconductor.org/help/docker/ those seem to include rstudio though, which is a bummer.

jashapiro:
Closes #49

The main goal of this PR is to convert the building of the docker image to use `renv`. To that end, I changed the project settings to use `renv`, added an `renv.lock` file, modified the docker scripts, and added a section to the README discussing the changes. 

Since Bioconductor has now updated to the latest version, 3.14, I chose to update the Docker image to R 4.1.1, which will result in that version being applied. I think that will be what we are testing against in CI, as there is no particular version control in the test package (this PR will test that assumption).

This actually broke functions that depend on miQC, but I think that is okay (see below for discussion of fixes).

There are a few details worth noting:

- The `renv.lock` file actually lives in the  `docker` subdirectory, because Docker does not like symlinks that point outside its build environemnt. So instead I put the main file there and symlink to it from the project root, so the `renv` package knows where to find it. 
- I put the `dependencies.R` file in the docker directory, as that seemed the main reason for the file. For example, we use `eisaR` etc. for building indexes, but those aren't part of `scpcaTools` itself at this point. But it seems to make sense to keep them part of the docker image, as we have it now, and they are also useful for development.
- Updating to R 4.1.1 and Bioconductor 3.14 resulted in breaking the miQC test functions, due to a change in the default behavior. Adding the `enforce_left_cutoff = FALSE` option to the section of code where we extract the `prob_compromised` fixed this error, but means that it will not work with prior versions. Because of this incompatibility, I did set the minimum version for `miQC` in the package description.
- While I was modifying the `miQC` functions, I added a column to the output of `scpcaTools::add_miQC()` about whether a cell would pass miQC filtering as implemented by `miQC::filterCells()`. I thought this might be useful down the line, as it  takes advantage of the updating filtering implemented im `miQC` to avoid removing high-gene-count, low-mito-content cells when the model is a bit wonky.

## Questions to reviewers

- Is the use of `renv` clear enough? I wasn't sure how much depth to go into at this point (we might want to move instructions to CONTRIBUTING.md), so what am I missing that we need?
- I did some test builds of the docker image, but it would be good to confirm that it works for others.
- Do the CI tests pass on github? (Not really for a reviewer, unless you call CI a reviewer.) (update: they do!)
- Are the miQC changes okay? Should I leave the `miQC_pass` column or save it for later?


docker/scripts/install_scpca_deps.sh
```diff
@@ -30,51 +30,51 @@ apt-get -y --no-install-recommends install \
...
+#     'scran', \
+#     'scuttle', \
+#     'SingleCellExperiment', \
+#     'SummarizedExperiment', \
+#     'tximport'), \
+#     update = FALSE), \
+#   warning = function(w) stop(w))"
+#
+# rm -rf /tmp/downloaded_packages
+# rm -rf /tmp/Rtmp*
```
allyhawkins, line 80: I think we can go ahead and remove the commented out portion now? Unless there was a reason you were keeping it in there?

R/add_miQC.R
```diff
@@ -12,7 +12,7 @@
 #'
 #' @export
 #'
-add_miQC <- function(sce, seed = NULL){
+add_miQC <- function(sce, posterior_cutoff = 0.7, seed = NULL){
```
allyhawkins, line 15:
```suggestion
add_miQC <- function(sce, posterior_cutoff = 0.75, seed = NULL){
```
I believe this should be 0.75 if we are sticking with the default for the filtering function in miQC. I don't think we really have rationale to not use the default right now. https://bioconductor.org/packages/devel/bioc/vignettes/miQC/inst/doc/miQC.html#33_Extras

README.md
```diff
@@ -19,8 +19,25 @@ The package can be installed from github with:
...
 
 To test the import functions in this package, you must also have installed the [`scpcaData` package](https://github.com/AlexsLemonade/scpcaData).
 This can be done by running `remotes::install_github("AlexsLemonade/scpcaData")`
+
+## Docker image and `renv`
+
+This repository also includes a Dockerfile and associated scripts for building a Docker image that includes the `scpcaTools` package and some additional packages used in other ScPCA repositories, notably [scpca-nf](https://github.com/AlexsLemonade/scpca-nf).
+To support this, we use [`renv`](https://rstudio.github.io/renv/index.html) to track the versioned set of packages that will be installed in that Docker image.
+
+Packages that are required for the main `scpcaTools` package should be included in the `renv.lock` file as they are installed and used.
```
allyhawkins, line 32: Just to be clear if you were to want to add a package to `scpcaTools`, you would need to install it while in the R project and add it to the package using `usethis()` before running `renv::snapshot()`? If so, maybe including a sentence here that explicitly states these steps might be beneficial.

allyhawkins (approved):
I think this looks pretty good. After some annoying Docker and `renv` issues on my side everything worked as it was supposed to when I tested it. I think the instructions overall look good and the only question I had for clarification was expanding on the exact steps you would take when adding a new package to `scpcaTools`. Maybe we don't need to go into that much detail, but it could be helpful to include one more sentence with explicit instructions. 

In terms of the miQC changes, I am okay with adding in the additional column, but I think we should think about whether we want to make the `ccdl_suggests` column and if that will be just a miQC pass column or if it will include additional criteria. I don't know if we necessarily need to have two suggested filtering columns as that might be a lot so let's start with this but it would probably be worth revisiting AlexsLemonade/alsf-scpca#140 now to start deciding how much suggested filtering we want to do on top of miQC and then from there we can modify this column if we feel that's necessary.

README.md
jashapiro: Adding to the package with `usethis()` is not required: `renv` will scan all R code, so it should find things whether or not they are in the DESCRIPTION file. In particular, it will find things for the test functions, whether or not they are required for the main package. Also while all of the packages do need to be installed, they may or may not need to be mentioned specifically: dependencies may not be explicitly mentioned, but will appear in the lockfile. 

So yeah, this is a bit vague, but I'm not quite sure what to do about it, without going into a lot of depth...

allyhawkins: Ah okay. I think the instructions you have are good then, I wouldn't go too much more in detail to avoid the risk of being more confusing.

PR #76, opened 2021-12-02
Make emptyDropsCellRanger default filtering method 

### LINKED ISSUE #73: Update filter_counts to use emptyDropsCellRanger() by default
Currently, our [filter_counts()](https://github.com/AlexsLemonade/scpcaTools/blob/main/R/filter_counts.R) function uses `DropletUtils::emptyDrops()` for the core filtering setup. With the [update to Bioconductor 3.14](https://github.com/AlexsLemonade/scpcaTools/pull/72), we now have the option to use `DropletUtils::emptyDropsCellRanger()`, which may be a bit more robust for single nucleotide data.

We may want to modify `filter_counts` to use the CellRanger version by default, but perhaps leave the original as an option for testing.

allyhawkins:
Closes #73. Here I have modified the filtering function to use `DropletUtils::emtpyDropsCellRanger` as the default filtering method with the option of using `emptyDrops` instead. I added an argument `cr_like` that by default is set to `TRUE` and when set to `FALSE`, filtering will be performed using `emptyDrops`. 

I'm leaving this as a draft PR right now because this does not pass all of the checks with the test functions. In working on this, I noticed that `emptyDropsCellRanger` does not always work as expected. We have a test function that creates a fake sce object and for some reason when trying to apply `emptyDropsCellRanger` to the counts from this fake sce object, I'm getting the following errors: 

``` 
 Error in if (any(still.zero)) { : missing value where TRUE/FALSE needed 
```
This is the same error I was receiving previously when working with `emptyDropsCellRanger()` when there was a bug in the code, which I checked and this version has the [correction in the ambient function that I had noted on their PR previously](https://github.com/MarioniLab/DropletUtils/pull/66#issuecomment-911793467). If I use a real sce (from one of our real samples) or if I create simulated counts using `DropletUtils:::simCounts()` then this works without issue but I can't find any obvious differences between the counts from using `DropletUtils:::simCounts()` and `scpcaTools::sim_sce()`. They are both sparse matrices with genes as the rows and cells as the columns. I think there may be an issue with how we create our simulated sce object that could be causing an issue but I'm not quite sure where that issue is. I'm going to leave this in draft state as I work on that and to see if anyone else has any thoughts on this.

### DIFF
--- a/tests/testthat/test-filter_counts.R
+++ b/tests/testthat/test-filter_counts.R
@@ -1,5 +1,5 @@
 set.seed(1665)
-sce <- sim_sce(n_cells = 100, n_genes = 200, n_empty = 100)
+sce <- sim_sce(n_cells = 200, n_genes = 200, n_empty = 100000)
 
 test_that("Cell filtering with emptyDropsCellRanger() is as expected", {
   filtered_sce <- filter_counts(sce)
--- a/tests/testthat/test-filter_counts.R
+++ b/tests/testthat/test-filter_counts.R
@@ -1,5 +1,5 @@
 set.seed(1665)
-sce <- sim_sce(n_cells = 1000, n_genes = 200, n_empty = 100)
+sce <- sim_sce(n_cells = 100, n_genes = 200, n_empty = 100)
 
 test_that("Cell filtering with emptyDropsCellRanger() is as expected", {
   filtered_sce <- filter_counts(sce)
--- a/tests/testthat/test-filter_counts.R
+++ b/tests/testthat/test-filter_counts.R
@@ -1,8 +1,15 @@
 set.seed(1665)
-sce <- sim_sce(n_cells = 100, n_genes = 200, n_empty = 1000)
+sce <- sim_sce(n_cells = 1000, n_genes = 200, n_empty = 100)
 
-test_that("Cell filtering is as expected", {
+test_that("Cell filtering with emptyDropsCellRanger() is as expected", {
   filtered_sce <- filter_counts(sce)
+  expect_lt(ncol(filtered_sce), ncol(sce))
+  expect_true(all(colnames(filtered_sce) %in% colnames(sce)))
+  expect_equal(nrow(filtered_sce), nrow(sce))
+})
+
+test_that("Cell filtering with emptyDrops() is as expected", {
+  filtered_sce <- filter_counts(sce, cr_like = FALSE)
   expect_lt(ncol(filtered_sce), ncol(sce))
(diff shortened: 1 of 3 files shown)


tests/testthat/test-filter_counts.R
```diff
@@ -1,13 +1,20 @@
 set.seed(1665)
-sce <- sim_sce(n_cells = 100, n_genes = 200, n_empty = 1000)
+sce <- sim_sce(n_cells = 100, n_genes = 200, n_empty = 100)
```
jashapiro, line 2: Did you intend to reduce the # empty? At first I didn't expect it would matter but I would expect a larger number is more realistic for estimation of ambient profile. I am noted that the default for `DropletUtils:::simCounts()` is 10,000 empty cells, which could make a difference in the estimation of the empty cells. 

Update:
Looking at the docs for `[emptyDropsCellRanger](https://rdrr.io/github/MarioniLab/DropletUtils/man/emptyDropsCellRanger.html)`:
>emptyDropsCellRanger takes barcodes whose total count ranks within a certain range - by default, (45,000, 90,000] - to compute the ambient profile.

Even with 10,000 empty droplets, as `simCounts()`, it would seem like there would be not ambient cells by that definition, but I still wonder if that is part of the problem. 

In any even, I would try increasing `n_empty`, and possibly `n_cells` as well to see if that makes a difference. If that doesn't work, you might be able to get somewhere with adding some of `emptyDropsCellRanger` options like `ind.min` and `ind.max` to specify which ranks are "ambient".

jashapiro (commented):
This looks good. 

I suspect the test failures are some combination of the fact that `EmptyDropsCellRanger()` has a lot more parameters and expectations about what the cell counts will look like, and the very small SCE we are making with `sim_sce` is violating those. `DropletUtils::simCounts()` creates more cells by default, and uses a different distribution for generating gene counts. 
With and with some of the parameters of `EmptyDropsCellRanger()` like `umi.min` we may just not meet the minimums. So I might try more cells, more genes, more emptys and see if that works?

tests/testthat/test-filter_counts.R
jashapiro: Confirming my suspicion: I just did a test with `n_empty = 100000` and `emptyDropsCellRanger` worked!

allyhawkins: I had been playing around with both `n_empty` and `n_cells`, but I guess had not gone high enough! I'll update it now so that it has numbers where it works. Thanks for finding this!

jashapiro (approved):
⚡

PR #79, opened 2022-01-10
renv and docker updates for 0.1.2

jashapiro:
In doing some testing to prepare for a new release, I found that we had not installed `optparse` in the docker image, which was predictably breaking things downstream.

In adding that package, I ended up making a few other changes that seemed worth putting in this release, as they should not have major effects (he says with trepidation). 

The main ones are:

* updating renv to 0.15, as that comes with better support for bioconductor packages.
* updating R to 4.1.2 to match the RStudio server version we have installed
* updating all other packages to current versions (this was a bit of an accident, but hopefully will not cause trouble)

Open questions:

* The docker image does not currently have packages for SCE clustering/dimensionality reduction installed (`bluster`, `scater`, etc.). Should we include these in anticipation of future additions to `scpca-nf`?
* Are there other packages that should be included in the Docker image? I checked everything currently used in `scpca-nf` but there might be other uses for the docker image that I didn't anticipate.


allyhawkins (approved):
I think this looks good as is. I don't think we need to include the additional packages, `scater`, `bluster`, etc. because we aren't quite sure what we will need and will probably have to do an update regardless if we miss something or change our minds when we actually get to that point. I think including everything that's in `scpca-nf` should cover our bases and I'm not coming up with another reason outside of that to add additional packages (yet).

PR #81, opened 2022-01-19
Update R-CMD-check workflow

jashapiro:
Tests seem to be failing with some regularity, so I am trying to make them more reliable here, using an updated R test workflow based on https://github.com/r-lib/actions/blob/master/examples/check-standard.yaml


jashapiro:
This doesn't seem to be working. I will try another approach

PR #83, opened 2022-01-20
Remove direct magrittr dependency

jashapiro:
Since we now require R 4.1.0 by way of our dependency on `miQC` 1.2 (and therefore Bioconductor 3.14), I thought it was a reasonable time to strip the dependency on `magrittr` for this package. At least the direct dependency (I am assuming we import some packages that still import `magrittr`, but if we don't import them directly, that dependency may eventually go away as packages update). 

This is maybe as much a proposal as a full PR; it was very simple to implement with a cross-file search and replace, plus a few small additional changes.


allyhawkins (approved):
This all looks good to me.

PR #86, opened 2022-02-07
Round counts for alevin import

### LINKED ISSUE #85: Proposal: round count matrices on import (optionally)
I ran into trouble with Seurat when using an internal function that expected an integer count matrix. With the `crlike_em` quantification of alevin-fry, this assumption was not met. I imagine that there are other places where this issue could occur, and a downstream user might be stumped for a while (as I was!).

At import, we can round count matrices to integer values. This could happen by default, but with an option to turn rounding off. 

We will also need to update [scpca-docs](https://github.com/AlexsLemonade/scpca-docs) to say that we are using rounded EM results if we make this change.

Leaving things as they are... This would require more documentation, and maybe even a warning when non-integer counts were read in.

**Additional context**
There is a potential small loss of precision here, but at the scale & noisiness of single-cell experiments, I don't expect it to be large.

At the other end, an integer matrix is potentially smaller to store (less so with sparse matrices), so there can be some savings here.

jashapiro:
Closes #85 

This PR updates the alevin import function to add an option to round the imported data to integers before creating a `SingleCellExperiment` object. Since I don't think either kallisto or cellranger create non-integer counts, I didn't update those. 

I chose to update only the top-level function that creates the SCE itself (rather than the subfunctions that do the reading from alevin output) in order to preserve precision at the `collapse_intron_counts` step. 

As implemented, this _should_ require no changes to downstream scripts in `scpca-nf`, though we will want to update `scpca-docs` to indicate that all counts are rounded.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: scpcaTools
 Type: Package
 Title: Single cell data tools for ScPCA
-Version: 0.1.2
+Version: 0.1.3
 Authors@R: c(
     person("Ally", "Hawkins",
            email = "ally.hawkins@ccdatalab.org",
--- a/R/read_alevin.R
+++ b/R/read_alevin.R
@@ -12,6 +12,8 @@
 #'   only counts aligned to spliced cDNA ("spliced") or all spliced and unspliced cDNA ("unspliced").
 #'   Applies if `intron_mode` or `usa_mode` is TRUE.
 #'   Default is "spliced".
+#' @param round_counts Logical indicating in the count matrix should be rounded to integers on import.
+#'   Default is TRUE.
 #' @param tech_version Technology or kit used to process library (i.e. 10Xv3, 10Xv3.1).
 #'
 #' @return SingleCellExperiment of unfiltered gene x cell counts matrix.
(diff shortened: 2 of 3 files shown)


allyhawkins (approved):
This looks good to me. The only thing I wonder here is if we want to remove the rounding that happens as part of the [filtering for `emptyDropsCellRanger` or leave it](https://github.com/AlexsLemonade/scpcaTools/blob/6cb2b0dcb75bd7b41353af59a17f5fbe7610aafc/R/filter_counts.R#L40)? I'm not sold on removing it, since it will throw an error there if it's not already rounded though so maybe best to leave it in and to be sure that any matrix that gets input is rounded. Just wanted to point it out here just in case you have other opinions.

jashapiro:
> This looks good to me. The only thing I wonder here is if we want to remove the rounding that happens as part of the [filtering for `emptyDropsCellRanger` or leave it](https://github.com/AlexsLemonade/scpcaTools/blob/6cb2b0dcb75bd7b41353af59a17f5fbe7610aafc/R/filter_counts.R#L40)? I'm not sold on removing it, since it will throw an error there if it's not already rounded though so maybe best to leave it in and to be sure that any matrix that gets input is rounded. Just wanted to point it out here just in case you have other opinions.

I think I agree about leaving it to prevent potential errors, just in case somebody does use an unrounded matrix. I was planning to leave rounding in the seurat functions that were going to need it, for the same reason. I think we should generally strive to check that our functions work with unrounded data, but testing for that will probably be a challenge...

PR #87, opened 2022-02-07
Update release checklist to add version update

jashapiro:
I realized that a critical and important step was missing from the release checklist: make sure the version number of the package gets updated! So I added it here.


jaclyn-taroni (approved):
Linking to `main` makes sense 👍🏻 🚀

PR #91, opened 2022-02-09
Temporarily handle failure for `emptyDrops`

### LINKED ISSUE #90: Handle errors for emptyDropsCellRanger
When processing the Green project again, there was a failure in the new filtering strategy of using `emptyDropsCellRanger` that was causing the entire run to fail. 
Screen Shot 2022-02-08 at 2 30 50 PM
 
When looking at the previous QC report for this sample, it has really low cell counts and also failed miQC so this isn't surprising. I thought we should figure out a way to handle these errors and was back and forth on if we should just not filter at all or if we should filter using `emptyDrops`, as that is much less likely to be unable to find an ambient profile or throw an error like this so wanted to see if anyone else had opinions or thoughts before finishing up the changes to the function. 

We could just not perform filtering in this event and return the unfiltered sce. The question with this is then do we just remove the filtered file altogether (this seems hard and not necessary for just a few samples)? Or do we add a warning for that sample in the QC report and/or metadata that the filtered file is the same as the unfiltered file. I think I'm favoring this approach because if a sample is this poor quality, are we confident in being able to filter it and this seems the most straight forward to explain if we add a warning in the QC report if the unfiltered SCE and filtered SCE are equivalent. 

In testing, using `emptyDrops` is able to identify an ambient profile when `emptyDropsCellRanger` cannot, so we could use `emptyDrops` in the event of failures. We can then add a field to the metadata that has the type of `emptyDrops` filtering that was performed. The downfall of this would be explaining in the docs why there are two different methods, which I don't think is completely straight forward... 

**Additional context**
Tagging @jashapiro and @jaclyn-taroni for any thoughts on this.

allyhawkins:
Closes #90. This PR temporarily handles failures in either `emptyDrops` or `emptyDropsCellRanger` by implementing a hard cutoff for filtering. I am setting a default value for the hard cutoff at a total UMI count of greater than 100. I think this is fairly reasonable, but let me know if others have differing opinions. 

In doing this, I added a parameter for the `umi_cutoff` to the `filter_counts` function which is only invoked in the event of a failure of one of the `DropletUtils` methods. I also added to the metadata of the SCE, the filtering type being used as either `emptyDrops`, `emptyDropsCellRanger`, or `UMI cutoff`. I also added the `umi_cutoff` that was used to the metadata (e.g. 100 as default), in the case of that being the method used. 

For the QC report, I added in a line in the `Cell Statistics` table that has the filtering method that was used. In doing this, I chose to change the line in the `Pre processing information` that previously had `Filtering method` to be `Alevin fry droplet detection`. This refers to using the unfiltered permit list or using a knee method. I was unsure on this name but thought filtering method didn't fit with how we describe it in the docs, since we talk about filtering more in the context of `emptyDrops`. 

I also added a warning statement below that table that if the filtering method is `UMI cutoff` to let the user know that there may be a low number of droplets. Are we okay with this? or would we prefer a different warning message or no message at all? 

Finally, I adjusted the test functions slightly to test the different types of filtering we can do in `filter_counts` and in order to get the test for generating the QC report to pass, I needed to actually perform filtering using `filter_counts` as that adds in the metadata that we now need to access to add into the QC report. 

**Next Steps:**
In response to these changes we will also need to update the documentation and incorporate the filtering type to the `metadata.json` file that is produced as part of `scpca-nf`. If we are okay with this approach then I will go ahead and file those issues. 
I can imagine that we will want to further tweak this, but I think this general approach will solve the problem of failing because of filtering until we can explore this issue further. 

I've included an example of what the new QC report looks like for a sample that is unable to pass filtering and would have to be filtered with a hard cutoff. 
[SCPCL000018_qc_report.html.zip](https://github.com/AlexsLemonade/scpcaTools/files/8036543/SCPCL000018_qc_report.html.zip)

### DIFF
--- a/R/filter_counts.R
+++ b/R/filter_counts.R
@@ -37,9 +37,20 @@ filter_counts <- function(sce, cr_like = TRUE, fdr_cutoff = 0.01, seed = NULL, .
 
   # calculate probability of being an empty droplet
   if(cr_like){
-    empty_df <- DropletUtils::emptyDropsCellRanger(m = round(counts(sce)), ...)
+    empty_df <- tryCatch(
+      DropletUtils::emptyDropsCellRanger(m = round(counts(sce)), ...),
+      error = function(x){NULL}
+    )
+    # if emptyDropsCellRanger fails, do regular emptyDrops and add filtering method to metadata
+    if(is.null(empty_df)){
+      empty_df <- DropletUtils::emptyDrops(counts(sce), ...)
+      metadata(sce)$emptyDrops_method <- "emptyDrops"
+    } else {
+      metadata(sce)$emptyDrops_method <- "emptyDropsCellRanger"
+    }
   } else {
     empty_df <- DropletUtils::emptyDrops(counts(sce), ...)
+    metadata(sce)$emptyDrops_method <- "emptyDrops"
   }
 
   if(any(empty_df$FDR > fdr_cutoff & empty_df$Limited, na.rm = TRUE)){
--- a/R/filter_counts.R
+++ b/R/filter_counts.R
@@ -37,8 +37,8 @@ filter_counts <- function(sce, cr_like = TRUE, fdr_cutoff = 0.01, seed = NULL, u
     stop("cr_like must be set as TRUE or FALSE")
   }
 
-  if(!is.numeric(umi_cutoff) & umi_cutoff >= 0){
-    stop("umi_cutoff must be an integer greater than or equal to 0")
+  if(!is.numeric(umi_cutoff) | umi_cutoff < 0){
+    stop("umi_cutoff must be a number greater than or equal to 0")
   }
 
 
--- a/R/filter_counts.R
+++ b/R/filter_counts.R
@@ -11,7 +11,7 @@
 #' @param fdr_cutoff FDR cutoff to use for DropletUtils::emptyDropsCellRanger or DropletUtils::emptyDrops.
 #'   Default is 0.01.
 #' @param seed An optional random seed for reproducibility.
-#' @param umi_cutoff The minimum UMI count for cells to pass filtering, only used if emptyDropsCellRanger and emptyDrops fails.
+#' @param umi_cutoff The minimum UMI count for cells to pass filtering, only used if emptyDropsCellRanger or emptyDrops fails.
 #'   Default is 100.
 #' @param ... Any arguments to be passed into DropletUtils::emptyDropsCellRanger or DropletUtils::emptyDrops.
 #'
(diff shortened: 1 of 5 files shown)


R/filter_counts.R
```diff
@@ -23,7 +25,7 @@
 #' \dontrun{
 #' filter_counts(sce = sce_object)
 #' }
-filter_counts <- function(sce, cr_like = TRUE, fdr_cutoff = 0.01, seed = NULL, ...) {
+filter_counts <- function(sce, cr_like = TRUE, fdr_cutoff = 0.01, seed = NULL, umi_cutoff= 100, ...) {
```
jashapiro, line 28:
```suggestion
filter_counts <- function(sce, cr_like = TRUE, fdr_cutoff = 0.01, seed = NULL, umi_cutoff = 100, ...) {
```

R/filter_counts.R
```diff
@@ -35,18 +37,39 @@ filter_counts <- function(sce, cr_like = TRUE, fdr_cutoff = 0.01, seed = NULL, .
...
+      error = function(x){NULL}
+    )
+    metadata(sce)$filtering_method <- "emptyDropsCellRanger"
   } else {
-    empty_df <- DropletUtils::emptyDrops(counts(sce), ...)
+    empty_df <- tryCatch(
+      DropletUtils::emptyDrops(counts(sce), ...),
+      error = function(x){NULL}
+    )
+    metadata(sce)$filtering_method <- "emptyDrops"
```
jashapiro, line 56: I don't love the repetition here. I have some ideas for how to clean this up, but I can't do it as a comment. I will file a PR to this branch with my ideas.

R/filter_counts.R
```diff
@@ -35,18 +37,39 @@ filter_counts <- function(sce, cr_like = TRUE, fdr_cutoff = 0.01, seed = NULL, .
...
+  # if emptyDrops doesn't fail, filter by FDR cutoff
+  if(!is.null(empty_df)){
+    if(any(empty_df$FDR > fdr_cutoff & empty_df$Limited, na.rm = TRUE)){
+      warning(glue::glue("`niters` may be set too low for emptyDrops filtering.",
+                         " Current value is {empty_df@metadata$niters}."))
+    }
+    cells <- rownames(empty_df)[which(empty_df$FDR <= fdr_cutoff)]
+  } else {
+    # if emptyDrops fails, filter using a hard UMI threshold
+    cells <- sce$sum > umi_cutoff
```
jashapiro, line 68: I think `>=` aligns better with the statement in the docs ("minimum UMI count"), and also makes more intuitive sense to me.

```suggestion
    cells <- sce$sum >= umi_cutoff
```

jashapiro (commented):
This looks good. I have a few small suggestions which are in the comments below.  The biggest one is a change in the warning format, to make it a bit more prominent and visually appealing.

As the filter_counts function got more complicated though, I thought it would benefit from a bit of streamlining/refactoring, which I made an attempt at in a derived branch. I will file a PR to this branch with those changes shortly.

allyhawkins:
Thanks for taking a look at this and for the refactoring help @jashapiro. I went ahead and incorporated those changes here. I had struggled trying to find a less repetitive way of doing this, so I like this a lot better and I really like the warning statement changes too!

jashapiro (approved):
LGTM, with one logic change suggested below.

PR #100, opened 2022-03-29
Make Seurat a bit more robust and fail more elegantly

### LINKED ISSUE #99: Seurat HTOdemux fails if unmeasured HTOs are included
`Seurat::HTODemux()` sometimes fails with the following error: 
```
Error in Seurat::HTODemux(seurat_obj, assay = "HTODemux") : 
  Cells with zero counts exist as a cluster.
```

This seems to occur mostly when trying to run `HTODemux` when there are oligos included in the SCE that are not measured (whether by design or because cells with those tags were not detected). We should be able to handle the design question by removing unexpected HTOs, but I would like to see if we can make it more robust to the case where tags were not detected for some other reason.

jashapiro:
This PR should make calling the Seurat `HTODemux()` function more robust, particularly with unfiltered or loosely filtered data. In particular, it removes cellhash HTOs that are found very rarely  before performing calling, and also removes any lingering cells that might have no reads. I picked a threshold of < 0.1% of cells with counts greater than 1 for a each HTO, which I found in a few real data sets seemed to do a good job of eliminating spurious mappings in the data sets that I was able to play with. I considered using a higher threshold, which might make it even more robust, but I didn't want to get rid of HTOs that might correspond to real but rare cells in the multiplexed set. 

In the event that `HTODemux()` still fails, I still add a column for the results in the main `colData` object, but with `NA` values rather than real results. I don't include any results in the `altExp`.

closes #99 (until the next place Seurat breaks)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: scpcaTools
 Type: Package
 Title: Single cell data tools for ScPCA
-Version: 0.1.4
+Version: 0.1.5
 Authors@R: c(
     person("Ally", "Hawkins",
            email = "ally.hawkins@ccdatalab.org",
--- a/R/add_cellhash_calls.R
+++ b/R/add_cellhash_calls.R
@@ -123,7 +123,7 @@ add_demux_hashedDrops <- function(sce, altexp_id = "cellhash", ...){
   if(is.null(sample_ids)){
     sample_ids <- rowData(altExp(sce, altexp_id))$barcode_id
   }
-  if (is.null(sample_ids)){
+  if(is.null(sample_ids)){
     warning("No sample ids are present for demux results, using row names")
     sample_ids <- rownames(altExp(sce, altexp_id))
   }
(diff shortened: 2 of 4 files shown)


jashapiro:
> This looks good to me. I am fine with the threshold that you chose for filtering, as I don't think we have strong rationale for picking a number that's higher than that and risking losing potentially rare cells as you stated.
> 
> My only comment is about replacing the failed results with `NA` and also replacing the doublets and negative results with `NA`. I wonder if it would make more sense to put `Demux_fail` or something that indicates it was a failure of the actual function to demultiplex. I know you will be able to distinguish because an entire column will either be `NA` versus only part of the column, but I think it still might be useful to somehow point out that this library failed in a more direct way.

I can't really think of a better solution here, other than leaving out the column completely, which I didn't want to do, as I wanted some indication that the demux was attempted. Using NA for all values seemed the most straightforward, as it shouldn't immediately break any workflows that might expect particular values.

PR #102, opened 2022-05-09
Handle miQC filtering errors

### LINKED ISSUE #101: [BUG] miQC fails to filter sometimes
**Describe the bug**
`add_miQC` sometimes fails when the `miQC::filterCells()` step fails 

**To Reproduce**

The error is somewhat stochastic, making it a bit hard to track down. We can reproduce failure somewhat frequently with the file `SCPCL000161_unfiltered.rds`, but it does depend on the seed used.

The failure does seem to occur only when `enforce_left_cutoff = TRUE`, but we would like to keep that option. 

**Expected behavior**
miQC info should be added!

**Version Information**
Failure occurs using in the Docker image exported as part of this package.

**Additional context**

This error could possibly be handled downstream, but it seems better to try to fix it in the `add_miQC` function if possible.

jashapiro:
After some debugging, I found that the errors we were getting were occurring when the `miQC` model fitting resulted in two components that were very similar (almost 1 component). The result of this was almost all cells getting probability compromised of ~ 0.5, so they all "pass", which causes trouble for the `enforce_left_cutoff = TRUE` code path. 

fail_fit

So here I am fixing it by checking whether `miQC::filterCells()` works before proceeding much past building the model. If it fails, we try again a couple times before continuing.

I am currently setting `miQC_pass` to NA in the case where we have a model but it failed to filter properly. This is a bit different from the behavior when the model totally failed and we set prob_compromised to NA and leave `miQC_pass` and `miQC_model` off entirely. I almost think I should do the same thing in this case as this really is a kind of model failure, so we might not want to report the bad model at all. Let me know what you think.

closes #101

### DIFF
--- a/R/add_miQC.R
+++ b/R/add_miQC.R
@@ -33,29 +33,37 @@ add_miQC <- function(sce, posterior_cutoff = 0.75, seed = NULL){
   # set seed
   set.seed(seed)
 
+  model <- NA
+  pass_cells <- NA
+  model_attempt <- 0
   # generate linear mixture model of probability of cells being compromised
-  model <- tryCatch(
-    miQC::mixtureModel(sce),
-    error = function(x){NA}
-  )
-  if (is(model, "flexmix")  && length(model@components) < 2){
-    # first model fit gave only one component, give it one more chance.
-    model <- miQC::mixtureModel(sce)
+  while(model_attempt < 3 &&
+        (!is(model, "flexmix") || length(model@components) < 2 ) &&
+        is.na(pass_cells)){
+    model_attempt <- model_attempt + 1
+    try({
+      model <- miQC::mixtureModel(sce)
+      # filter step can fail, so we will try this too, keeping the passing cell ids
+      pass_cells <- colnames(
+        miQC::filterCells(sce,
+                          model = model,
+                          posterior_cutoff = posterior_cutoff,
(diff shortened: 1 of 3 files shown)


R/add_miQC.R
```diff
@@ -34,29 +34,43 @@ add_miQC <- function(sce, posterior_cutoff = 0.75, seed = NULL){
...
+    model_attempt <- model_attempt + 1
+    try({
+      model <- miQC::mixtureModel(sce)
+      # filter step can fail, so we will try this too, keeping the passing cell ids
+      pass_cells <- colnames(
+        miQC::filterCells(sce,
+                          model = model,
+                          posterior_cutoff = posterior_cutoff,
+                          verbose = FALSE)
+      )
```
allyhawkins, line 54: Can you check the indentation on this? It seems funny here in that line 43 is more indented than line 45 which doesn't seem correct to me?

allyhawkins (approved):
This looks pretty good. I just had one comment about fixing indentation that seemed off to me that should probably be fixed before merging.

R/add_miQC.R
jashapiro: It does look strange, but 43 is indented for the `while` conditions, and 45 is the inside of the `while` block, so they are correct by the rules I was following.

PR #110, opened 2022-06-02
Change filename used for child multiplex QC report

allyhawkins:
In testing the latest changes in `scpca-nf`, I ran into an error where the filename had an extension of `.Rmd` rather than `.rmd` so it couldn't find the child template notebook to add in. I tried to change the filename, rather than change it to use `.Rmd`, but github wasn't noticing that as a change and letting me commit that... So here I'm updating the name of the template report. 

Here's a screenshot of the error I received: 
Screen Shot 2022-06-02 at 4 09 01 PM

### DIFF
--- a/inst/rmd/qc_report.rmd
+++ b/inst/rmd/qc_report.rmd
@@ -357,7 +357,7 @@ In such situations, the calculated probability of compromise may not be valid (s
 ```
 
 <!-- Next section only included if multiplex data is present --> 
-```{r, child='multiplex_qc.rmd', eval = has_multiplex}
+```{r, child='multiplex_qc.Rmd', eval = has_multiplex}
 
 ```
 


jashapiro:
Because `.rmd` is the extension we are using elsewhere, I'd rather leave the code and move the file. As you saw, git won't recognize case-only changes (on a case-insensitive file system like the mac), but you can get around this by telling git to do the renaming itself:

```
git mv inst/rmd/multiplex_qc.Rmd inst/rmd/multiplex_qc.rmd
```

allyhawkins:
Thanks for the suggestion @jashapiro. I reverted the change back to use the filename `multiplex_qc.rmd` and changed the file using the command you recommended so the only change in this PR is the file itself changing names.

jashapiro (approved):
⚡

PR #111, opened 2022-06-07
Remove warning for multiplexed samples with only one HTO 

allyhawkins:
In running the multiplexed samples, there was one library that contains cellhashing data, but only has one HTO and only one sample. Because this only contains one sample, we plan on skipping demultiplexing alltogether (otherwise Seurat will fail). Additionally, because this only has one sample, when creating the `metadata.json` file as part of `scpca-nf` the `has_multiplex` flag will get reported as false because it does not have more than one sample (the changes to skip demultiplexing will get filed as a separate PR in `scpca-nf`). It would be confusing to have it be false in the metadata file and then have a warning produced here that said "This library is multiplexed and contains data from more than one sample." However, because this sample still has cellhashing data, it would be helpful to report those statistics. 

I have gone ahead and separated out the check for if there is cell hashing data present in the object and then the check for if the data is multiplexed (i.e. has more than one sample). If the library has more than one sample then the warning will get printed, otherwise it will get skipped. The rest of the report remains the same. 

I attached a copy of the updated report for this specific library.
[SCPCL000533_qc_report.html.zip](https://github.com/AlexsLemonade/scpcaTools/files/8855500/SCPCL000533_qc_report.html.zip)

### DIFF
--- a/inst/rmd/qc_report.rmd
+++ b/inst/rmd/qc_report.rmd
@@ -81,7 +81,7 @@ if (has_cite){
 }
 
 # only print out multiplexed warning if more than one sample is present
-has_multiplex = if (length(sample_id) > 1){TRUE} else {FALSE}
+has_multiplex <- length(sample_id) > 1
 if(has_multiplex){
   # convert sample ID to bullet separated list 
   multiplex_samples <- paste0("<li>", paste(sample_id, collapse = '</li><li>', "</li>"))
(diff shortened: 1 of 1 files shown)


allyhawkins:
Thanks for catching that! I went ahead and combined everything into one check for `has_multiplex` to create the sample list to print and then print out the warning all in one block rather than have it separated out into two.

jashapiro (approved):
LGTM!

PR #114, opened 2022-07-05
Conversion of SCE objects to AnnData

### LINKED ISSUE #106: Export SCE objects to AnnData

We currently store SCE RDS objects as our principal data structure, but to use python based tools (notably [scverse](https://scverse.org/) tools like scanpy or [scvi](https://docs.scvi-tools.org/en/stable/index.html), we will often need [anndata](https://anndata.readthedocs.io/en/latest/) or related structures.

While we have [instructions for conversion in our docs](https://scpca.readthedocs.io/en/latest/getting_started.html#what-if-i-want-to-use-scanpy), those only cover the minimal use case of converting the data matrix to another format.

As we add components to the SCE object like dimensionality reduction matrices & other modalities, it would be nice to have encapsulated functions that convert all (or most) parts of the SCE object to anndata format that we can reuse across analysis projects, including downstream analysis and integration.

The important part here is that the function should convert as many data and metadata components as possible between the two formats.

This may also enable alternative download formats for ScPCA portal in the future if we choose to go that direction.

In some previous work, we have seen that writing anndata to hdf5 can be slow using the existing tools. What is somewhat unclear is whether that slowness is inherent in the hdf5 format or a limitation of some of the R/python communication that was required. If we can mitigate the slow performance by writing out a bunch of matrices and then reading them in entirely within python, that might be worth thinking about. 

**Additional context**

There are further instructions in https://theislab.github.io/scanpy-in-R/; I anticipate we would want to use the `reticulate` version, possibly with exploring the `renv` python environment management.  

Alternatively, we might want to look again at [zellkonvertor](https://bioconductor.org/packages/release/bioc/html/zellkonverter.html) which should do some of the heavy lifting including using [basilisk](https://bioconductor.org/packages/release/bioc/html/basilisk.html) to handle the python dependencies.

allyhawkins:
Closes #106. 

Here I'm initiating a function to convert SCE objects to AnnData objects and hdf5 files that can be read directly into python. The approach that I took was to use the [zellkonverter](https://theislab.github.io/zellkonverter/reference/AnnData-Conversion.html) package with [basilisk](https://bioconductor.org/packages/3.15/bioc/html/basilisk.html). 

I mostly followed the [section on converting between SCE and AnnData in the zellkonverter documentation](https://theislab.github.io/zellkonverter/articles/zellkonverter.html#converting-between-singlecellexperiment-and-anndata-objects). I'm using the provided `zellkonverterAnnDataEnv` and running it in basilisk which will load the conda environment with the specific dependencies and packages needed to use zellkonverter. Within the function I ran `basilisk::basiliskRun` and then created a function that converts the object and then writes the object to an hdf5 file. In testing, I also tested that the hdf5 file that is created was able to be read into python and has an AnnData object that you would expect. 

Using this was the most straight forward way I could identify to convert directly to AnnData and maintain most of the information still present in the original SCE. The returned AnnData object will have the counts data stored as `X`, which is what is expected in AnnData and then stores any additional data, such as the `logcounts` data if present as a layer that can be accessed with anndata.layers["logcounts"]. It also converted the `colData`, `rowData`, and `metadata` from the original object. The only issue that I noticed when converting was the miQC model that we store in the metadata. That doesn't have a python format that is equivalent that can be stored, so I decided to remove the miQC model if present and print out a warning that it won't be included in the conversion. Is there anything else I should be checking for to make sure that it gets converted successfully? 

I will also note that there were warnings provided by zellkonverter if there was something that was present in the SCE object that could not be converted, which I found comforting. 

The last thing I did was start a test function for making sure that the function works correctly. I added a check to make sure that the function runs and produces the expected anndata object by using [`expect_snapshot`](https://testthat.r-lib.org/articles/snapshotting.html). This returns a markdown with the output from running the code which will contain an anndata object and no warning. All of the tests then compare the output to each other to make sure it's the same. We could check explicitly that the contents are the same between SCE and AnnData, but there are slight differences in how the data is stored and what format. For example the type of data frame is different for the `colData` and some of the data gets converted to factors so we would have to make sure we accounted for all of that and didn't miss it to ensure that tests passed. Instead I decided to check for the column names of the `colData` and `rowData` to provide some preliminary checks that the object looks like it should. Are there any other checks we should have right now? 

The other approaches I played around with were constructing the SCE object directly using the [`anndata` package in R](https://anndata.dynverse.org/index.html) or just using `reticulate` directly following the [scanpy directions](https://theislab.github.io/scanpy-in-R/#converting-from-r-to-python-1). However, the zellkonverter approach was working well and doing a lot of the work for us so I wanted to try that first and get some initial thought before proceeding further. 

Zellkonverter and basilisk did print out a bunch of warnings to increase the version of `renv` to 0.15.5 so that is included here, but let me know if we don't want to go that route for any reason. 

The last thing to note is that this doesn't consider CITE-seq and the counts data for CITE-seq will get lost in the current conversion. However we should be able to add them in to the object, I just thought that was for another PR and right now we could focus on the initial conversion itself.


jashapiro:
I've started to look through this, and my first question is whether you tested [zellkonverter::writeH5AD()](https://theislab.github.io/zellkonverter/reference/writeH5AD.html) as a potentially simpler approach. You would not get the `reticulate` object out, but I'm not sure whether we would use that at all, aside from maybe testing?

I'm wondering this now partly because I'm curious if it might work better in the test environment. We may need more to figure out what test environment we really need though.

DESCRIPTION
```diff
@@ -22,6 +22,7 @@ RoxygenNote: 7.2.0
 Depends: 
     R (>= 4.1.0)
 Imports: 
+    basilisk,
```
jashapiro, line 25: Can we make these new imports `Suggests` rather than `Imports`?

If we do that we will need to add checks to the function with `requireNamespace()` to fail elegantly, but I think that is probably the way we should try to go.

R/sce_to_anndata.R
```diff
@@ -0,0 +1,55 @@
+
+#' Convert SingleCellExperiment objects to AnnData file stored as HDF5 file
+#'
+#' @param sce SingleCellExperiment object to be converted to AnnData
+#' @param anndata_file Name of
+#'
+#' @return Converted AnnData object
+#'
+#' @import SingleCellExperiment
+#' @import reticulate
```
jashapiro, line 10: I'm guessing this is required because we use the `rna_adata$write()` method later? I'd rather avoid it if we can, because importing means it has to be part of the package, I think, and we can't use it as a `Suggests`, which is kind of annoying.

jashapiro (commented):
This looks like a good start, and I confirmed that it passed tests on my local machine, so the mystery of why it fails in CI remains. I have not tried to test it on an ubuntu machine, but that is probably the next step.  I think that it probably take a bit of time to sort out, unfortunately. It might require a bigger change in the way we do the CI & dependency installation to fix this and the Seurat test failures that I was just skipping, but I haven't thought too much about it as yet...

allyhawkins:
So originally I wasn't using `zellkonverter::writeH5AD()` because I really wanted to be able to find a way to check and make sure that the object was being created correctly and by using basilisk and `SCE2AnnData` I was able to kind of do that. However, the checks looked like they were failing because of issues with using the python environment and calling the environment that zellkonverter uses. I went through a series of troubleshooting to create a conda environment and package it in renv and then also created our own basilisk environment object within the function using the conda environment available in the package, but conda still stores everything with full paths which I _think_ could be part of the reason that there were no failures locally but there are with CI. I also wasn't quite sure if that was even the reason I just saw that it was failing because basilisk couldn't find the environment. 

So I ended up going back to the much more straight forward approach of just directly writing the HDF5 file which works and stores the AnnData object as we would want it, other than needing to remove the miQC model. This also meant removing the checks that compared the AnnData object to the SCE, but we can still check the file creation and that the function doesn't have errors which I think is a good start for now and there's probably room for improvement. I also put `zellkonverter` into `suggests` and removed the basilisk and reticulate packages all together.

tests/testthat/test-sce_to_anndata.R
```diff
@@ -0,0 +1,15 @@
+set.seed(1665)
+sce <- sim_sce(n_cells = 100, n_genes = 200, n_empty = 0)
+colData(sce) <- DataFrame("test_column" = sample(0:10, 100, rep= TRUE))
+rowData(sce) <- DataFrame("test_row" = sample(0:10, 200, rep = TRUE))
+
+test_that("Conversion of SCE to AnnData works as expected", {
+  anndata_file <- "test_anndata.h5"
+
+  # test that the H5 file is created
+  expect_snapshot(sce_to_anndata(sce, anndata_file))
```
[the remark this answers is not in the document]
allyhawkins: I had thought about this too. When you read it back in it creates an SCE object again, but that might be okay to at least test that the dimensions are correct and col/row data is maintained like I had been doing.

jashapiro (approved):
Looks good! Just one little comment on the docs and a suggestion to use `NULL`. I have a tiny fear that the `NULL` will cause trouble in the conversion, so it is probably worth testing that (the test data does not include a miQC model), but I expect it to work. 

Otherwise, I think this looks good! 

The one remaining thing that you mentioned in your initial comments was the handling of alt experiments. I agree that it is correct to make that a separate issue, but I wanted to put some thoughts here while I am thinking about it. It looks to me that AnnData can't really handle those at this point except by adding the additional measures as "genes" and then annotating them in the metadata. I don't love that solution, as I think it puts an odd burden on the user to know that the matrix has other data. I think the best solution for now may be to simply output multiple AnnData objects, one for each modality. There is the [`MuData` class](https://github.com/scverse/muon) which wraps up multiple AnnData objects, but I am not sure how easy the support within R will be, or whether that would really be worth the effort.

PR #116, opened 2022-07-12
Convert SCE to Seurat

### LINKED ISSUE #107: Function to convert SCE objects to Seurat

We currently store SCE RDS objects as our principal data structure, but to use [Seurat](https://satijalab.org/seurat/) and Seurat-based tools we will often need the Seurat data structures structures.

While we have [instructions for conversion in our docs](https://scpca.readthedocs.io/en/latest/faq.html#what-if-i-want-to-use-seurat-instead-of-bioconductor), those only cover the minimal use case of converting the data matrix to another format.

As we add components to the SCE object like dimensionality reduction matrices & other modalities, it would be nice to have encapsulated functions that convert all (or most) parts of the SCE object to Seurat format that we can reuse across analysis projects, including downstream analysis and integration.

The important part here is that the function should convert as many data and metadata components as possible between the two formats.

This may also enable alternative download formats for ScPCA portal in the future if we choose to go that direction.

**Additional context**

We do some conversion in the [add_demux_seurat()](https://github.com/AlexsLemonade/scpcaTools/blob/f869f6f1efcd62f89b939f11bbb19f544e324fa7/R/add_cellhash_calls.R#L159-L260)  function, so we might off of that approach for this conversion, as it did catch some edge cases we might need to account for.

allyhawkins:
Closes #107. 

This PR adds a function to convert SCE objects to Seurat objects including the addition of any metadata or column/row data that is in the original SCE object. One important thing to note is that I originally tried to construct the Seurat object with the addition of the rowdata and metadata during the `CreateSeuratObject`, but that doesn't work. Seurat will create the object and give you no warning or error and leave the `var.features` and `misc` slot empty, unless you directly add it later on, so that's what I ended up doing. 

In doing this I also accommodated for the potential presence of any number of `altExp` assays that could be found in the original SCE. I incorporated some of the catches that were used in `add_demux_seurat`, subsetting cells to be those without zero counts before adding the new assay. I also added a check that the assay was added in case there was an issue with converting it so that a warning would be created. I'm not sure if this is entirely necessary. 

I included a test function here as well where I simulated creating an SCE with an `altExp` and checked the the object appeared as it should. Keep in mind that the `colnames` will not be exactly the same if there are 0 calls and that the `meta.data` in Seurat will have additional columns besides what was originally just in the `colData(sce)`. Should I include other checks? Do we need one for an SCE without an `altExp`? Are there other checks that are missing from the function itself? 

Also this is passing my local checks, but I'm unsure if it will pass GHA. I did put in the skip for checking if Seurat was installed, but I didn't need that to pass checks locally.


R/sce_to_seurat.R
```diff
@@ -0,0 +1,82 @@
...
+  rowdata <- as.data.frame(rowData(sce))
+
+
+  # create seurat object
+  seurat_obj <- Seurat::CreateSeuratObject(counts = sce_counts,
+                                           meta.data = coldata)
+
+  # add rowdata and metadata separately adding it while creating leaves the slots empty without warning
+  seurat_obj[["RNA"]]@var.features <- rowdata
+  seurat_obj@misc <- metadata(sce)
```
jashapiro, line 44: Just wondering if we want to remove the model here. I don't think we would need it in any Seurat application.

R/sce_to_seurat.R
```diff
@@ -0,0 +1,82 @@
...
+    seurat_obj_cells <- colnames(seurat_obj)
+
+    # round counts and calculate total counts
+    alt_counts <- round(counts(altExp(sce, name)))
+    alt_sum <- Matrix::colSums(alt_counts)
+
+    # identify cells with > 0 counts and intersect with cells already in seurat
+    alt_seurat_cells <- names(alt_sum)[alt_sum > 0]
+    cells_to_keep <- intersect(seurat_obj_cells, seurat_obj_cells)
+    alt_counts <- alt_counts[, cells_to_keep]
```
jashapiro, line 63: I'm not sure if the alt assays are required to have non-zero counts. (I also think this could have been a bit of a red herring when I was doing this conversion in earlier steps, so the whole thing could be unnecessary.)

Either way, a little bug (which may suggest that the whole thing is unnecessary, as all alt cells are being kept at the moment)
```suggestion
    # identify cells with > 0 counts and intersect with cells already in seurat
    alt_seurat_cells <- names(alt_sum)[alt_sum > 0]
    cells_to_keep <- intersect(seurat_obj_cells, alt_seurat_cells)
    alt_counts <- alt_counts[, cells_to_keep]
```

R/sce_to_seurat.R
```diff
@@ -0,0 +1,82 @@
...
+
+    # subset seurat object to only have cells that will be in alt_counts before adding new assay
+    seurat_obj <- seurat_obj[, cells_to_keep]
+
+    # add new assay along with associated row data, if any
+    rowdata <- as.data.frame(rowData(altExp(sce, name)))
+    seurat_obj[[name]] <- Seurat::CreateAssayObject(counts = alt_counts)
+    seurat_obj[[name]]@var.features <- rowdata
```
jashapiro, line 71: Do we want this to be `rowData(altExp(sce, name))`?

tests/testthat/test-sce_to_seurat.R
```diff
@@ -0,0 +1,49 @@
...
+
+  if (!requireNamespace("Seurat", quietly = TRUE)) {
+    skip("The Seurat package was not installed. Skipping tests")
+  }
+
+  seurat_object <- sce_to_seurat(sce)
+
+  # check that column names of Seurat object are derived from SCE object
+  # they won't necessarily be equal if some cells with 0 counts were removed
+  expect_true(any(colnames(seurat_object) %in% colnames(sce)))
```
jashapiro, line 33: The other way around we might not expect a full match, but all of the seurat cells should be in the sce, right?
```suggestion
  expect_true(all(colnames(seurat_object) %in% colnames(sce)))
```

tests/testthat/test-sce_to_seurat.R
```diff
@@ -0,0 +1,49 @@
...
+
+  # check that attached metadata/coldata/rowdata in SCE are present in seurat object
+  coldata_sce <- as.data.frame(colData(sce))
+  rowdata_sce <- as.data.frame(rowData(sce))
+
+  # can't directly check that coldata is equal because of added columns in seurat object
+  expect_true(any(colnames(coldata_sce) %in% colnames(seurat_object@meta.data)))
```
jashapiro, line 40: Here we expect all of the sce columns to end up in the seurat object, right? Unless some are changing names?
```suggestion
  expect_true(all(colnames(coldata_sce) %in% colnames(seurat_object@meta.data)))
```

tests/testthat/test-sce_to_seurat.R
```diff
@@ -0,0 +1,49 @@
...
+  expect_equal(rowdata_sce, seurat_object[["RNA"]]@var.features)
+  expect_equal(metadata(sce), seurat_object@misc)
+
+  # check that altExp data was converted and rowData
+  expect_false(is.null(seurat_object[[alt_name]]))
```
jashapiro, line 45: We can make this test a bit more explicit... we might want to add similar tests to other places, like to test that the created object is in the expected assay?
```suggestion
  expect_s4_class(seurat_object[[alt_name]], 'Assay')
```

jashapiro (commented):
This looks good, with a few small thoughts and suggestions below. 

I'm curious if you tried testing it without the test-skipping line? The error I was getting was quite unexpected to me, and I always wondered if it might go away. So this might be a good time to see if it does.

allyhawkins:
@jashapiro thanks for taking a look at this. Based on your comments I went ahead and removed the check for non-zero counts before creating the object and things seemed to work as expected. Did you want me to keep the one for the RNA counts and just remove the one for `altExp` counts? I don't think we need either since I've never run into that issue before, but wanted to see if you had seen other things when you were working with the cell hashing data. 

I also removed the miQC model as you had suggested. 

I did try to remove the test skipping but those [checks failed](https://github.com/AlexsLemonade/scpcaTools/runs/7330033891?check_suite_focus=true). I wasn't sure if the changes you are making in https://github.com/AlexsLemonade/scpcaTools/pull/117 are going to help with that or if I should continue to leave the test skipping in?

jashapiro:
> @jashapiro thanks for taking a look at this. Based on your comments I went ahead and removed the check for non-zero counts before creating the object and things seemed to work as expected. Did you want me to keep the one for the RNA counts and just remove the one for `altExp` counts? I don't think we need either since I've never run into that issue before, but wanted to see if you had seen other things when you were working with the cell hashing data.

I think I saw that when working with unfiltered cells, so it may not be relevant here, but I think it makes sense to leave in. There shouldn't be any zero counts in these data, but in case someone tries to use this function with an unfiltered dataset it might save a headache, so I would leave it.

I don't know if #117 will fix the Seurat issue either, though that was part of my goal there. I think it is fine to leave the test skipping for now, and we can remove it later if we figure anything out on that front.

PR #117, opened 2022-07-13
Update R tests

jashapiro:
Since we started development, the R actions have evolved, so I wanted to see if the new ones could do what we needed with less code and more reliability.  This PR is to test that!

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -23,12 +23,6 @@ jobs:
       - name: Remove renv
         run: rm -f renv.lock .Rprofile
       - uses: r-lib/actions/setup-pandoc@v2
-      # - name: Install system dependencies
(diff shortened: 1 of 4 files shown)


jashapiro:
Okay, this seems to work now with a few additional changes, and even the Seurat tests work!

For the new dependency installation, I added a `Remotes:` section to the `DESCRIPTION` which allows the `scpcaData` package to be added automatically.

The other change is adding `sessioninfo` as a suggestion, which we seemed to miss before somehow. I also made the notebook that uses it fall back on the built-in `sessionInfo` if not installed.

Oh, the best benefit? The new dependency installation uses a cache, so it is _much_ faster for repeated tests!

allyhawkins (approved):
🎉 🎉 looks good and yay to faster tests! Thanks so much for doing this!

Issue #119, opened 2022-07-18
Update documentation for `sce_to_anndata` to include note about conversion of string annotations to categoricals


allyhawkins:
Originally noted in https://github.com/AlexsLemonade/sc-data-integration/pull/42#issuecomment-1187703050, the conversion to `AnnData` was converting some of the metadata columns to categorical factors and not all. The reason for this is based on the presence of duplicate gene names or duplicate entries in the column. We should add a note in the documentation for that function about how columns are chosen to be converted.

allyhawkins:
Closed by #120

PR #120, opened 2022-07-21
Change warning for miQC removal to message

### LINKED ISSUE #118: Change miQC removal in SCE conversion to message rather than warning
We should change the warnings for removal of `miQC` in SCE conversion functions (both Seurat and AnnData) to be just messages. 

This is expected... @allyhawkins On reflection, we should probably change it to a message? I don't think it rises to the level of warning.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/sc-data-integration/pull/42#discussion_r922404783_

allyhawkins:
Closes #118 and #119. 

This just has a few small changes. The first is changing the previous warning messages when miQC was removed to be just a message. I also added a small note in the documentation for the SCE to AnnData function about the conversion of rowData to categorical data that was discovered when applying the function in data integration.

### DIFF
--- a/R/sce_to_anndata.R
+++ b/R/sce_to_anndata.R
@@ -5,6 +5,9 @@
 #' @param anndata_file Path to output AnnData file. Must be an `.h5` or `.hdf5`
 #'
 #' @return original SingleCellExperiment object used as input (invisibly)
+#' **Note that any columns present in the `rowData` of an SCE object that contains
+# duplicated information, e.g. duplicate gene identifiers, are converted to
+# categorical data by the `anndata` package.
 #'
 #' @import SingleCellExperiment
 #'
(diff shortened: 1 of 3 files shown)


R/sce_to_anndata.R
```diff
@@ -5,6 +5,9 @@
 #' @param anndata_file Path to output AnnData file. Must be an `.h5` or `.hdf5`
 #'
 #' @return original SingleCellExperiment object used as input (invisibly)
+#' **Note that any columns present in the `rowData` of an SCE object that contains
+# duplicated information, e.g. duplicate gene identifiers, are converted to
+# categorical data by the `anndata` package.
```
jashapiro, line 10: Roxygen ignores comment lines without a single quote, so this change will require a rebuild to update the documentation.
```suggestion
#' duplicated information, e.g. duplicate gene identifiers, are converted to
#' categorical data by the `anndata` package.
```

jashapiro (approved):
It looks like the test failure was transient. You will need to rebuild the docs with the suggested change I made, but otherwise this looks good to me!

R/sce_to_anndata.R
allyhawkins: I was really hoping this was a transient failure. Fingers crossed it works this time. Thanks for taking a look!

PR #125, opened 2022-09-08
Use temp directory for intermediate files when rendering

### LINKED ISSUE #124: Use temp directory for intermediate files.
We will need the report rendering to use a temporary directory for intermediate files, as the package directory that is the source for the template may not be writeable. 

Looking at https://rmarkdown.rstudio.com/docs/reference/render.html it looks like we will need to specify the `intermediate_dir` and the most logical thing will be to make that a temp directory.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/issues/208#issuecomment-1240991545_

jashapiro:
Closes #124

This update is mainly to fix the rendering errors caused by trying to render Rmd template files within the package, which is not such a good idea. So instead we use the R tempdir for the rendering and any intermediate files. 

I am not sure that I needed to specify both `intermediates_dir` and `knit_root_dir`, but it seemed reasonable to do so.

I also took the opportunity to update the Docker image and `renv.lock` file to use R 4.1.3 and current package versions. Everything seems to work fine locally, so I don't expect any major changes.

### DIFF
--- a/R/generate_qc_report.R
+++ b/R/generate_qc_report.R
@@ -6,6 +6,7 @@
 #' @param output The output file path that will be created.
 #'   If the file name does not include an extension, ".html" will be added automatically.
 #'   Any directories in the path will be created as needed.
+#' @param ... Additional arguments to pass to rmarkdown::render()
 #'
 #' @return The full path of the output file
 #' @export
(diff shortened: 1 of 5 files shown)


allyhawkins (approved):
This looks good to me, thanks for doing this!

jashapiro:
Tested local docker build, so we should be good to go.

PR #128, opened 2022-09-29
Add scran to scpcaTools docker image

allyhawkins:
In working on AlexsLemonade/scpca-nf#207 I was testing my additions to the workflow and getting an error that the `scran` library wasn't found. I realized we don't have `scran` in the docker image that we build for `scpcaTools` yet so here I'm adding it to the dependencies list and then adding scran and associated dependencies to the lock file.


jashapiro (approved):
LGTM! Unfortunately, we will need to make a new release now :sad:

allyhawkins:
> LGTM! Unfortunately, we will need to make a new release now :sad:

Can we just use `edge` for development after this is merged and then make a new release after we incorporate the QC report changes?

jashapiro:
> Can we just use `edge` for development after this is merged and then make a new release after we incorporate the QC report changes?

Yes, we can. That is probably the better plan!

PR #130, opened 2022-09-29
Add scater to Docker image

allyhawkins:
After fixing the missing `scran` library problem and re-testing https://github.com/AlexsLemonade/scpca-nf/pull/221, it looks like we are missing `scater` from the docker image too. I really thought it was in there but forgot we use `scuttle` now for all the QC steps... So this adds in `scater`. This should be all the libraries we need now.


jashapiro (approved):
🎢

PR #131, opened 2022-10-06
Update QC report to include post processing stats and UMAP

### LINKED ISSUE #121: Add UMAP visualization to QC report

People like to see UMAPs of the data, so we should provide them if we can!

If there is a UMAP matrix in the provided SCE object, we should show it in the report somewhere (probably near the top). If we wanted to, we could fall back to tSNE or PCA if those are present. 

We could calculate the UMAP within the notebook if it is missing, but I think it may make sense to make this more of an optional thing, and only show it if the UMAP was previously calculated and will be present in the download. 

**Additional context**

Some of the implementation may depend on the implementation of https://github.com/alexslemonade/scpca-nf/issues/203; if we can not include a soft filter such that only some cells are included in the UMAP visualization, we may need to pass in a 3rd object which is a further filtered version of the object (with damaged or low-information cells removed) from which to plot the UMAP.

We may also want to add some more summary lines to the report that describe the number of cells that were removed by miQC filtering and/or low information content (few genes).

allyhawkins:
Closes #121 

Here I started making changes to the QC report to incorporate the additional processed SCE object that we are now creating as output in `scpca-nf`. I added a param to the report for the `processed_sce` and set the default to `NULL`, similar to the setup for the `filtered_sce`. I also added to the `Cell Statistics` table additional rows to include the method used for filtering low quality cells, number of cells, and normalization method. I also thought we should include any parameters that might be altered such as the min genes cutoff and the probability of a cell being compromised cutoff used for miQC but we currently don't store those in the metadata of the SCE object and we probably should. I still added a line for them in the table here and if we like that change we can file an issue to add that change to the script in `scpca-nf`. 

The other thing I did was add a UMAP if the `processed_sce` param was not null. I wasn't quite sure what we wanted for color here so I tried a few things. I think if we aren't going to be coloring by clusters then we probably want to color by either total UMI count or by total genes detected which is what I did for now. I did also play around with plotting the gene expression of the most highly variable gene too and that might be promising. I included a screenshot here: 

image

I think if we went this route we would want to also include the gene symbol, but I didn't do much formatting since I wasn't sure if that was really a joke or not... We could also plot a few of the most HVGs and make a faceted UMAP... 

Another thought I had was that we might want to include a plot showing the filtering cutoffs of the cells that were kept vs. thrown out after using miQC + min gene threshold. I didn't add it in yet because I think that might be a separate issue, but just wanted to mention that here. 

I'm attaching a current copy of the QC report with the new additions and filing this as a draft PR just to get some initial feedback on the UMAP mostly. Once we decide the direction we want to go with the UMAP then I'll polish up the plot and report and add in a description around it like we have for the other plots. 
[test.html.zip](https://github.com/AlexsLemonade/scpcaTools/files/9726903/test.html.zip)

### DIFF
--- a/inst/rmd/qc_report.rmd
+++ b/inst/rmd/qc_report.rmd
@@ -384,12 +384,12 @@ In such situations, the calculated probability of compromise may not be valid (s
 ```{r}
 if(has_processed){
   
-  has_var_genes <- !is.null(processed_meta$highly_variable_genes)
-  
+  # check that processed SCE contains UMAP and PCA
   if(!"UMAP" %in% reducedDimNames(processed_sce)){
     if(!"PCA" %in% reducedDimNames(processed_sce)){
       
-      if(has_var_genes){
+      # if PCA is missing, check for HVG, if no HVG use all genes 
+      if(!is.null(processed_meta$highly_variable_genes)){
         subset_genes <- processed_meta$highly_variable_genes
       }else{
         subset_genes <- rownames(processed_sce)
@@ -402,26 +402,11 @@ if(has_processed){
                                      dimred = "PCA")
   }
   
-  if(has_var_genes){
-    max_gene <- processed_meta$highly_variable_genes[1]
-    var_gene_exp <- counts(processed_sce[max_gene,]) %>%
-      as.matrix() %>%
-      t() %>%
-      as.data.frame() %>%
-      tibble::rownames_to_column("barcode")
-    
-    coldata_df <- colData(processed_sce) %>%
-      as.data.frame() %>%
-      tibble::rownames_to_column("barcode") %>%
-      dplyr::left_join(var_gene_exp, by = "barcode")
-    
-    colData(processed_sce) <- DataFrame(coldata_df, row.names = coldata_df$barcode)
-  }
-  
+  # create UMAP colored by number of genes detected 
  scater::plotReducedDim(processed_sce, 
                                 dimred = "UMAP", 
                                 point_size = 0.1, 
-                                colour_by = max_gene) +
+                                colour_by = "detected") +
   theme_bw() 
 
 }
--- a/inst/rmd/qc_report.rmd
+++ b/inst/rmd/qc_report.rmd
@@ -407,7 +407,10 @@ if(has_processed){
                          dimred = "UMAP", 
                          point_size = 0.1, 
                          colour_by = "detected") +
-    theme_bw() 
+    theme_bw() + 
+    labs(
+      color = "Number of genes detected"
+    )
 
 }
 ```
--- a/inst/rmd/qc_report.rmd
+++ b/inst/rmd/qc_report.rmd
@@ -381,6 +381,53 @@ This can be the result of a low-quality library or may occur if there is no mito
 In such situations, the calculated probability of compromise may not be valid (see [miQC vignette](https://bioconductor.org/packages/3.13/bioc/vignettes/miQC/inst/doc/miQC.html#when-not-to-use-miqc) for more details).
 
 
+```{r}
+if(has_processed){
+  
+  has_var_genes <- !is.null(processed_meta$highly_variable_genes)
+  
+  if(!"UMAP" %in% reducedDimNames(processed_sce)){
+    if(!"PCA" %in% reducedDimNames(processed_sce)){
+      
+      if(has_var_genes){
+        subset_genes <- processed_meta$highly_variable_genes
+      }else{
+        subset_genes <- rownames(processed_sce)
+      }
+      processed_sce <- scater::runPCA(processed_sce,
+                                      subset_row = subset_genes)
+    }
+    
+    processed_sce <- scater::runUMAP(processed_sce, 
+                                     dimred = "PCA")
+  }
+  
+  if(has_var_genes){
+    max_gene <- processed_meta$highly_variable_genes[1]
+    var_gene_exp <- counts(processed_sce[max_gene,]) %>%
+      as.matrix() %>%
+      t() %>%
+      as.data.frame() %>%
+      tibble::rownames_to_column("barcode")
+    
+    coldata_df <- colData(processed_sce) %>%
+      as.data.frame() %>%
+      tibble::rownames_to_column("barcode") %>%
+      dplyr::left_join(var_gene_exp, by = "barcode")
+    
+    colData(processed_sce) <- DataFrame(coldata_df, row.names = coldata_df$barcode)
+  }
+  
+ scater::plotReducedDim(processed_sce, 
+                                dimred = "UMAP", 
+                                point_size = 0.1, 
+                                colour_by = max_gene) +
+  theme_bw() 
+
+}
+```
+
+
 <!-- Next section included only if CITE-seq data is present -->
 ```{r, child='cite_qc.rmd', eval = has_cite}
 ```
--- a/inst/rmd/qc_report.rmd
+++ b/inst/rmd/qc_report.rmd
@@ -3,6 +3,7 @@ params:
   library: Example
   unfiltered_sce: !r scpcaTools:::sim_sce()
   filtered_sce: NULL
+  processed_sce: NULL
   date: !r Sys.Date()
 
 title: "`r glue::glue('ScPCA QC report for {params$library}')`"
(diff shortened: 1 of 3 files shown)


inst/rmd/qc_report.rmd
```diff
@@ -204,17 +207,39 @@ knitr::kable(processing_info, align = 'r') |>
...
+# if processed sce exists add filtering and normalization table
+if(has_processed){
+  processed_meta <- metadata(processed_sce)
+  
+  basic_statistics <- basic_statistics |>
+    mutate(
+      "Method used to filter low quality cells" = format(processed_meta$ccdl_filter_method),
+      "Cells after filtering low quality cells" = format(dim(processed_sce)[2], big.mark = ',', scientific = FALSE),
+      "Normalization method"                    = format(processed_meta$normalization),
+      "Minimum genes per cell cutoff"           = format(processed_meta$min_gene_cutoff)
```
jashapiro, line 227: I noticed in the report that this was NA, but this should usually be set, correct? Or are we not using a minimum even with miQC?

inst/rmd/qc_report.rmd
```diff
@@ -286,7 +311,7 @@ ggplot(grouped_celldata, aes(x = med_rank, y = med_sum, color = pct_passed)) +
 ```
 
 The total UMI count of each droplet (barcode) plotted against the rank of that droplet allows visualization of the distribution of sequencing depth across droplets. 
-The droplets that are expected to contain cells were identified with [`DropletUtils::emptyDrops()`](https://bioconductor.org/packages/release/bioc/html/DropletUtils.html), which uses both the total UMI counts and expressed gene content ([Lun  _et al._ 2019](https://doi.org/10.1186/s13059-019-1662-y)).
+The droplets that are expected to contain cells were identified with [`DropletUtils::emptyDropsCellRanger()`](https://bioconductor.org/packages/release/bioc/html/DropletUtils.html), which uses both the total UMI counts and expressed gene content ([Lun  _et al._ 2019](https://doi.org/10.1186/s13059-019-1662-y)).
```
jashapiro, line 314: Do we want to say something like "unless otherwise specified"?

I think we may also have to update/add the citation to the alevin-fry paper for the details of that method? Or say something about how it derived from the Lun et al method.

inst/rmd/qc_report.rmd
```diff
@@ -356,6 +381,41 @@ This can be the result of a low-quality library or may occur if there is no mito
...
+  
+  # create UMAP colored by number of genes detected 
+  scater::plotReducedDim(processed_sce, 
+                         dimred = "UMAP", 
+                         point_size = 0.1, 
+                         colour_by = "detected") +
+    theme_bw() + 
+    labs(
+      color = "Number of genes detected"
+    )
```
jashapiro, line 413: With `plotReducedDim` this doesn't work, for some reason. We have to go back to

```suggestion
    guides(
      color = guide_colorbar(title="Number of genes detected"))
    )
```

unless you can find some way around it!

We probably want to break that legend into two lines with a careful `\n`

inst/rmd/qc_report.rmd
```diff
@@ -356,6 +381,41 @@ This can be the result of a low-quality library or may occur if there is no mito
...
+                                      subset_row = subset_genes)
+    }
+    
+    processed_sce <- scater::runUMAP(processed_sce, 
+                                     dimred = "PCA")
+  }
+  
+  # create UMAP colored by number of genes detected 
+  scater::plotReducedDim(processed_sce, 
+                         dimred = "UMAP",
```
jashapiro, line 407: Just simplifying a bit since this is a default `UMAP`
```suggestion
  scater::plotUMAP(processed_sce, 
```

inst/rmd/qc_report.rmd
```diff
@@ -356,6 +381,41 @@ This can be the result of a low-quality library or may occur if there is no mito
...
+    }
+    
+    processed_sce <- scater::runUMAP(processed_sce, 
+                                     dimred = "PCA")
+  }
+  
+  # create UMAP colored by number of genes detected 
+  scater::plotReducedDim(processed_sce, 
+                         dimred = "UMAP", 
+                         point_size = 0.1,
```
jashapiro, line 408: Maybe a bit bigger?
```suggestion
                         point_size = 0.3, 
```

jashapiro (commented):
This looks like a good start to me! I think the plot looks pretty good as it is, but of course there are places for improvement!

> I think if we went this route we would want to also include the gene symbol, but I didn't do much formatting since I wasn't sure if that was really a joke or not... We could also plot a few of the most HVGs and make a faceted UMAP...

I like this initial plot of the gene counts; it adds a nice bit of color that does have some functional information in it. I think it is a good choice until we add clustering, or in the absence of clustering.

I do also like the idea of faceting a few different genes as a secondary view, but I would probably end up wanting to do that more manually than with the `plotUMAP` function, so I might save it for another day/PR. I would implement this as our own function, which is easier given that we are already working within a package! 

(Briefly, I would join the UMAP matrix to the expression table with the genes of interest, `pivot_longer` to put all the expression values into a single column with a separate column of gene names, then plot with `facet_wrap`)

> Another thought I had was that we might want to include a plot showing the filtering cutoffs of the cells that were kept vs. thrown out after using miQC + min gene threshold. I didn't add it in yet because I think that might be a separate issue, but just wanted to mention that here.

I kind of like this idea, but what kind of plot were you thinking of? Was this going to be some kind of distribution plot, or would we overlay it one of the other plots we already have somehow?

inst/rmd/qc_report.rmd
allyhawkins: this is usually set but we haven't added it to the metadata yet, so we need to add it to the metadata of the SCE object for it to print the value here. That needs to happen in `scpca-nf` though and not here.

allyhawkins:
> I do also like the idea of faceting a few different genes as a secondary view, but I would probably end up wanting to do that more manually than with the plotUMAP function, so I might save it for another day/PR. I would implement this as our own function, which is easier given that we are already working within a package!

This sounds good to me! I'll leave the faceted plots showing HVGs for another PR and file an issue about adding that. 

> I kind of like this idea, but what kind of plot were you thinking of? Was this going to be some kind of distribution plot, or would we overlay it one of the other plots we already have somehow?

I was thinking something similar to this plot where we show both distributions and then we just label the `keep` cells. But here we would only have one dashed line indicating the num genes detected cutoff. Would also want to file an issue about this to do in a separate PR.

image

allyhawkins:
Thanks for taking a look at this! I went ahead and added a section for the UMAP and a brief description. I was struggling with a heading title and the detail in the description, so any feedback on that would be great! 

I also added a "unless otherwise specified" to the filtering statement and a "adapted from" in the reference. These were the lines from the alevin-fry paper, about where they got `emptyDropsCellRanger` from: 

> To compare the results from different quantification tools in the snRNA-seq setting under a consistent and robust barcode filtering approach, we implemented the emptyDrops_CR functionality of STARsolo in R. The emptyDrops_CR filtering method is, itself, reverse engineered from the hybrid filtering strategy of Cell Ranger, which combines filtering based on various thresholds with the statistical testing method introduced by Lun et al. ([13](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8933848/#R40)). This functionality is now included in the DropletUtils R package as the emptyDropsCellRanger function.

I also made the small adjustments to the UMAP, but I couldn't remove the `dimred` argument because it throws an error that it's required. I also thought about using another color scale but we would ned a colorblind friendly continuous scale? Any suggestions on other ones besides viridis that you like? 

I filed the other issues that will need to be addressed too. 

Attached is the most updated copy of the report. 
[SCPCL000001_test.html.zip](https://github.com/AlexsLemonade/scpcaTools/files/9737020/SCPCL000001_test.html.zip)

inst/rmd/qc_report.rmd
jashapiro: Note that this change wasn't just removing `dimred`, but also changing to `plotUMAP()`, which removes the need for that argument (as long as there is a reduced dimension table named UMAP).

jashapiro (approved):
LGTM! My only comment below is about the name of the section: For some reason "Dimensionality Reduction" seems a bit opaque to me. I don't feel very strongly about this at all.

allyhawkins:
> LGTM! My only comment below is about the name of the section: For some reason "Dimensionality Reduction" seems a bit opaque to me. I don't feel very strongly about this at all.

I would agree that it's vague. I wasn't really sure what to name it... But "UMAP" also seemed opaque. I also think we might update this as we add in additional UMAPs showing HVGs. If you have other ideas I am definitely open to them.

PR #135, opened 2022-10-13
replace Matrix.utils with DelayedArray

### LINKED ISSUE #134: [BUG] Matrix.utils has been removed from CRAN
**Describe the bug**
The removal of `Matrix.utils` from CRAN means we need to remove it as a dependency from our code, as the package will no longer build

The only place that we use the function directly is in `collapse_intron_counts()`, which is in turn called by `read_alevin()` and `read_kallisto()`

For `read_alevin()`, we should be able to use on `fishpond` to perform the same functions for reading and collapsing directly. 

I don't think kallisto stores ambiguous counts separately, so we should be able to simply drop the intron reads and not use the `Matrix.utils::aggregate.Matrix()` function at all.

We can probably perform this step within the `read_kallisto()` function itself and remove the `collapse_intron_counts()` function completely.

jashapiro:
Closes #134

Here I am replacing `Matrix.utils::aggregate()` with the more maintained (and in Bioconductor) `DelayedArray::rowsum()`. There were a couple of changes to tests required and to the `read_tximport` function, because  `DelayedArray::rowsum()` _only_ works on sparse Matrices, and otherwise falls back to base `rowsum`, which fails on non-sparse `Matrix` types.

While this looks like it adds a dependency, it does not really, as `DelayedArray` is already required.

Speaking of dependencies, I moved `scran` and `scater` to the `DESCRIPTION` file.

I also fixed a warning I saw in the use of `zellkonverter` while testing


jashapiro:
I made one more change to quiet a warning: I moved the requirements for generating the QC report to "Suggests", and added checks for those packages at the start of `generate_qc_report()` function.

allyhawkins (approved):
Thank you for doing this! LGTM

PR #137, opened 2022-10-27
UMAPs faceted by highly variable genes

### LINKED ISSUE #132: Add faceted UMAPs showing expression of top HVG to QC report
As mentioned in https://github.com/AlexsLemonade/scpcaTools/pull/131, we may want to consider adding in additional UMAPs to the QC report showing cells colored by the top genes identified as highly variable genes. Here we would want to grab the top HVGs and have a UMAP for each HVG, faceting the UMAPs together so they are one panel. 

I'm including some thoughts that @jashapiro mentioned in https://github.com/AlexsLemonade/scpcaTools/pull/131#pullrequestreview-1133359583 for reference: 

> I do also like the idea of faceting a few different genes as a secondary view, but I would probably end up wanting to do that more manually than with the `plotUMAP` function, so I might save it for another day/PR. I would implement this as our own function, which is easier given that we are already working within a package!

> (Briefly, I would join the UMAP matrix to the expression table with the genes of interest, `pivot_longer` to put all the expression values into a single column with a separate column of gene names, then plot with `facet_wrap`)

allyhawkins:
Closes #132 

Here I added a section at the end of the report showing a grid of UMAPs with the most highly expressed genes. I started with 12 genes, but we can change that based on what others think of the plots. I did some adjusting to how the plot looks, but I'm not sold on any of it so if others have suggestions on how to make it pretty, I'm open to those! 

I also was wondering if we should make the number of genes we select be a parameter, or do we want to just stick with a number? 

The other thing I did here was utilize the gene symbols that we store in the `rowData` of our objects. I could have plotted ensembl gene ID's as the names, but I thought only Josh would understand those so our users might appreciate actually knowing what the genes were. 

The only other thing I didn't do was make sure that the plots are showing up in order of most variation to least variation among the top 12 genes that I grabbed. If we do want to ensure that happens there's a bit more work that we have to do in creating the plot. 

Here's an example of what the plot looks like right now after rendering in the report: 
image

### DIFF
--- a/inst/rmd/qc_report.rmd
+++ b/inst/rmd/qc_report.rmd
@@ -384,10 +384,10 @@ In such situations, the calculated probability of compromise may not be valid (s
 ## Dimensionality Reduction 
 
 ```{r message=FALSE}
+# use to check if variable genes are present, need for both UMAPs
+has_var_genes <- !is.null(processed_meta$highly_variable_genes)
+
 if(has_processed){
-  
-  # use to check if variable genes are present, need for plotting later
-  has_var_genes <- !is.null(processed_meta$highly_variable_genes)
 
   # check that processed SCE contains UMAP and PCA
   if(!"UMAP" %in% reducedDimNames(processed_sce)){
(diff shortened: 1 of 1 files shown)


inst/rmd/qc_report.rmd
```diff
@@ -420,7 +423,83 @@ if(has_processed){
...
+The above plot shows the UMAP (Uniform Manifold Approximation and Projection) embeddings for each cell, coloring each cell by the total number of genes detected per cell.
+The below plots show the same UMAP embeddings, coloring each cell by the gene expression of the gene indicated in the label for each panel. The genes chosen for plotting are the top 12 highly variable genes identified in the library.
+
+```{r message=FALSE}
+if(has_processed){
+  
+  if(has_var_genes){
+    
+    # select top genes to plot 
+    top_genes <- processed_meta$highly_variable_genes %>%
```
jashapiro, line 435: A question that may inform later questions... are the hvgs in the metadata in order from most variable to least? If so, we might want to make this an ordered factor to preserve that. If not, we should change so it is, or this might not be the "top" genes.

jashapiro (commented):
Just the one question to start... on my phone, but I thought I'd ask.

inst/rmd/qc_report.rmd
allyhawkins: Yes, when I was playing around with this before I checked that they are in order of variance and the first gene that's returned is the one with the highest variation.

inst/rmd/qc_report.rmd
```diff
@@ -420,7 +423,83 @@ if(has_processed){
...
+      as.data.frame() %>%
+      tibble::rownames_to_column("barcode")
+    
+    # grab rowdata as data frame to later combine with gene expression data
+    # rowdata contains the mapped gene symbol so can use that to label plots instead of ensembl gene id from rownames 
+    rowdata_df <- rowData(processed_sce) %>%
+      as.data.frame() %>%
+      tibble::rownames_to_column("ensembl_id") %>%
+      dplyr::select(ensembl_id, gene_symbol) %>%
+      dplyr::filter(ensembl_id %in% top_genes)
```
jashapiro, line 451: I _think_ this will put everything in the correct order.

Note I also changed the pipe operator to match the earlier part of the notebook. This notebook also doesn't use dplyr prefixes in general. We don't load many packages, but I tend to feel that `dplyr` is okay in certain contexts... and I guess I had decided this was one of them.

```suggestion
    rowdata_df <- rowData(processed_sce) |>
      as.data.frame() |>
      tibble::rownames_to_column("ensembl_id") |>
      select(ensembl_id, gene_symbol) |>
      filter(ensembl_id %in% top_genes) |>
      mutate(ensembl_id = factor(ensembl_id, levels = top_genes)) |>
      arrange(ensembl_id) |>
      mutate(gene_symbol = factor(gene_symbol, levels = gene_symbol))
      
```

allyhawkins:
Thanks for the suggestion! I went ahead and implemented the change to factors and reordered and that switched the order of the plots. I also removed the magrittr pipe and put the base R pipe. I had forgotten that we had been using that here. Should be ready for another look!

inst/rmd/qc_report.rmd
```diff
@@ -420,7 +423,87 @@ if(has_processed){
 
 Compromised cells, calculated using `miQC` described above, are filtered and removed from the library prior to downstream analyses. 
 The raw counts from all remaining cells are then normalized prior to selection of highly variable genes and dimensionality reduction. 
-Here we show the UMAP (Uniform Manifold Approximation and Projection) embeddings for each cell, coloring each cell by the total number of genes detected per cell.
+The above plot shows the UMAP (Uniform Manifold Approximation and Projection) embeddings for each cell, coloring each cell by the total number of genes detected per cell.
+The below plots show the same UMAP embeddings, coloring each cell by the gene expression of the gene indicated in the label for each panel. The genes chosen for plotting are the top 12 highly variable genes identified in the library.
```
jashapiro, line 427: Simplifying the wording here a bit.
```suggestion
The plots below show the same UMAP embeddings, coloring each cell by the expression level of the labeled gene.
The genes chosen for plotting are the 12 most variable genes identified in the library.
```

inst/rmd/qc_report.rmd
```diff
@@ -420,7 +423,87 @@ if(has_processed){
...
+
+```{r message=FALSE}
+if(has_processed){
+  
+  if(has_var_genes){
```
jashapiro, line 432: I think we only need `if(has_var_genes){` here? 

In the current setup, if `has_var_genes` is False, the `ggplot` will fail as `coldata_df` wasn't yet set up, correct?

For this to work, we will want to initialize `has_var_genes <- FALSE` before line 387, or make the line something like: 
``` 
if(has_processed & has_var_genes){
```

inst/rmd/qc_report.rmd
```diff
@@ -420,7 +423,87 @@ if(has_processed){
...
+    
+    
+  }
+  
+  ggplot(coldata_df, aes(x = UMAP1, y = UMAP2, color = gene_expression)) + 
+    geom_point(alpha = 0.1, size = 0.001) + 
+    facet_wrap(~ gene_symbol) +
+    scale_color_viridis_c() +
+    labs(
+      color = "Log-normalized \ngene expression"
```
jashapiro, line 487: I don't think we need to split this onto two lines, given that we are putting the gradient horizontal at the top (or bottom)
```suggestion
      color = "Log-normalized gene expression"
```

inst/rmd/qc_report.rmd
```diff
@@ -420,7 +423,87 @@ if(has_processed){
...
+    ) +
+    theme(legend.position = "top",
```
jashapiro, line 489: I might like the bottom better?
```suggestion
    theme(legend.position = "bottom", 
```

inst/rmd/qc_report.rmd
```diff
@@ -420,7 +423,87 @@ if(has_processed){
...
+    scale_color_viridis_c() +
+    labs(
+      color = "Log-normalized \ngene expression"
+    ) +
+    theme(legend.position = "top", 
+          axis.text = element_text(size = 8, color = "black"),
+          axis.title = element_text(size = 8, color = "black"),
+          legend.title = element_text(size = 8),
+          legend.text = element_text(size = 6)) + 
+    guides(colour = guide_colorbar(title.position = "top", title.hjust = 0.5))
```
jashapiro, line 494: If putting the legend on the bottom, probably the bottom for the title too...
```suggestion
    guides(colour = guide_colorbar(title.position = "bottom", title.hjust = 0.5))
```

inst/rmd/qc_report.rmd
```diff
@@ -420,7 +423,87 @@ if(has_processed){
...
+    facet_wrap(~ gene_symbol) +
+    scale_color_viridis_c() +
+    labs(
+      color = "Log-normalized \ngene expression"
+    ) +
+    theme(legend.position = "top", 
+          axis.text = element_text(size = 8, color = "black"),
+          axis.title = element_text(size = 8, color = "black"),
+          legend.title = element_text(size = 8),
+          legend.text = element_text(size = 6)) +
```
jashapiro, line 493: I probably wouldn't make these quite so small... 8 pt is already pretty small!

I wonder if we want to play with `strip.text` and `strip.background`. I think the grey background is okay, but we might go a bit smaller, or match the axis text, or no border on the background?

I might try:
```
strip.background = element_rect(fill = 'transparent'),
strip.text = element_text(size = 8) 
```

But not something to spend much time on.

jashapiro (approved):
This looks good to me, with one logic question and a couple of aesthetic suggestions, but I'm really not too concerned about the latter, so I am approving!

PR #138, opened 2022-11-01
Plot removal of low quality cells in QC report

### LINKED ISSUE #133: Add plot showing filtered cells to QC report
Currently in the QC report we show the miQC model and the cells are colored by probability compromised, but we don't actually show a plot with the cells that are filtered prior to downstream processing to create the UMAP. 

We should consider adding a plot that highlights cells that were kept vs. discarded. One thought I had was to have a similar plot to the miQC plot with genes detected on the x-axis and mito content on the y-axis but have the color correspond to if a cell was filtered or not. We would also include a line to indicate the cutoff used for min number of genes detected. 

See https://github.com/AlexsLemonade/scpcaTools/pull/131#issuecomment-1271918909 for an example.

allyhawkins:
Closes #133 

Here I added a plot below the miQC diagnostics plot highlighting which cells were removed and which cells were kept for further downstream analyses. For this plot I have a check that the library has both the `filtered_sce` and `processed_sce` and if that is the case, I pull the cutoffs and filtering method used from the `processed_sce`. I then used the `colData` from the filtered SCE object since that contains all the cells as well as the individual prob compromised values for each cell prior to removing cells in the processed SCE object. Depending on the filter method I labeled the cells as either keep or remove and then plotted them in a scatter plot, indicating the minimum gene cutoff as a vertical line. 

The only other change I had to make outside of the chunk that makes this plot is actually grabbing the `colData` prior to creating the miQC model plot, otherwise we lose the `prob_compromised` column and can't use that for labeling cells. 

I've attached a copy of the current plot after rendering in the report, but definitely open to suggestions on how to improve the aesthetics or if there's something we can change to make things more informative. 

image

### DIFF
--- a/inst/rmd/qc_report.rmd
+++ b/inst/rmd/qc_report.rmd
@@ -397,10 +397,12 @@ The dotted vertical line indicates the minimum gene cutoff used for filtering.
 ```{r}
 if(has_filtered & has_processed){
   
+  # grab cutoffs and filtering method from processed sce
   min_gene_cutoff <- processed_meta$min_gene_cutoff
   prob_compromised_cutoff <- processed_meta$prob_compromised_cutoff
   filter_method <- processed_meta$ccdl_filter_method
   
+  # add column to coldata labeling cells to keep/remove based on filtering method
   if(filter_method == "miQC"){
     
     filtered_coldata_df <- filtered_coldata_df %>%
--- a/inst/rmd/qc_report.rmd
+++ b/inst/rmd/qc_report.rmd
@@ -403,30 +403,21 @@ if(has_filtered & has_processed){
   filter_method <- processed_meta$ccdl_filter_method
   
   # add column to coldata labeling cells to keep/remove based on filtering method
-  if(filter_method == "miQC"){
+  filtered_coldata_df <- colData(filtered_sce) |>
+    as.data.frame()
     
-    filtered_coldata_df <- filtered_coldata_df %>%
-      dplyr::mutate(ccdl_filter = ifelse(prob_compromised <= prob_compromised_cutoff & 
-                                           detected >= min_gene_cutoff, 
-                                         "Keep", 
-                                         "Remove"))
-    
-  } else if(filter_method == "Minimum_gene_cutoff") {
-    
-    filtered_coldata_df <- filtered_coldata_df %>%
-      dplyr::mutate(ccdl_filter = ifelse(detected >= min_gene_cutoff, 
-                                         "Keep", 
-                                         "Remove"))
-    
-  }
+  filtered_coldata_df <- filtered_coldata_df |>
+    dplyr::mutate(ccdl_filter = ifelse(rownames(filtered_coldata_df) %in%  colnames(processed_sce), 
+                                       "Keep", 
+                                       "Remove")) 
   
   ggplot(filtered_coldata_df, aes(x = detected, y = subsets_mito_percent, color = ccdl_filter)) + 
       geom_point(alpha = 0.5, size = 1) +
       geom_vline(xintercept = min_gene_cutoff, linetype = "dashed") +
       labs(x = "Number of genes detected",
            y = "Mitochondrial percentage",
-           color = "Filter") +
-    ggtitle(stringr::str_replace(filter_method, "_", " ")) +
+           color = "Filter"
+           title = stringr::str_replace(filter_method, "_", " ")) +
     theme(plot.title = element_text(hjust = 0.5),
           legend.position = c(1, 1),
           legend.justification = c(1,1),
--- a/inst/rmd/qc_report.rmd
+++ b/inst/rmd/qc_report.rmd
@@ -422,7 +422,7 @@ if(has_filtered & has_processed){
   glue::glue("
     <div class=\"alert alert-warning\">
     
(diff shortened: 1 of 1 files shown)


inst/rmd/qc_report.rmd
```diff
@@ -381,6 +385,65 @@ If the model has failed to fit properly, the pattern of cells may differ, and th
...
+    
+  }
+  
+  ggplot(filtered_coldata_df, aes(x = detected, y = subsets_mito_percent, color = ccdl_filter)) + 
+      geom_point(alpha = 0.5, size = 1) +
+      geom_vline(xintercept = min_gene_cutoff, linetype = "dashed") +
+      labs(x = "Number of genes detected",
+           y = "Mitochondrial percentage",
+           color = "Filter") +
+    ggtitle(stringr::str_replace(filter_method, "_", " ")) +
```
jashapiro, line 429: Making the title part of labs?
```suggestion
           color = "Filter"
           title = stringr::str_replace(filter_method, "_", " ")) +
```

inst/rmd/qc_report.rmd
```diff
@@ -381,6 +385,65 @@ If the model has failed to fit properly, the pattern of cells may differ, and th
...
+                                         "Remove"))
+    
+  } else if(filter_method == "Minimum_gene_cutoff") {
+    
+    filtered_coldata_df <- filtered_coldata_df %>%
+      dplyr::mutate(ccdl_filter = ifelse(detected >= min_gene_cutoff, 
+                                         "Keep", 
+                                         "Remove"))
+    
+  }
```
jashapiro, line 421: I think we want to simplify this to use the processed data. This also means we don't need different code for miQC & the simpler cutoff. 

We can probably move making `filtered_coldata_df` to here as well, since we won't need the `prob_compromised` at all
```suggestion
  filtered_coldata_df <- colData(filtered_sce) |>
    as.data.frame()
    
  filtered_coldata_df <- filtered_coldata_df |>
    dplyr::mutate(ccdl_filter = ifelse(rownames(filtered_coldata_df) %in%  colnames(processed_sce), 
                                       "Keep", 
                                       "Remove")) 
```

inst/rmd/qc_report.rmd
```diff
@@ -381,6 +385,65 @@ If the model has failed to fit properly, the pattern of cells may differ, and th
...
+    theme(plot.title = element_text(hjust = 0.5),
+          legend.position = c(1, 1),
+          legend.justification = c(1,1),
+          legend.background = element_rect(color = "grey20", size= 0.25),
+          legend.title = element_text(hjust = 0.5))
+} else {
+  glue::glue("
+    <div class=\"alert alert-warning\">
+    
+    No filtering was performed on this library.
```
jashapiro, line 439: Is this accurate? Do we want to distinguish filtering from processing?

jashapiro (commented):
I think we actually want to take a very different approach here, which should actually simplify the code quite a bit. What you are doing here is recalculating the filtering we did to make the `processed_sce` object, but since we have that object, I think we should be using it directly. 

I am not sure the code suggestion I made is 100% correct, but I hope it makes sense as the idea!

inst/rmd/qc_report.rmd
```diff
@@ -381,6 +381,55 @@ If the model has failed to fit properly, the pattern of cells may differ, and th
...
+  filter_method <- processed_meta$ccdl_filter_method
+  
+  # add column to coldata labeling cells to keep/remove based on filtering method
+  filtered_coldata_df <- colData(filtered_sce) |>
+    as.data.frame()
+    
+  filtered_coldata_df <- filtered_coldata_df |>
+    dplyr::mutate(ccdl_filter = ifelse(rownames(filtered_coldata_df) %in%  colnames(processed_sce), 
+                                       "Keep", 
+                                       "Remove"))
```
jashapiro, line 407: `tidy`ing up a bit.

```suggestion
  filtered_coldata_df <- colData(filtered_sce) |>
    as.data.frame() |>
    tibble::rownames_to_column("barcode") |>
    dplyr::mutate(ccdl_filter = ifelse(barcode %in%  colnames(processed_sce), 
                                       "Keep", 
                                       "Remove")) 
```

inst/rmd/qc_report.rmd
```diff
@@ -381,6 +381,55 @@ If the model has failed to fit properly, the pattern of cells may differ, and th
...
+    as.data.frame()
+    
+  filtered_coldata_df <- filtered_coldata_df |>
+    dplyr::mutate(ccdl_filter = ifelse(rownames(filtered_coldata_df) %in%  colnames(processed_sce), 
+                                       "Keep", 
+                                       "Remove")) 
+  
+  ggplot(filtered_coldata_df, aes(x = detected, y = subsets_mito_percent, color = ccdl_filter)) + 
+    geom_point(alpha = 0.5, size = 1) +
+    geom_vline(xintercept = dmin_gene_cutoff, linetype = "dashed") +
```
jashapiro, line 411: I think a typo?
```suggestion
    geom_vline(xintercept = min_gene_cutoff, linetype = "dashed") +
```

jashapiro (approved):
This looks much simpler! Thanks for making these changes.

I think I found a typo, so probably worth running one more check before merging.

Issue #136, opened 2022-10-14
Use fishpond for alevin-fry reading


jashapiro:
In #134, I suggested that a solution to the removal of `Matrix.utils` might be to replace our matrix reading and collapse code with `fishpond`, but I did not implement that in #135 in order to minimize the time to restoring a working version of the package. Nonetheless, this seems like it might be a worthwhile addition, particularly if we want to expand to retaining both spliced and unspliced data (possibly in conjunction with https://github.com/AlexsLemonade/scpca-nf/issues/214).

Using `fishpond` to read and process those data allows us to build on code that already exists to simplify or replace the `read_alevin_mtx` function.

PR #140, opened 2022-11-02
Use matrix math to do collapse introns

### LINKED ISSUE #139: [BUG] DelayedArray::rowsum fails with too many groups
**Describe the bug**
It turns out that my simple solution for replacing `Matrix.utils::aggregate` does not really work with full data sets

**To Reproduce**
Load in a full alevin-fry data set and try to collapse it with `scpcaTools::collapse_intron_counts`; we get an error that there are two many groups

**Additional context**
I think the solution is to implement #136, but we still want a version of this function that works if we are going to continue to support base alevin with an intron mode. Otherwise, we should probably remove the support for data types we are not using.

jashapiro:
Closes #139

While doing #136 is probablye *still* the right way to do this, I decided to get back to basics and use matrix multiplication to do the collapsing of intron counts.

So here I am taking the vector of genes that we were using as a grouping variable, finding the unique gene names, and then making a sparse matrix with the original count rows as columns and output genes as rows with 1s to indicate the columns that should be added for each row. Then we do a simple matrix multiplication (taking advantage of whatever sparse matrix optimizations are already present) to get the new matrix of collapsed/aggregated/summed counts.

In testing this, I am a bit confused why the original failed in the case I ran into; I would have expected the tests we already had to catch it, as they do load a real alevin-fry dataset. But here we are.

### DIFF
--- a/R/collapse_intron_counts.R
+++ b/R/collapse_intron_counts.R
@@ -53,8 +53,16 @@ collapse_intron_counts <- function(counts,
   }
   # remove -I -U -A at the end of the gene names
   genes <- str_remove(rownames(counts), "-[IUA]$")
-  # aggregate Matrix counts by gene name
-  counts <- DelayedArray::rowsum(counts, genes) |> as("dgCMatrix")
+
+  # aggregate Matrix counts by gene name using matrix math
+  # get unique gene names, encode genes as factors
+  ugenes <- unique(genes)
+  fgenes <- factor(genes, levels = ugenes)
+  # create a transformer matrix, with 1s at positions that correspond to old/new names, zeros elsewhere
+  transformer <- Matrix::sparseMatrix(as.integer(fgenes), 1:length(fgenes), x =1)
+  rownames(transformer) <- ugenes
+  # matrix multiplication to collapse
+  counts <- transformer %*% counts
   # drop extra slots silently
   counts <- suppressWarnings(BiocGenerics::updateObject(counts))
   return(counts)
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -28,7 +28,6 @@ Depends:
 Imports: 
     BiocGenerics,
     dplyr,
-    DelayedArray,
     DropletUtils,
     glue,
     jsonlite,


R/collapse_intron_counts.R
```diff
@@ -53,8 +53,16 @@ collapse_intron_counts <- function(counts,
...
-  # aggregate Matrix counts by gene name
-  counts <- DelayedArray::rowsum(counts, genes) |> as("dgCMatrix")
+
+  # aggregate Matrix counts by gene name using matrix math
+  # get unique gene names, encode genes as factors
+  ugenes <- unique(genes)
+  fgenes <- factor(genes, levels = ugenes)
+  # create a transformer matrix, with 1s at positions that correspond to old/new names, zeros elsewhere
+  transformer <- Matrix::sparseMatrix(as.integer(fgenes), 1:length(fgenes), x =1)
+  rownames(transformer) <- ugenes
```
allyhawkins, line 63: I'm a little confused on how these dimensions match up? The number of rows is going to be all the genes, while the rownames will be only the unique genes?

jashapiro: The number of rows is the number of genes, the number of columns is gene parts (gene-U,-A, etc.). 

What I am doing here is making a sparse matrix with values at specific locations, corresponding to a factor level that is the unique gene (from `as.integer(fgenes)` and position in the gene part list (from `1:length(fgenes)`), which corresponds to the rows of the original counts matrix. 

So if there are 600 gene parts from 200 genes, this will be a 200x600 matrix where each column has a single value with a value of 1, and each row will have 3, corresponding to the three parts of that gene. Then we multiply that transformer by our 600xN_cells matrix to get a 200xN_cells matrix at the end.

jashapiro: I only fill in the row names, as those will be preserved in the output. The gene part names (columns) will go away, so I didn't bother adding those in.

jashapiro: To be a bit more explicit... `sparseMatrix(c(1,1,2,2), c(1,2,3,4), x=1)` will give something like:
```
     [,1] [,2] [,3] [,4]
[1,]    1    1    0    0
[2,]    0    0    1    1
```
(but sparse) 

So I am only indirectly specifying the dimensions by the max value of the levels of the `fgenes` object for the number of rows, and its length for the number of columns.

allyhawkins: Ahh this makes so much more sense now. Thanks for the example! I was thinking that it was making a matrix with the dimensions equivalent to the number of gene parts x gene parts, which is why I was getting tripped up.

PR #141, opened 2022-11-03
Update to Bioconductor 3.16

jashapiro:
The main goal of this PR is to update the dependencies of scpcaTools to use Bioconductor 3.16 (and hence R 4.2), which adds support for Arm processors, and therefore should allow the package to run with native R on newer Apple laptops. 

This will allow for a number of nice downstream changes, assuming all goes to plan.

There were a few deprecation warnings that I encountered while making this switch, so I fixed those along the way, along with making sure that the conda environment for `zellkonverter` is built during the creation of the docker image. (I am not sure whether that step requires internet access, so it seemed worth  being safe here).

One error that I encountered is that miQC now throws an error if no cells are removed, which isn't great for the case where we were setting the cutoff to be 1. So I pulled the relevant code into a new function that is used where we just wanted to add the results of `prob_compromised` to the unfiltered table. 

I am currently testing building the docker image from this PR (hence draft status), but wanted to get the core filed so others could test using the package directly with R 4.2. It may be worth noting that R 4.2.2 was just released, but I am using 4.2.1 because that is what I had installed. If we want to go up another point, I wouldn't expect any major issues (and the automated testing should actually test in that context).


allyhawkins (approved):
Thanks for getting this started! Everything worked as expected on my end when I tested things.

jashapiro:
Okay, lets see what happens when the edge docker image goes live!

PR #144, opened 2022-11-16
Use fishpond for reading in Alevin-fry USA mode data

### LINKED ISSUE #136: Use fishpond for alevin-fry reading

allyhawkins:
Closes #136 

This PR does some reorganization in the functions that read in counts data for Alevin, Alevin-fry, and Kallisto. In particular, I made some modifications to use [`fishpond::loadFry()`](https://mikelove.github.io/fishpond/reference/loadFry.html) to read in any alevin-fry data that has been quantified using the USA mode. If it doesn't use USA mode, we can't load it in with that function and can still use our `collapse_intron_counts` function. 

I made two major changes in this PR and the rest of the changes are propagating those changes throughout to make sure that things are able to pass and that tests were up to date. 

1. I altered the `read_alevin` function to now do a check for `if(usa_mode)` and if that's true, the data is directly read in using `loadFry()`. This creates a SCE object directly, so within this `if` block I also used a check for `intron_mode` indicating if the data had been aligned to a reference containing intron sequences. If that is the case then there are two assays in the object, a `counts` assay that contains the unspliced and spliced counts totaled (what we were previously using for nuclei sequencing) and then a `spliced` assay that contains only the spliced cDNA counts. 
2. For non-USA mode data, which encompasses Alevin, non-USA alevin-fry, and Kallisto, I decided to create a separate `build_sce` function (I was iffy on the name so open to other ideas). This function takes in the original counts matrix and a flag indicating if the data is in `intron_mode` or not. For data in `intron_mode`, I called `collapse_intron_counts` using both `unspliced` and `spliced` to create two separate matrices. Those are then combined into a SCE object so that the `counts` assay contains the unspliced data and the `spliced` assay contains the spliced counts. For non-intron mode, the counts assay is directly added to the SCE object. 

One thing that I came across while testing the `build_sce` function was that the number of genes for unspliced and spliced didn't always match up and the dimensions are required to match for them both to be assays in an SCE object. So I chose to identify genes that are either only unspliced or only spliced and then added those as new rows to the corresponding matrix with 0s so that the dimensions match up. I think this is probably our best bet, but I'm open to other thoughts on this. 

In making these changes I made a few minor changes to arguments throughout: 

- Removed the `which_counts` argument everywhere except for in the `collapse_intron_counts` function. The reasoning here is that I am now going off the assumption that if `intron_mode` is on we are always creating objects with the unspliced counts. However, we could add an option to turn that off and just return the spliced cDNA even if aligning to intronic regions if we think that's necessary. 
- I added the `round_counts` option to Kallisto so that I could use it in `build_sce`. I couldn't remember why we needed it, but I didn't think it would hurt and it was easier to round counts across a list rather than when the assays were already in the sce object and keep it only for the alevin function. 
- I think we should be able to remove the `mtx_format` argument now, and also the `read_alevin_mtx` function. We should be able to use `usa_mode` alone to delineate if alevin-fry with USA was used. All other objects should be treated the same. I kept this in for now, but will delete it if we decide that's okay. 

My main other question is if we are okay with the choice to make the unspliced counts the main `counts` assay for all data aligned to the splici index and then storing the spliced counts in the `spliced` assay. For data in ScPCA that are single-cell, this would basically mean the current counts matrix that we have would be moved to `spliced` and the new `counts` assay would be totaled across `U`, `S`, and `A` labeled genes. The only other thought I had was if we also wanted to include a matrix with just `U` so users could actually do RNA velocity... but then we have 3 copies of the matrix and maybe things would be getting a little big.


R/build_sce.R
```diff
@@ -0,0 +1,91 @@
...
+
+    # no intron mode then just collapse
+    # spliced <- collapse_intron_counts(counts, which_counts = c("spliced"))
+    assay_list <- list(counts = counts)
+
+  }
+
+  if (round_counts){
+
+    assay_list <- purrr::map(assay_list, round)
```
jashapiro, line 83: I love purrr, but for this case, to avoid the dependency (though it is probably there anyway), I might just use `lapply`
```suggestion
    assay_list <- lapply(assay_list, round)
```

R/read_alevin.R
```diff
@@ -4,16 +4,14 @@
 #' @param mtx_format Logical indicating if input data is in matrix market format.
 #'   Default is FALSE.
 #' @param intron_mode Logical indicating if the files included alignment to intronic regions.
+#'   If TRUE, the main "counts" assay will contain counts aligned to introns and an additional "spliced"
+#'   assay will contain counts aligned to spliced cDNA only.
```
jashapiro, line 8: I think we should probably rename this... Not quite sure how. 

I was going to say `spliced_only` or something like that, but in the case where there are no unspliced data, I'm not quite sure what we are planning to do.

Maybe `include_unspliced` since that is what you use later to store this value?

R/build_sce.R
```diff
@@ -0,0 +1,91 @@
...
+#' build_sce(counts)
+#'
+#' # build a SCE object with unspliced cDNA as the main counts assay and spliced
+#' # counts as a second assay
+#' build_sce(counts,
+#'           intron_mode = TRUE)
+#'
+#'}
+build_sce <- function(counts,
+                      intron_mode = FALSE,
```
jashapiro, line 31: Maybe changing this option too?
```suggestion
                      include_unspliced = FALSE,
```

R/build_sce.R
```diff
@@ -0,0 +1,91 @@
...
+
+    # create list of assays to use as input to create SCE object
+    assay_list <- list(counts = unspliced,
+                       spliced = spliced)
+
+  } else {
+
+    # no intron mode then just collapse
+    # spliced <- collapse_intron_counts(counts, which_counts = c("spliced"))
+    assay_list <- list(counts = counts)
```
jashapiro, line 77: I'm confused by this change. I assume we want to get just spliced in this case, in case (like w/ kallisto) there are introns that might have been included in the data matrix.

jashapiro (commented):
This looks like a good start, thanks for tackling it!

I want to think a bit about what the options are that we want to support... Some of this goes back a ways so my memory may be fuzzy. 

Before alevin-fry & USA mode, we had two ways of mapping: with a spliced-only transcriptome and to a reference with the spliced transcriptome plus introns. My memory is that Kallisto would create spliced and intron counts; I don't remember what salmon alevin would do exactly. So we were using collapse_intron_counts to merge things together in whatever case we had, whether spliced + intron or USA. 

If there were no introns, or no unspliced, `collapse_intron_counts` basically does nothing, so calling it should not be a problem. 

Either way, `intron_mode` seems ambiguous in this context. I feel like `include_unspliced` with a TRUE default makes sense, but we might want to add a check and warning if there are no unspliced data (if there are no genes with `-I` or `-U` suffixes, though the latter should be handled by fishpond and never make it to this point).

If this is unclear, my apologies... I'm probably a bit foggy myself at the moment. I would be happy to pair to discuss it a bit more.


Other notes: I think the build_sce function is a good idea, but I made a few suggestions to implementation, notably to ensure that matrices are congruent. One other note there is that I removed some if statements: there is no problem with making a zero-row matrix, so we don't need to test for it.

allyhawkins:
> Before alevin-fry & USA mode, we had two ways of mapping: with a spliced-only transcriptome and to a reference with the spliced transcriptome plus introns. My memory is that Kallisto would create spliced and intron counts; I don't remember what salmon alevin would do exactly. So we were using collapse_intron_counts to merge things together in whatever case we had, whether spliced + intron or USA.

> If there were no introns, or no unspliced, collapse_intron_counts basically does nothing, so calling it should not be a problem.

So the only way we were using `collapse_intron_counts` previously was if we were aligning to the `splici` index (spliced + introns). However if we aligned to just the spliced only transcriptome, we didn't use `collapse_intron_counts`, because we specifically have a check in there that there needs to be genes with either `-I` (how alevin, non-USA alevin-fry, and kallisto label the intronic reads) or `-U` (how USA mode alevin-fry labels their genes). The intent of that function was to only use it if we needed to collapse the matrix due to aligning to the spliced + intron transcriptome. We could change it and re-arrange those checks a little, but I didn't do that because we don't really need to do any adjusting if a spliced-only reference is used. 

> Either way, intron_mode seems ambiguous in this context. I feel like include_unspliced with a TRUE default makes sense, but we might want to add a check and warning if there are no unspliced data (if there are no genes with -I or -U suffixes, though the latter should be handled by fishpond and never make it to this point).

So I actually started working on this by adding an `include_unspliced` argument instead, but we still need a way to know if the splici index is used vs spliced only index, which is what `intron_mode` was for. I almost think we may need both.

R/build_sce.R
allyhawkins: Sorry for the confusion here, this wasn't a change from the original, this was a change from me testing things out and I forgot to remove the commented out line. When I went to test using line 76, the kallisto test fails and anything not aligned to a reference so `intron_mode=FALSE` was failing. So that's why I think we need to distinguish between alignment to introns vs. including introns in the output with the two arguments.

jashapiro:
> So I actually started working on this by adding an `include_unspliced` argument instead, but we still need a way to know if the splici index is used vs spliced only index, which is what `intron_mode` was for. I almost think we may need both.

Can we check for the presence of `-I`/`-U`?

allyhawkins:
Thanks for the earlier comments on this @jashapiro. I went through it again and incorporated the `include_unspliced` argument, fully replacing the need for `intron_mode`. This is set as `TRUE` by default, but will throw an error if you don't have a matrix that's been aligned to a reference that contains unspliced reads by searching for reads ending in either `-I` or `-U`. I put this error check in the `build_sce` function since both `read_alevin` and `read_kallisto` will need that check and they can't do the check until they read in the counts matrix which the immediate next step is building the SCE. 

This also means there are now three options in `build_sce`. If `include_unspliced` is `TRUE` then both matrices are returned. If `include_unspliced` is not true, but there are intronic reads in the counts matrix, `collapse_intron_counts` is called with the option `which_counts= "spliced"`. If `include_unspliced` is `FALSE` and a spliced only reference was used the counts matrix is directly put into the SCE object. 

I think the one thing that is still undecided here is the naming of the assays. I think we definitely want to have the unspliced assay be `counts` and then if not including unspliced the `counts` assay will just contain spliced cDNA. That way either way there is a `counts` assay. However, for the extra spliced only counts I had been using `spliced` as the assay name, but you made a comment about it but I was unclear from the comment if you actually were convinced on changing that.

jashapiro:
> I think the one thing that is still undecided here is the naming of the assays. I think we definitely want to have the unspliced assay be `counts` and then if not including unspliced the `counts` assay will just contain spliced cDNA. That way either way there is a `counts` assay. However, for the extra spliced only counts I had been using `spliced` as the assay name, but you made a comment about it but I was unclear from the comment if you actually were convinced on changing that.

I think I thought that was fine. I wanted to be sure there was an indication in the somewhere in the object about whether the `counts` were spliced or not, but you did already have that in `meta$transcript_type` , so I think we are good. It shouldn't matter most of the time, as I _hope_ people will pick a setting and stick with it!

I do wonder though if we want to say "unspliced" or something different there: "unspliced" implies _not_ spliced, so I think we should probably rename it "total" or something like that?

Oh, I also meant to comment about whether we want a separate "unspliced-only" set at all; I think for most cases we should be fine without it, as someone could get at that with the difference between total and spliced. I looked at the `scVelo` option for fishpond, and it uses S+A and U as those two fields. I had a worry that it might try to do something fancier to do spliced as [S + A(p)] and  [U + A(1-p)] where the ambiguous reads are split between spliced and unspliced sets, but it appears not. If people do want to go that far, they can go back to the RAD files.

allyhawkins:
> I do wonder though if we want to say "unspliced" or something different there: "unspliced" implies not spliced, so I think we should probably rename it "total" or something like that?

To clarify you mean changing `meta$transcript_type <- c("unspliced", "spliced")` to `meta$transcript_type <- c("total", "spliced")`? 

> Oh, I also meant to comment about whether we want a separate "unspliced-only" set at all; I think for most cases we should be fine without it, as someone could get at that with the difference between total and spliced. I looked at the scVelo option for fishpond, and it uses S+A and U as those two fields. I had a worry that it might try to do something fancier to do spliced as [S + A(p)] and [U + A(1-p)] where the ambiguous reads are split between spliced and unspliced sets, but it appears not. If people do want to go that far, they can go back to the RAD files.

I'm good with this plan to leave out the unspliced-only set.

jaclyn-taroni:
Just chiming in to say if I saw/read something that said "unspliced" I would not assume it meant unspliced + spliced (provided I followed all that correctly)

jashapiro:
> To clarify you mean changing `meta$transcript_type <- c("unspliced", "spliced")` to `meta$transcript_type <- c("total", "spliced")`?

Correct. Will get back to the full review later!

R/build_sce.R
```diff
@@ -0,0 +1,113 @@
...
+  if(include_unspliced & !has_unspliced){
+    stop("No counts corresponding to intronic reads detected.
+          If `include_unspliced` is TRUE a reference with spliced and unspliced reads must be used.")
+  }
+
+  # if using intron mode and want to include the unspliced data get counts for both unspliced and spliced
+  if(include_unspliced & has_unspliced) {
+
+    unspliced <- collapse_intron_counts(counts, which_counts = c("unspliced"))
+    spliced <- collapse_intron_counts(counts, which_counts = c("spliced"))
```
jashapiro, line 55: Realizing that going back in time, we would want to change this `which_counts  =  "total"`, but that could break things. But I think we can do it if we modify `collapse_intron_counts` to allow either option with the same result, that is probably worth doing.

So for consistency, the following suggestion, and the downstream effects thereof, unfortunately.

```suggestion
    total <- collapse_intron_counts(counts, which_counts = "total")
    spliced <- collapse_intron_counts(counts, which_counts = "spliced")
```

jashapiro (commented):
This looks good! My only remaining thought was that the use of `unspliced` in `build_sce()` and `collapse_intron_counts()` is probably something we should try to get away from, since it is a bit confusing. I think it is probably worth doing in this PR, but if you disagree, we can tackle it later.

I'll also note that I think the change to `read_alevin` is going to break `scpca-nf`, so we will need an update there to use the new version of the function as soon as this goes in.

allyhawkins:
> This looks good! My only remaining thought was that the use of unspliced in build_sce() and collapse_intron_counts() is probably something we should try to get away from, since it is a bit confusing. I think it is probably worth doing in this PR, but if you disagree, we can tackle it later.

I'm on board with this so went ahead and made this change which was actually way simpler than I thought. We don't actually use `unspliced` for anything in `collapse_intron_counts`... do we want to change how that argument is set up all together? The default mode is to get the total counts unless `spliced` is used. 

I also went ahead and removed the `mtx_format` option and the associated function since we don't need that anymore. And I updated the main `import_quant_data` function since we don't test that, just all the other functions it runs. I then updated the main readme since we mention using that function and use of `which_counts`. 

> I'll also note that I think the change to read_alevin is going to break scpca-nf, so we will need an update there to use the new version of the function as soon as this goes in.

PR on this incoming!

R/collapse_intron_counts.R
```diff
@@ -19,13 +19,14 @@
...
 #' collapse_intron_counts(counts,
-#'                        which_counts = "unspliced")
+#'                        which_counts = "total")
 #' }
 #'
 #'
 #'
 collapse_intron_counts <- function(counts,
-                                   which_counts = c("spliced", "unspliced")){
+                                   which_counts = c("spliced", "unspliced", "total")){
```
jashapiro, line 28:
```suggestion
                                   which_counts = c("spliced", "total", "unspliced")){
```

R/collapse_intron_counts.R
```diff
@@ -19,13 +19,14 @@
...
+
   which_counts <- match.arg(which_counts)
```
jashapiro, line 31: I feel like we should warn people about the fact that "unspliced" not a good choice. 
I don't love my wording here, but I want something...

```suggestion
  which_counts <- match.arg(which_counts)

  if (which_counts ==  "unspliced"){
    warn("which_counts = 'unspliced' is deprecated, as its behavior is ambiguous (and included both spliced and unspliced counts) in previous versions"
```

tests/testthat/test-read_alevin.R
```diff
@@ -11,7 +11,8 @@ sample_ids = c("A1", "A2")
 
 test_that("reading salmon alevin data works", {
   sce <- read_alevin(alevin_dir,
-                     sample_id = sample_ids )
+                     sample_id = sample_ids,
+                     include_unspliced = FALSE)
```
jashapiro, line 15: Can you add `expect_equal(sce@metadata$transcript_type, c("spliced"))` to this test?

tests/testthat/test-read_kallisto.R
```diff
@@ -5,7 +5,8 @@ kallisto_dir <- file.path(dir, "Breast_Cancer_3p_LT/kallisto_txome")
 sce_size <- c(60666, 30339)
 
 test_that("Check that kallisto import works", {
-  sce <- read_kallisto(kallisto_dir)
+  sce <- read_kallisto(kallisto_dir,
```
jashapiro, line 8: can we add testing the `mapping_tool` and `transcript_type` here too (if we set it?)

jashapiro (commented):
Okay, almost... I have a few more suggestions here for more consistency and a bit of warning and some more testing...

tests/testthat/test-read_kallisto.R
allyhawkins: We don't add any metadata to the kallisto objects, but we could?

jashapiro: I think we should... Since the transcript type in `counts` can vary, it would be a nice update to have it in the object, I think.

allyhawkins:
I removed the `unspliced_only_genes`/`total_only_genes` after some testing because you're right we should only have spliced only genes not the other way around. That should be updated now. 

I also added some metadata to the kallisto objects since we didn't have that before and included the `transcript_type`. I also updated the tests to reflect this.

allyhawkins:
> I removed the unspliced_only_genes/total_only_genes after some testing because you're right we should only have spliced only genes not the other way around. That should be updated now.

I had it backwards and was accidentally reducing the dimensions. There are unspliced only genes or total only genes making the dimensions of that larger than the spliced dimensions. I think we do want to keep both in there which also ensures the matrices are always in the same order.

jashapiro:
> > I removed the unspliced_only_genes/total_only_genes after some testing because you're right we should only have spliced only genes not the other way around. That should be updated now.
> 
> I had it backwards and was accidentally reducing the dimensions. There are unspliced only genes or total only genes making the dimensions of that larger than the spliced dimensions. I think we do want to keep both in there which also ensures the matrices are always in the same order.

It doesn't hurt, to be sure. (Out of curiousity: was that in real data, or the made up data in the test suite? I can imagine I added an unspliced-only gene in the test even though I can't imagine that in reality!)

jashapiro (approved):
LGTM! Thanks for all the updates!

allyhawkins:
> It doesn't hurt, to be sure. (Out of curiousity: was that in real data, or the made up data in the test suite? I can imagine I added an unspliced-only gene in the test even though I can't imagine that in reality!)

That was in real data. I was testing things with one of the alevin-fry runs from benchmarking.

PR #147, opened 2022-11-22
Account for feature data in `read_alevin`

allyhawkins:
This PR makes some slight changes to `read_alevin` to account for being able to read in both feature data and RNA data. We were previously using an option for `usa_mode` and if it was true, `fishpond::loadFry` was used to load in alevin-fry data, accounting for spliced and unspliced reads. Part of this relied on the data being in USA mode, which is not the case for feature data.

I double checked that this function does work with both USA and non-USA data as there is a check at the beginning of the function that grabs the mode from the json files. However, we were explicitly telling `loadFry` to store either spliced reads only or unspliced and spliced reads in the counts matrix which doesn't really apply to feature data and isn't really what we want. We would instead just want all of the reads to be totaled and stored in `counts` without any additional `spliced` assay.

To account for this, I made a few updates: 

- Changed the option from `usa_mode` to `fry_mode`, which now serves just as a way to say that this data was quantified using alevin-fry and therefore can be read in using `loadFry` rather than specifically referring to USA mode. 
- Added a `feature_data` option which is by default set to `FALSE`. This signifies if the data is feature data or not and if that's the case, the `transcript_type` is set to `feature_counts` and there is only one assay returned, `counts`. This is the case for both alevin and alevin-fry quantified data. 
- I also added a check that both `include_unspliced` and `feature_data` cannot be true at the same time. 

I tested this with some feature data and it should work as expected, so I think this should handle the concerns mentioned in https://github.com/AlexsLemonade/scpca-nf/pull/242#pullrequestreview-1190089579.

### DIFF
--- a/R/read_alevin.R
+++ b/R/read_alevin.R
@@ -1,7 +1,7 @@
 #' Read in counts data processed with Alevin or Alevin-fry
 #'
 #' @param quant_dir Path to directory where output files are located.
-#' @param fry_mode Logical indicating if Alevin-fry was used, if USA mode was invoked.
+#' @param fry_mode Logical indicating if Alevin-fry was used for quantification.
 #'   Implies the input data is in matrix market format.
 #'   Default is FALSE.
 #' @param include_unspliced Whether or not to include the unspliced reads in the counts matrix.
@@ -96,7 +96,7 @@ read_alevin <- function(quant_dir,
     if(include_unspliced){
       assay_formats <- list("counts" = c("S", "A", "U"), "spliced" = c("S", "A"))
       meta$transcript_type <- c("total", "spliced")
-    } else if(!include_unspliced & !feature_data) {
+    } else if(!feature_data) {
       assay_formats <- list("counts" = c("S", "A"))
       meta$transcript_type <- "spliced"
     } else {
@@ -119,7 +119,7 @@ read_alevin <- function(quant_dir,
     if(include_unspliced){
       meta$transcript_type <- c("total", "spliced")
 
-    } else if(!include_unspliced & !feature_data) {
+    } else if(!feature_data) {
       meta$transcript_type <- "spliced"
     } else {
       meta$transcript_type <- "feature_counts"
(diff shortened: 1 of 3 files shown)


R/read_alevin.R
```diff
@@ -1,16 +1,19 @@
 #' Read in counts data processed with Alevin or Alevin-fry
 #'
 #' @param quant_dir Path to directory where output files are located.
-#' @param usa_mode Logical indicating if Alevin-fry was used, if USA mode was invoked.
+#' @param fry_mode Logical indicating if Alevin-fry was used, if USA mode was invoked.
```
jashapiro, line 4:
```suggestion
#' @param fry_mode Logical indicating if Alevin-fry was used for quantification.
```

R/read_alevin.R
```diff
@@ -73,28 +82,35 @@ read_alevin <- function(quant_dir,
...
+        stop("Output files not in USA mode")
+      }
     }
 
     # define assays to include in SCE object based on include_unspliced
     if(include_unspliced){
       assay_formats <- list("counts" = c("S", "A", "U"), "spliced" = c("S", "A"))
       meta$transcript_type <- c("total", "spliced")
-    } else {
+    } else if(!include_unspliced & !feature_data) {
```
jashapiro, line 99: We already know that include_unspliced is false, since we checked it in the first if...
```suggestion
    } else if(!feature_data) {
```

R/read_alevin.R
```diff
@@ -103,8 +119,10 @@ read_alevin <- function(quant_dir,
     if(include_unspliced){
       meta$transcript_type <- c("total", "spliced")
 
-    } else {
+    } else if(!include_unspliced & !feature_data) {
```
jashapiro, line 122:
```suggestion
    } else if(!feature_data) {
```

jashapiro (approved):
Looks good! I just (arguably) simplified a bit of logic, and updated one comment.

PR #148, opened 2022-11-22
Function to combine SCE objects

### LINKED ISSUE #146: Function for combining SingleCellExperiment objects
Before being able to do integration, we will need to be able to merge SCE objects into one object to obtain combined highly variable genes and calculate PCA. Additionally, the input for the integration methods we plan to support use a merged SCE object. 

This function should take as input a list of individual SCE objects, all already containing normalized data. Inside the function we will need to match up all of the `colData`, filling in any data that is only found in part of the libraries being integrated with `NA`. We will also want to subset to only shared genes. The new columns of this merged object should be named with the cell barcode and the library ID as barcodes could be overlapping between libraries. 

The output should be the merged object with a new column in the `colData` corresponding to the library ID for each cell.

sjspielman:
Closes #146 

This PR adds a function `merge_sce_list()` which will format a named list of SCEs and combine them with `cbind()` using similar approaches taken for data integration preparation, but with lots of `purrr` refactoring 🎉

I'm filing this as a draft PR to get some initial feedback on the general approach and discuss some ways to deal with bugs. Note there are also no unit tests yet, but those will be written once we're all on board with the function itself.

Current functionality:
 
- subsets to shared features
- rowData column names are updated to `colname-**library**`, except for those specified in `preserve_rowdata_columns` to remain unchanged
- only specified colData columns should be retained. Trickily, this has to be two arguments for the main (`retain_coldata_columns_main`)  vs altExps (`retain_coldata_columns_altexps`), _but_ there is a lot of gnarliness with altExps so keep reading!
- add in a `batch_column` column with library name

As I hinted at the main problem at this point is how we deal with altExps. Currently there is a flag indicating whether to keep or discard them, but the logic is turned off for now because `cbind` doesn't want to merge SCEs with altExps, even if it's all matching. So I see two ways we can deal with this:
- Actually remove altExps no matter what
- We'd have to first merge the main experiment, and then merge any altExps that are meant to match up (there may be >1 altExp!) as "stand-alone" SCEs, and then place this merged altExp SCE into the altExp slot of the merged main experiment SCE. This involves quite a bit of bookkeeping and an input arguments (probably a named list?) for specifying:
  - altExp names that should match up for cbind'ing (list names)
  - column names to retain for _each_ of those altExps. (vector for each provided list name)

I've been testing this code out for now with - 
```
# libraries with CITEseq for exploring altExps functionalitu
path <- "/your/path/sc-data-integration/results/scpca/citeseq_sce"
sce1 <- readr::read_rds(file.path(path, "SCPCS000221/SCPCL000295_processed_citeseq.rds"))
sce2 <- readr::read_rds(file.path(path, "SCPCS000222/SCPCL000296_processed_citeseq.rds"))

sce_list <- list(
  "SCPCL000295" = sce1,
  "SCPCL000296" = sce2
)

# helpful function arguments
retain_coldata_columns_altexps <- "sizeFactor"
retain_coldata_columns_main <- c("louvain_10", "miQC_pass")
preserve_rowdata_columns <- "gene_symbol"
```

I'll also note that Roxygen got bumped probably when I ran `devtools::document()`? Should I drop this back down? I think though I vaguely remember that we just kind of embrace a back and forth on this, maybe?


jashapiro:
I'm working on a more complete intial review, but as a quick note, in case you wanted to work on this some today, we will want to add unit tests for this function. I would expect those to test the basic join case, as well as some of the more complex cases with ADTs etc. You can make a quick stub file with `usethis::use_test()` in the console when you have the source file open.

R/merge_sce_list.R
```diff
@@ -0,0 +1,138 @@
...
+                           retain_coldata_columns_altexps = c(),
+                           preserve_rowdata_columns = NULL) {
+  # Ensure `sce_list` is named  ----------------------
+  if (is.null(names(sce_list))) {
+    stop("Individual SingleCellExperiment objects in `sce_list` must be named.")
+  }
+  # Ensure `sce_list` has >=2 items --------------------------------------------
+  if (length(sce_list) < 2) {
+    stop("The `sce_list` must contain 2 or more SingleCellExperiment objects to merge.")
+  }
```
jashapiro, line 51: Again, can we handle this with a warning? If there is only one thing in the list, just return it!

R/merge_sce_list.R
```diff
@@ -0,0 +1,138 @@
...
+
+  # Subset SCEs to shared features and ensure appropriate naming ------------------
+
+  # First, obtain intersection among all SCE objects
+  shared_features <- sce_list |>
```
jashapiro, line 56: new pipe ftw.

R/merge_sce_list.R
```diff
@@ -0,0 +1,138 @@
...
+    purrr::map(rownames) |>
+    purrr::reduce(intersect)
+
```
jashapiro, line 59: Probably worth having a check here that there are not zero shared features. _This_ would be a stopping error.

sjspielman:
> we will want to add unit tests for this function.

Yes for sure and I am very familiar with `testthat`. As noted in my original comment I haven't started on testing yet because I want to work out the ADT situation first, but I do plan to start adding tests for simpler SCEs that have no altExps.

R/merge_sce_list.R
```diff
@@ -0,0 +1,138 @@
...
+
+  process_columns <- function(sce, sce_name, retain_columns) {
```
jashapiro, line 61: I'd pull this out to a separate function, maybe with a more descriptive name. What kind of processing are we doing here? It would still have a doxygen comments, just no  `#' @export` line, so it is available only inside the package (by default).

R/merge_sce_list.R
```diff
@@ -0,0 +1,138 @@
...
+      original_colnames,
+      glue::glue("{original_colnames}-{sce_name}")
+    )
+
+    ##### colData #####
+    # Retain only the columns present in `retain_columns`
+    coldata_names <- names(colData(sce))
+    if (!(all(retain_columns %in% coldata_names))) {
+      stop("Error: Provided columns in `retain_coldata_columns` are not present
+            in all SCE objects.")
```
jashapiro, line 81: Again, I'd probably rather handle this with a warning, filling in any missing values with `NA`. This one might be more of a judgement call.

R/merge_sce_list.R
```diff
@@ -0,0 +1,138 @@
...
+    # helper function to rename and subset columns in an SCE object when
+    # formatting for merging
+
+    ##### rowData #####
+    # Add `sce_name` ID to rowData column names except for those
+    #  present in `preserve_rowdata_columns`
+    original_colnames <- colnames(rowData(sce))
+
+    colnames(rowData(sce)) <- ifelse(
+      original_colnames %in% preserve_rowdata_columns,
```
jashapiro, line 71: `preserve_rowdata_columns` is not an argument to the function, but it should be. (I know that this works because it is an internal function,  but I am hugely not a fan of functions going outside their defined scope, no matter what. It makes for unexpected results too often.)

R/merge_sce_list.R
```diff
@@ -0,0 +1,138 @@
...
+    # Retain only the columns present in `retain_columns`
+    coldata_names <- names(colData(sce))
+    if (!(all(retain_columns %in% coldata_names))) {
+      stop("Error: Provided columns in `retain_coldata_columns` are not present
+            in all SCE objects.")
+    }
+    colData(sce) <- colData(sce) |>
+      as.data.frame() |>
+      dplyr::select( dplyr::all_of(retain_columns) ) |>
+      DataFrame()
```
jashapiro, line 86: This double conversion seems excessive when the default method is pretty easy.
```suggestion
    colData(sce) <- colData(sce)[, retain_columns]
```

R/merge_sce_list.R
```diff
@@ -0,0 +1,138 @@
...
+
+    # Add batch column
+    sce[[batch_column]] <- sce_name
```
jashapiro, line 89: Ooh, I didn't know you could do this! Nifty.  

But `batch_column` will need to be a function argument too.

R/merge_sce_list.R
```diff
@@ -0,0 +1,138 @@
...
+
+
+    return(sce)
+  }
+
+
+
+  prepare_sce_for_merge <- function(sce, sce_name) {
```
jashapiro, line 97: Move out to its own function too! Pass in `shared_features` and any other needed variables.

R/merge_sce_list.R
```diff
@@ -0,0 +1,138 @@
...
+    # helper function for preparing SCE for merging
+
+    # If specified, remove altExps
+    #if (retain_altexps == FALSE) {
+      sce <- removeAltExps(sce)
+    #}
```
jashapiro, line 103: I expect that we will want to handle altExps better, but I would do this in a separate PR. For now, file an issue, blocked by this issue, to add support for altExps in merging.

R/merge_sce_list.R
```diff
@@ -0,0 +1,138 @@
...
+
+    # Subset to shared features
+    sce <- sce[shared_features,]
+
+    # Add `sce_name` to colData row names so cell barcodes can be mapped to originating SCE
+    # Note that this will also affect altExps
+    colnames(sce) <- glue::glue("{colnames(sce)}-{sce_name}")
```
jashapiro, line 110: `barcode` is not currently a default column, but we might want to add it. I like the idea retaining the barcodes as their own thing to save any potential parsing. It's not really important, but kind of nice. More importantly, I might make `cell_id` a full column in `colData`, not just in the `colnames`, as it saves effort when pulling the coldata into tidyverse...

R/merge_sce_list.R
```diff
@@ -0,0 +1,138 @@
...
+
+    # Rename rowData and subset colData columns for both main and any present altExps
+    sce <- process_columns(sce, sce_name, retain_coldata_columns_main)
+    for (altexp_name in altExpNames(sce)) {
+      altExp(sce, altexp_name) <- process_columns( altExp(sce, altexp_name),
+                                                   sce_name,
+                                                   retain_coldata_columns_altexps)
+    }
```
jashapiro, line 118: Since we are removing sces for now, drop this chunk.
```suggestion
```

jashapiro (commented):
This looks like a good start. 

My biggest comment is to rearrange so the helper functions are fully separated out; I don't like function definitions within functions for a few reasons, one of which is that it makes things harder to test.  If the helper functions are separate, you can write separate tests for each of them.

I had a few other comments throughout, many of which ended up being suggestions to use warnings instead of errors when possible.

And of course, my earlier comment: tests! I think we have found that writing tests for these kinds of functions is really helpful, both for making sure that it does what we expect, and for catching later changes that might break things in unexpected ways.

jashapiro:
Just to pull it out... I think the way to deal with altExps is to remove them all for the first PR, and think about how to "properly" do it later. 

My guess is that you will need to separate out the altExps into their own list (list of lists if there is more than one?) and then merge those separately with `cbind`, followed by adding them back to the merged main sce. And we would only want to do this if _all_ samples had the same ADTs, with the same features.

R/merge_sce_list.R
```diff
@@ -0,0 +1,138 @@
...
+#'  not specific to the given library's preparation or statistics. For example,
+#'  such an array might contain items like "Gene", "ensembl_ids", "gene_symbol", etc.
+#'
+#' @return A SingleCellExperiment object containing all SingleCellExperiment objects
+#'   present in the inputted list
+#' @export
+#'
+#' @import SingleCellExperiment
+merge_sce_list <- function(sce_list = list(),
+                           batch_column = "batch",
```
allyhawkins, line 33: I'm torn on this... if you remember our lengthy discussion on what a batch is, I wonder if this is confusing or if we should this opportunity to switch to `library_id` instead? Technically we are using the `library_id` and putting it into a column named `batch`, but everywhere else in `scpca-nf` and in `scpcaTools` we refer to it with `library_id` which is why I might slightly favor changing to that over batch...

R/merge_sce_list.R
```diff
@@ -0,0 +1,138 @@
...
+                           retain_altexps = TRUE,
+                           retain_coldata_columns_main = c("sum",
+                                                           "detected",
+                                                           "total",
+                                                           "subsets_mito_sum",
+                                                           "subsets_mito_detected",
+                                                           "subsets_mito_percent",
```
allyhawkins, line 40:
```suggestion
                                                           "subsets_mito_percent"
                                                           "prob_compromised",
```
I think we probably also want to keep the probability compromised values determined from miQC.

R/merge_sce_list.R
allyhawkins: I'm going to agree with this. And I also think we might want to consider having a separate function specifically for handling if an object has altExp data and probably merging that and adding it as a second assay like you mentioned in your initial PR.

R/merge_sce_list.R
allyhawkins: Fun fact about now using `fishpond::loadFry` is that the barcodes are now stored in their own column in the `colData`.

allyhawkins (commented):
Most of my comments align with what Josh already had noted, so just added a few more thoughts.

R/merge_sce_list.R
```diff
@@ -77,13 +86,16 @@ merge_sce_list <- function(sce_list = list(),
...
             in all SCE objects.")
     }
     colData(sce) <- colData(sce) |>
       as.data.frame() |>
       dplyr::select( dplyr::all_of(retain_columns) ) |>
-      DataFrame()
+      S4Vectors::DataFrame()
+
+    # Update rownames to include the sce_name
+    rownames(colData(sce)) <- glue::glue("{rownames(colData(sce))}-{sce_name}")
```
jashapiro, line 98: This should be handled by the later change to `colnames(sce)`, shouldn't it?

R/merge_sce_list.R
```diff
@@ -0,0 +1,138 @@
...
+#'
+#' @return A SingleCellExperiment object containing all SingleCellExperiment objects
+#'   present in the inputted list
+#' @export
+#'
+#' @import SingleCellExperiment
+merge_sce_list <- function(sce_list = list(),
+                           batch_column = "batch",
+                           retain_altexps = TRUE,
+                           retain_coldata_columns_main = c("sum",
```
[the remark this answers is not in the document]
sjspielman: Now that altexps is temporarily out of play, I've renamed this to `retain_coldata_cols` (for main only) and there is only 1 argument.

R/merge_sce_list.R
```diff
@@ -0,0 +1,138 @@
...
+                                                           "total",
+                                                           "subsets_mito_sum",
+                                                           "subsets_mito_detected",
+                                                           "subsets_mito_percent",
+                                                           "miQC_pass"),
+                           retain_coldata_columns_altexps = c(),
+                           preserve_rowdata_columns = NULL) {
+  # Ensure `sce_list` is named  ----------------------
+  if (is.null(names(sce_list))) {
+    stop("Individual SingleCellExperiment objects in `sce_list` must be named.")
```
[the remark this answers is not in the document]
sjspielman: > this should still work, using the order they were provided as the names

So you'd envision populating the batch column with indices in that scenario? Want to make sure I understand your comment but can do if that's what you're thinking!

R/merge_sce_list.R
sjspielman: Interesting.. In my mind `batch` is a much more general term to use here than `library_id` but consistency within the ScPCA universe is important! 
My reading of your comment is to change the default value to "library_id" but keep the argument `batch_column` name? Yes/no?

R/merge_sce_list.R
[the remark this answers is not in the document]
jashapiro: How do you feel about about dropping the `retain`?I feel like it is maybe a bit redundant? Also,`retain` feels like a TRUE or FALSE question to me, and that is how it is used with altExps in the other (currently postponed?) option.

R/merge_sce_list.R
allyhawkins: Yes, I think that's what I'm getting to. I just want to make sure we have an understanding that "batch" here is in reality just the library ID, or what we have been using as library ID throughout scpca. So I like that solution of `batch_column` with the default of `library_id`.

R/merge_sce_list.R
sjspielman: I think this may not actually work. It seems like if there is only _one_ column in `retain_cols` then this object simply becomes a vector which isn't great! Do you see that on your end too?

R/merge_sce_list.R
sjspielman: > Fun fact about now using `fishpond::loadFry` is that the barcodes are now stored in their own column in the `colData`.

Interesting.. We'll need an argument for `barcode_column_name` which we can use to check if barcodes are present already so we don't add them in again.

R/merge_sce_list.R
sjspielman: Ok, I see that if it's length==1 then you have to do something like this for example 

```
retain_cols <- "sum"
S4Vectors::DataFrame(sum = colData(sce)[,"sum"])
```

but I'm not having success with any form of eval to programmatically get the column name. Unless you have other thoughts I may revert back to the `dplyr` detour approach

R/merge_sce_list.R
sjspielman: I'm good with going back to this, on the way.

R/merge_sce_list.R
jashapiro: I think you can do 
```
colData(sce)[, cols, drop=FALSE]
```

sjspielman:
I believe I've gotten most of the reviews in except for the column subsetting situation which I reverted (for now at least) back to `dplyr`. All the (21! ) tests pass on my side and docs are updated so `CMD check` _should_ complain less (is saying `CMD check` should complain less the kiss of death? probably). Note that if this is too many tests we can purge some but they're all quick, cover lots of ground, and are not very redundant so I'm not worried much.

Some highlights of the changes I made:
- The `batch_column` default is now "library_id"
- There is now a `barcode_column` argument. This column either exists already and is retained no matter what, or it is created (and retained)!
- Early return with warning if only 1 SCE is in the list
- Error out if there are no shared features
- Columns are allowed to be missing and they are created with `NA` values. However, columns specified in `retain_coldata_cols` must be present somewhere in at least 1 SCE, or we error out. It doesn't make sense to return a bunch of `NA` columns.
- `prepare_sce_for_merge()` is its own function with docs and separate testing, but all tests are in the same `testthat` script to point back to this same `R/` script.

I've re-read over all my docs in hopes they are all correct for the current code, where _in hopes_ is doing a lot of work since my eyes have begun to glaze over as we approach 🦃 .

R/merge_sce_list.R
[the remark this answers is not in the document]
jashapiro: Yes, that is what I was thinking.

R/merge_sce_list.R
```diff
@@ -0,0 +1,199 @@
...
+  colData(sce) <- colData(sce) |>
+    as.data.frame() |>
+    dplyr::select( dplyr::all_of(retain_coldata_cols) ) |>
+    S4Vectors::DataFrame()
+  # this approach doesn't work when retain is length 1:
+  #colData(sce) <- colData(sce)[, retain_coldata_cols]
+
+  # Add `sce_name` to colData row names so cell barcodes can be mapped to originating SCE
+  original_rownames <- rownames(colData(sce))
+  rownames(colData(sce)) <- glue::glue("{rownames(colData(sce))}-{sce_name}")
```
jashapiro, line 193: I think we want to just update the whole SCE here?
I don't think you use `original_rownames` so I dropped it.

For organization, I might move this to be the first step of the `prepare` function, as it is one of the most important parts...

```suggestion
  # Add `sce_name` to colnames so cell ids can be mapped to originating SCE
  colnames(sce) <- glue::glue("{colnames(sce)}-{sce_name}")
```

tests/testthat/test-merge_sce_list.R
```diff
@@ -0,0 +1,251 @@
...
+add_sce_data <- function(sce) {
+
+  # add some coldata columns
+  colData(sce)[["sum"]] <- runif(nrow(colData(sce)))
+  colData(sce)[["detected"]] <- runif(nrow(colData(sce)))
+  colData(sce)[["total"]] <- runif(nrow(colData(sce)))
+
+  # add some rowdata columns
+  rowData(sce)[["gene_names"]] <- rownames(rowData(sce))
+  rowData(sce)[["other_column"]] <- runif(nrow(rowData(sce)))
```
jashapiro, line 11: Minor, but removing some redundancy
```suggestion
  # add some coldata columns
  colData(sce)[["sum"]] <- runif(ncol(sce))
  colData(sce)[["detected"]] <- runif(ncol(sce))
  colData(sce)[["total"]] <- runif(ncol(sce))

  # add some rowdata columns
  rowData(sce)[["gene_names"]] <- rownames(sce)
  rowData(sce)[["other_column"]] <- runif(nrow(sce))
```

R/merge_sce_list.R
```diff
@@ -0,0 +1,199 @@
...
+#'
+#' @return A SingleCellExperiment object containing all SingleCellExperiment objects
+#'   present in the inputted list
+#' @export
+#'
+#' @import SingleCellExperiment
+merge_sce_list <- function(sce_list = list(),
+                           batch_column = "library_id",
+                           barcode_column = "barcode",
+                           retain_coldata_cols = c("sum",
```
jashapiro, line 39: Retaining barcode explicitly rather than implicitly.
```suggestion
                           retain_coldata_cols = c("barcode,
                                                   "sum",
```

R/merge_sce_list.R
```diff
@@ -0,0 +1,199 @@
...
+  all_colnames <- purrr::map(sce_list, ~names(colData(.))) |>
+    unlist() |>
+    unname() |>
+    unique()
+
+  # Check that the `retain_coldata_cols` are present in at least one SCE, and
+  #  error if the column exists nowhere.
+  if (!(any(retain_coldata_cols %in% all_colnames))) {
+    stop("Error: The provided `retain_coldata_cols` are not present in any SCEs.")
+  }
```
jashapiro, line 94: The comment makes it sound like you are checking that _all_ retain columns are found in at least one SCE, but the check is for at least one of the columns. Is that correct? I think it makes sense the way the comment is written, so `all()` should work here.

R/merge_sce_list.R
```diff
@@ -0,0 +1,199 @@
...
+                          barcode_column,
+                          shared_features,
+                          # ensure that barcodes are retained
+                          c(barcode_column, retain_coldata_cols),
+                          preserve_rowdata_cols,
+                          all_colnames)
+
+
+  # Create the merged SCE from the processed list ------------------
+  merged_sce <- do.call(cbind, sce_list)
```
jashapiro, line 109: I prefer to name all purrr arguments that are being passed to the function, just in case the underlying function order changes. Which it shouldn't here, but still...

If we want to retain barcodes by default, I think that should be in the main argument list (commented there as well)

Since I added a lot of verbiage, I added a pipe for formatting...

```suggestion
  # Prepare SCEs
  sce_list <- sce_list |>
    purrr::imap(prepare_sce_for_merge,
                batch_column = batch_column,
                barcode_column =  barcode_column,
                shared_features = shared_features,
                retain_coldata_cols = retain_coldata_cols,
                preserve_rowdata_cols = preserve_rowdata_cols,
                expected_coldata_names = all_colnames) 


  # Create the merged SCE from the processed list ------------------
  merged_sce <- do.call(cbind, sce_list)
```

jashapiro (commented):
I added some more comments, with some just about style preferences and some a bit more substantive.

One of the more substiantive questions is about the of the `expected_coldata_names` argument. It seems like this is only used to find missing columns, but then those are immediately filtered down to retained columns. I may be missing something here.

The other is about the barcode behavior: If barcodes are present, we should retain them, and possibly use them for the cell ids (`colname(sce)`), but I would not add them if they are missing, as there is no particular guarantee that the cell id for a generic SCE is a barcode (for CellRanger output it is not, or not _only_ the barcode). I think what I was thinking before is that if there is a barcode, we might want the cell ids to be `{barcode}-{library}`, but in the absence of a barcode, just `{colnames(sce)}-{library}`.

Relatedly, we should only retain the `barcode` column if explicitly asked. We could add a `cell_id` column that is either `colnames(sce)` or `barcode`, which I think is what I was thinking about before knowing that there was a barcode column, but I'm not sure whether that is worth doing.

R/merge_sce_list.R
```diff
@@ -0,0 +1,199 @@
...
+    }
+  }
+
+  # Retain only the columns present in `retain_coldata_cols`
+  colData(sce) <- colData(sce) |>
+    as.data.frame() |>
+    dplyr::select( dplyr::all_of(retain_coldata_cols) ) |>
+    S4Vectors::DataFrame()
+  # this approach doesn't work when retain is length 1:
+  #colData(sce) <- colData(sce)[, retain_coldata_cols]
```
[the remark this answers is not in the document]
jashapiro: Well you got to this one already!

R/merge_sce_list.R
```diff
@@ -0,0 +1,199 @@
...
+#'
+#' @return An updated SCE that is prepared for merging
+prepare_sce_for_merge <- function(sce,
+                                  sce_name,
+                                  batch_column,
+                                  barcode_column,
+                                  shared_features,
+                                  retain_coldata_cols,
+                                  preserve_rowdata_cols,
+                                  expected_coldata_names) {
```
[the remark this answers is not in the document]
sjspielman: I've been thinking about why I did it this way, and while I have some thoughts about my original thought process, my current thought is that it was heavily pre-vacation induced and should really be cleaned up!!

R/merge_sce_list.R
sjspielman: I accepted this but I think I'd rather change it a bit. Since the `barcode_column` itself is an argument, I don't want to hardcode a default value here. Instead I'm going to just update `retain_coldata_cols` in the code to add in whatever the `barcode_column` value is.

jashapiro: I'd rather not do that. My thought is that the code should not modify the `retain_coldata_cols` at all. If the user is changing the default `barcode_column`, then they may need to change `retain_coldata_cols`, and the function could check for that (it already kind of does, in that if the `"barcode"` column does not exist, it will complain), but adding a column to retain that the user did not ask for seems like an odd behavior.

Do need to add a missing quote to my suggestion though, at the least!

sjspielman: @jashapiro Ok that said I'm getting confused with some of the feedback here on barcodes, so I'm going to try and suss it out here:

1.  Per your comment we should only "retain the `barcode` column if explicitly asked", I should definitely undo this accepted change. This means I should not, as I wrote, "update `retain_coldata_cols` in the code to add in whatever the `barcode_column` value is." If the user wants it they have to tell us `retain_coldata_cols`.
2. We need _all_ SCEs to either have or not have a barcode column, so if some are missing barcodes but some are not, we can't not add anything at all. We have to either make `NA` barcode column(s) where missing or remove barcode column(s) where present. We certainly do not want to be _removing_ any cell-identifying information though!
3. So, be it a barcode specifically or some other unique identifier (without presuming what exactly this information is), we need some kind of cell IDing information no matter what. An overall argument `cell_id_column` might be the way to go (rather than `barcode_column` which has an expectation about the type of information), and in the docs explain these _could_ be barcodes. We could then handle this column in a couple different ways:
- Simply require it. Users have to tell us unique cell ID'ing information. This avoids the trouble of what if some SCEs have cell IDs and others don't. I think this scenario is less likely than SCEs being consistent, but it certainly doesn't seem _un_likely that this may happen.
- If missing in any SCE, fill it in with `colnames(sce)`. This _may_ mean different forms of IDs if some SCEs have barcodes and some don't if some were originally missing.

I think we're on the same page now? Any preference for requiring cell info vs. creating it? My preference is to require it.

jashapiro: Agree mostly with point 1, except to say that the change I suggest was to the default value for the argument. So in my mind accepting the default value is "explicitly asked", but I can see that might not be a general view!

Agree on point 2, and yes, we would want to add `NA` values. Which is what we do for other retained columns without full data, correct?

The `cell_id_column` that you suggest is, I think, the `colnames` and I don't think we need to add a separate column for it; However, this is the one where we may need to be careful. It should either be the current  `colnames` + `batch_id` or `barcode_id` + `batch_id`. I don't think it should really matter either way and if they are not 100% consistent, I don't think that is a problem, so my initial thought was to use the barcode if it exists and the colname if it doesn't. (Later thought: we might reverse the order in that ID so that batch comes first)

sjspielman: > Agree mostly with point 1, except to say that the change I suggested was to the default value for the argument. So in my mind accepting the default value is "explicitly asked", but I can see that might not be a general view!

Good point! I'm going to flesh this out alongside the barcode situation, keep reading...

> Agree on point 2, and yes, we would want to add `NA` values. Which is what we do for other retained columns without full data, correct?

Correct! But now I'm circling back to disagreeing with myself and maybe advocating for NOT filling in with `NA`.🙃 

I think we have to require barcode consistency among their input SCEs. They either all have a barcode, or none have a barcode. If it varies among SCEs, we should error out. Any variation will prevent consistency among `cell_id_column` values. For example, if only some SCEs have barcodes, then we'll end up in either of these not great circumstances:
  -  cell IDs are `{batch}-NA` a `{batch}-{barcode} ` naming scheme (noting I like the idea of batch first for this value). This is not great.
  - Some SCEs have a `cell_id_column` scheme of `{batch}-{barcode}` but others may have scheme of `{batch}-{rowname}`, and that inconsistency is also not great.


What I am leaning to is:
- If there is a barcode column, create a `cell_id_column` (which would not be in the provided retained columns vector since we are making it) as `{batch}-{barcode}`. Keep the barcode column as a default as you originally suggested (although don't hardcode the `barcode_column` into the args).
- If there is NO barcode column, create a `cell_id_column` as `{batch}-{rowname}`.
  - If there are no rownames, `stop()` 
  - Do not, however, create a barcode column since these will be `NA`s all around and that doesn't seem like it helps anybody.

sjspielman: Note some of this TBD code I pushed for different ways we can handle the barcode column..
https://github.com/AlexsLemonade/scpcaTools/blob/261dfe8458024a0d71e900f5fa3af24103fff459/R/merge_sce_list.R#L71-L85

https://github.com/AlexsLemonade/scpcaTools/blob/261dfe8458024a0d71e900f5fa3af24103fff459/R/merge_sce_list.R#L163-L170

I haven't added anything for `cell_id_column` yet.

jashapiro: > * Some SCEs have a `cell_id_column` scheme of `{batch}-{barcode}` but others may have scheme of `{batch}-{rowname}`, and that inconsistency is also not great.

I don't think this is necessarily a problem! The cell ids are pretty arbitrary, so one random string vs. another is not going to show up as a problem. But if we are worried about it, I would just can all the worry and use `{batch}-{rowname}`.

We would then remove everything around `barcode_column`, including the option. Treat `barcode` just like any other retained column, filling in `NA` values when missing (but make it one of the default retained columns, for our own ease of use and good example).

jashapiro: > * If there is a barcode column, create a `cell_id_column` (which would not be in the provided retained columns vector since we are making it) as `{batch}-{barcode}`. Keep the barcode column as a default as you originally suggested (although don't hardcode the `barcode_column` into the args).

If we decide to remove the `barcode_column` logic, or even if we don't, it might make sense to make a `cell_id` column and include it _always_. (Do I contradict myself, then I contradict myself!) My reason is that this will allow us to trace back to prior (unmerged) SCE objects should the need arise, without parsing the merged cell_id/colnames.

> * If there is NO barcode column, create a `cell_id_column` as `{batch}-{rowname}`.
>   
>   * If there are no rownames, `stop()`
 This should never occur, as rownames of some kind are always required by DataFrame.

sjspielman: Ok I'm going for it - 
- barcode is going to be a normal column, and I'll remove the argument. 
- cell_id will be made and will always be retained; whether they ask or not they get it. 100000% correct that we need something to trace back, and I prefer it being a cell_id as `{batch}-{rowname}`.




> This should never occur, as rownames of some kind are always required by DataFrame. If just numbers, we would still want to include those.

Good to know!

sjspielman:
Cleanups have landed! The roxygen comments have been heavily cleaned up, and `barcode_column` lives no more as its own entity. Tests are passing for me as well. Note that I put "barcode" back into the `retain_coldata_cols` default!

A few things I'm wondering:
- We now end up with a `cell_id_column`. However, it's the same content as the rownames now! I'm kind of on the fence for how I feel about this redundancy since I see good arguments either way.
- How to reformat the rowData rownames vs. colData rownames? Currently the former is cell-batch but the latter is batch-cell. Does anyone have strong preferences here regarding order and/or consistency?

jashapiro (commented):
I'm going to come back to this for a full final review tomorrow, but for now, just a couple comments, mostly in reply to your questions.

> A few things I'm wondering:
> 
> * We now end up with a `cell_id_column`. However, it's the same content as the rownames now! I'm kind of on the fence for how I feel about this redundancy since I see good arguments either way.

Ah, I was thinking that `cell_id` would be _only_ the previous colnames. My thought is that the `colnames(merged_sce)` after the merge are the unique identifiers for each cell across samples, but if you want to trace a cell back to an individual SCE, you need the `colnames(sce)`, which we are storing as `cell_id`.

> * How to reformat the rowData rownames vs. colData rownames? Currently the former is cell-batch but the latter is batch-cell. Does anyone have strong preferences here regarding order and/or consistency?

I think the the rownames is `featurestat-batch`, not cell, so the inconsistency doesn't bother me too much. However, `batch-featurestat` is also good and more consistent, so I'd probably switch to that.

R/merge_sce_list.R
```diff
@@ -0,0 +1,200 @@
...
+#'  Original SCE contents are modified or retained as follows:
+#'  - The resulting colData slot will include a new column specified by
+#'    `batch_column` (default "library_id") that either holds the originating SCE
+#'    object's name (referred to as `sce_name` here), of if it is unnamed then
+#'    its index in the provided `sce_list`.
+#'  - The resulting colData slot will include another new column `cell_id_column`
+#'    (default "cell_id") that is formatted as `{sce_name}-{rowname}`, where
+#'    `sce_name` is the given batch (library) value and `rowname` is the colData
+#'    row's rowname which often, but not always, holds a unique cell barcode.
+#'    This column serves to match cells back to both originating batch and cell name.
```
allyhawkins, line 17: I would agree with Josh's comment that this should just be the original barcode or `rownames(sce)` prior to merging.

R/merge_sce_list.R
```diff
@@ -0,0 +1,200 @@
...
+  if (length(sce_list) < 2) {
+    warning("There are fewer than two SCE objects in the provided `sce_list` so there is nothing to merge.")
+    # Early return:
+    return(sce_list)
+  }
+
+  # Check `retain_coldata_cols` ----------------
+  if (length(retain_coldata_cols) == 0) {
+    warning("All colData will be removed from the the merged SCE.
+     Please check that `retain_coldata_cols` was correctly specified.")
```
allyhawkins, line 80:
```suggestion
    warning("All pre-existing colData will be removed from the the merged SCE.
     Please check that `retain_coldata_cols` was correctly specified.")
```
Just specifying that we are removing the existing `colData` since we do add the `cell_id` and `batch_column`.

R/merge_sce_list.R
```diff
@@ -0,0 +1,200 @@
...
+
+}
+
+
+
+#' Helper function to prepare an SCE object for merging
+#'
+#' @param sce The SCE object to be prepared
+#' @param sce_name The name of the SCE object
+#' @param batch_column The name of the batch column will which be added to the
```
allyhawkins, line 133:
```suggestion
#' @param batch_column The name of the batch column which will be added to the
```

R/merge_sce_list.R
```diff
@@ -0,0 +1,200 @@
...
+#'   colData slot
+#' @param cell_id_column The name of the cell_id column will which be added to the
```
allyhawkins, line 135:
```suggestion
#' @param cell_id_column The name of the cell_id column which will be added to the
```

allyhawkins (commented):
Just a few minor comments, the biggest being that I agree with the comment that the `cell_id_column` should only have the barcode or original rowname and not contain the batch ID. 

Also this is failing the R CMD check on GHA because the `.Rd` documentation files are not up to date. It looks like there are still arguments that are documented in those files that are not present in the functions anymore.

R/merge_sce_list.R
```diff
@@ -0,0 +1,200 @@
...
+#'  For example, such a vector might contain items like "Gene", "ensembl_ids", etc.
+#'
+#' @return A SingleCellExperiment object containing all SingleCellExperiment objects
+#'   present in the inputted list
+#' @export
+#'
+#' @import SingleCellExperiment
+merge_sce_list <- function(sce_list = list(),
+                           batch_column = "library_id",
+                           cell_id_column = "cell_id",
```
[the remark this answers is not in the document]
sjspielman: My concern is if they, by chance, already have something named `cell_id` we'd overwrite their data. That said "partly because it seems to be the norm for SCE functions" is a solid argument.

I might pop it at the end of options, then.

R/merge_sce_list.R
```diff
@@ -0,0 +1,198 @@
...
+#'  are unique library IDs. Default value is `library_id`.
+#' @param retain_coldata_cols A vector of colData columns which should be retained
+#'  in the the final merged SCE object.
+#' @param preserve_rowdata_cols A vector of column names that appear in originating
+#'  SCE objects' rowData slots which should not be renamed with
+#'  the given SCE object name or index is name is not given. These are generally
+#'  columns which are not specific to the given library's preparation or statistics.
+#'  For example, such a vector might contain items like "Gene", "ensembl_ids", etc.
+#' @param cell_id_column A character value giving the resulting colData colum name
+#'  to hold unique cell IDs formatted as their batch and original rowname. Default
```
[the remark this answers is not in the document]
allyhawkins: It actually looks like you do that in the code but just the docs need to be updated.

allyhawkins (commented):
Just a few more small questions that I think are all docs updates.

jashapiro (approved):
LGTM! I had a few minor suggestions in the main code, and some simplifications in the tests, which I did test locally, so I expect they should still pass. 

Approving because I don't think any of the changes I suggested are important enough to require rereview.

R/merge_sce_list.R
```diff
@@ -0,0 +1,198 @@
...
+#'  Original SCE contents are modified or retained as follows:
+#'  - The resulting colData slot will include a new column specified by
+#'    `batch_column` (default "library_id") that either holds the originating SCE
+#'    object's name (referred to as `sce_name` here), or if it is unnamed then
+#'    its index in the provided `sce_list`.
+#'  - The resulting colData slot will include another new column `cell_id_column`
+#'    (default "cell_id") that will contain the SCE's column names (i.e. colData
+#'    rownames). Often, but not always, the rowname holds a unique cell barcode.
+#'    This column serves to match cells back to both originating batch and cell name.
+#'  - The resulting colData rownames will be updated to match `cell_id`.
```
allyhawkins, line 17: One more update to docs to reflect that `cell_id_column` is the orginal rownames.

allyhawkins (approved):
One more docs update then good to go!

tests/testthat/test-merge_sce_list.R
```diff
@@ -0,0 +1,235 @@
...
+test_that("merging SCEs with matching genes works as expected", {
+
+  # First a warning for this early-return scenario
+  expect_warning(merge_sce_list(sce_list = list("sce1" = sce1)))
+
+  # Works as expected:
+  merged_sce <- merge_sce_list(sce_list,
+                               batch_column = batch_column,
+                               # "detected" should get removed:
+                               retain_coldata_cols = c("sum", "detected"),
```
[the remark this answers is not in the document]
sjspielman: Womp, this got swapped around at some point, will fix.

PR #149, opened 2022-11-22
doc and docker update

jashapiro:
Update docs to test again


allyhawkins (approved):
🤞 this works

jashapiro:
I added a change to the dockerfile here to fix the problem we saw where the latest build didn't include the latest changes. I think it was using the cache, because the dockerfile itself pulls the main branch using `remotes`. Now we use the github API to check if the package has changed, and if it has, the cache gets busted. This will force the final docker layers (including the remotes pull) to rerun and should get us the latest version.

PR #151, opened 2022-11-23
Minor QC report fixes

allyhawkins:
We were seeing some issues with the QC report, specifically the warnings for the newly added plots were not printing correctly and the table that included the transcript type was double printing. This PR should address those issues. I generated a report without a processed SCE object to make sure the warnings were printing correctly and the table looked okay and attached it here. 
[SCPCL000050_qc_report.html.zip](https://github.com/AlexsLemonade/scpcaTools/files/10077877/SCPCL000050_qc_report.html.zip)


jashapiro (approved):
LGTM, and will let us test the dockerfile changes from my last pr!

PR #155, opened 2022-11-28
Account for potential NA in gene symbols for UMAPs of HVGs in QC report

allyhawkins:
In testing AlexsLemonade/scpca-nf#245 with the new release of `scpcaTools`, I noticed that for one of the libraries I was testing the UMAPs of the HVGs was showing one plot labeled with `NA` rather than a gene symbol. In going back to figure out why that was the case, it was because the ensembl ID for that particular HVG did not map to a gene symbol and we are using gene symbols to label the plots. I adjusted for this by using the ensembl ID for the label if a gene symbol isn't available. I updated the code and also added a note in the report about the labels.

### DIFF
--- a/inst/rmd/qc_report.rmd
+++ b/inst/rmd/qc_report.rmd
@@ -503,7 +503,8 @@ if(has_processed & has_var_genes){
     tibble::rownames_to_column("ensembl_id") |>
     select(ensembl_id, gene_symbol) |>
     filter(ensembl_id %in% top_genes) |>
-    mutate(ensembl_id = factor(ensembl_id, levels = top_genes)) |>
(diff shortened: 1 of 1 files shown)


jashapiro (commented):
This looks good, but is there a commit that didn't push? Specifically, you said:

>  I updated the code and also added a note in the report about the labels.

but I don't see the note!

allyhawkins:
> but I don't see the note!

I committed but didn't push, should be added now in https://github.com/AlexsLemonade/scpcaTools/pull/155/commits/82c9acacac628ae6875e9567ea9bd027e21b5778

jashapiro (approved):
LGTM!

PR #156, opened 2022-11-30
add miQC options to add_miQC

jashapiro:
Does what it says on the tin... A small change to our `add_miQC()` function to add in some options that we can pass to miQC filtering, with the same defaults as the current version of miQC. I considered passing them with `...` but that woudl have required a few more checks to be sure there were not duplicated options, so I went with this slightly more verbose but explicit version.


allyhawkins (approved):
LGTM 🚀

PR #157, opened 2022-12-01
Integration functions for fastmnn and harmony

### LINKED ISSUE #153: Function for integrating with fastMNN
For the data integration workflow in `scpca-nf` and ultimately the workflow that will also be present in downstream analyses, we will want to be able to integrate using `fastMNN`. We should create a central function here that we can use in both workflows. This function should take as input a merged SCE object with a column that indicates the batch or library ID that each cell belongs to. The function should then perform integration using `fastMNN` and return the corrected embeddings. The main decision here that we will have to decide is if we want to return the SCE object with the embeddings adding in with `method_PCA` or just return the embeddings to be added into an object outside of the function.

### LINKED ISSUE #154: Function for integrating with Harmony
For the data integration workflow in `scpca-nf `and ultimately the workflow that will also be present in downstream analyses, we will want to be able to integrate using `harmony`. We should create a central function here that we can use in both workflows. This function should take as input a merged SCE object with a column that indicates the batch or library ID that each cell belongs to. The function should then perform integration using `harmony` and return the corrected embeddings. The main decision here that we will have to decide is if we want to return the SCE object with the embeddings adding in with the name `harmony_PCA` or just return the embeddings to be added into an object outside of the function.

sjspielman:
Closes #153
Closes #154 

This PR adds functions to integrate with `fastMNN` and `harmony`. The reason I have these in the same PR is because I ended up writing an overall wrapper function, `integrate_sces()`, which calls either of these two functions. Plus the `fastMNN` function is a particularly light lift (no prep requires beyond what is required for both methods). All tests are passing locally, and `renv` is updated with `batchelor` and `harmony`.

Some discussion items - 

- I prefixed arguments `harmony_do_PCA` and `harmony_covariate_cols` with the method name, but for the argument `return_corrected_expression` (admittedly a long name, thoughts?) which only applies to `fastMNN` I did not prefix. I'm on the fence about this inconsistency as well as whether it is an inconsistency! 
- I used `retain_assays` rather making this argument, for example, `remove_assays`. Any thoughts?
- Note that `harmony_covariate_cols` can't be handled by the `...` because this info needs to be added into a metadata df.
- I "hardcode" the final PCA returned as `{integration_method}_PCA`, but this can also be a function argument. Similarly, I assume the PCs being used for integration are called “PCA”.
- I do not do _anything_ with gene subsetting. I assume that all genes in the inputted SCE should be used for integration, meaning it’s on users to subset this down earlier if they prefer.

**edit** _rather disappointingly_, github somehow deleted a final long paragraph in this PR comment! It was about my "what's the `renv.lock` deal here?" journey, with the main punchline of - should I not have run `snapshot()` locally from the M1 but should I have done it from inside the docker?
**edit 2** - the build is failing because of the lack of packages, but the number of times I ran `git add renv.lock` and it added `docker/renv.lock` instead exhausted my soul. No idea why.


jashapiro:
> **edit 2** - the build is failing because of the lack of packages, but the number of times I ran `git add renv.lock` and it added `docker/renv.lock` instead exhausted my soul. No idea why.

The test failure is because harmony and batchelor need to be in `DESCRIPTION`. to get them installed in the test environment (which does not use `renv`). I'd suggest putting them in `Suggests:`, and then adding some logic to check that they are installed before use... For an example: see https://github.com/AlexsLemonade/scpcaTools/blob/2c383cf74ce14dffba21da1d0c97a26158c80b6b/R/sce_to_seurat.R#L23-L25

sjspielman:
> The test failure is because harmony and batchelor need to be in `DESCRIPTION`

😑 wow, sad day for me. .... Of course!

jashapiro (commented):
Okay, I couldn't resist a quick start of a review. I didn't go past the main args, because my proposals will hopefully result in a fair amount of code removal. Basically, I think we can be much more opinionated about what the function does and does not do, as the original methods are always there to fall back on.

R/integrate_sces.R
```diff
@@ -0,0 +1,193 @@
...
+#' @param retain_assays A character vector of uncorrected assays to retain in
+#'   the integrated SCE object. By default, both "counts" and "logcounts" are retained.
+#' @param return_corrected_expression A boolean indicating whether corrected expression
+#'   values determined by the given integration method should be included in the
+#'   integrated SCE object. Note that `harmony` does not calculate corrected expression,
+#'   so this argument is ignored for this integration method. Default is `TRUE`.
+#' @param harmony_do_PCA A `harmony`-specific argument (ignored otherwise) indicating whether
+#'   integration should perform PCA. If `FALSE` (default), `harmony` will use the existing PCs to perform
+#'   integration. If `TRUE`, `harmony` will will start from normalized expression when integrating and
+#'   re-calculate PCs.
```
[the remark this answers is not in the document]
sjspielman: I do not think we will ever start with corrected expression, no, so this can be removed entirely. We can explain in docs accordingly.
This is what `harmony` will do if you ask it to calculate PCs, just fwiw - 

https://github.com/immunogenomics/harmony/blob/63ebd73199b79ae0d7563425a765d76deb916fe3/R/ui.R#L125-L129

R/integrate_sces.R
```diff
@@ -0,0 +1,193 @@
...
+#'
+#' @return An update combined SCE object containing
+#'
+#' @import SingleCellExperiment
+#'
+#' @export
+integrate_sces <- function(combined_sce,
+                           integration_method,
+                           batch_column = "sample",
+                           retain_assays = c("counts", "logcounts"),
```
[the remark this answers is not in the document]
sjspielman: > are there situations where we would want to remove assays?

R/integrate_sces.R
```diff
@@ -0,0 +1,193 @@
...
+#'   values determined by the given integration method should be included in the
+#'   integrated SCE object. Note that `harmony` does not calculate corrected expression,
+#'   so this argument is ignored for this integration method. Default is `TRUE`.
+#' @param harmony_do_PCA A `harmony`-specific argument (ignored otherwise) indicating whether
+#'   integration should perform PCA. If `FALSE` (default), `harmony` will use the existing PCs to perform
+#'   integration. If `TRUE`, `harmony` will will start from normalized expression when integrating and
+#'   re-calculate PCs.
+#' @param harmony_covariate_cols A `harmony`-specific argument (ignored otherwise)
+#'   providing other columns in the combined SCE besides `batch_column` that `harmony`
+#'   should consider as covariates during integration.
```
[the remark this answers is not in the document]
sjspielman: I'll implement this and we can see what we think! 
On one hand, I do think the `batch_column` is nice to have for consistency with `merge_sce_list()` and it's standard-ish to supply the batch column as its own entity in integration methods. But, it's true these arguments are somewhat redundant.

R/integrate_sces.R
```diff
@@ -0,0 +1,159 @@
...
+                           seed = NULL,
+                           ...) {
+
+  # Set seed
+  set.seed(seed)
+
+  # make sure that input is a SingleCellExperiment
+  if(!is(combined_sce, "SingleCellExperiment")){
+    stop("The `combined_sce` must be a SingleCellExperiment object created with `scpcaTools::merge_sce_list().`")
+  }
```
allyhawkins, line 44: I think we may want two separate checks here. One check that this is a `SingleCellExperiment` object and then another check for the number of unique batches/samples that are provided. If there is only 1 sample in the SCE object then the error message you have here would be appropriate. 

Because of this, I think we do want a `batch_column` argument... I kind of like having a separate `covariate_cols` and `batch_column` argument because the `batch_column` is always required while the `covariate_cols` specifically refers to any extra covariates and is not a requirement.

tests/testthat/test-integrate_sces.R
```diff
@@ -0,0 +1,211 @@
...
+
+
+test_that("`integrate_fastmnn` fails as expected when logcounts is missing", {
+  logcounts(combined_sce) <- NULL
+  expect_error(integrate_fastmnn(combined_sce,
+                                 batch_column))
+})
+
+
+test_that("`integrate_harmony` works as expected when do_PCA is TRUE", {
```
allyhawkins, line 67:
```suggestion
test_that("`integrate_harmony` works as expected", {
```

allyhawkins (commented):
Thanks for tackling this! Overall I think this looks good with the changes that Josh suggested. The only major thing I am still unsure about is combining `batch_column` and `covariate_columns` into one argument as I think we should still keep them separate. It would allow you to check for multiple batches being present in the SCE object prior to integration.

R/integrate_sces.R
```diff
@@ -0,0 +1,180 @@
...
+#'   providing other columns in the combined SCE besides `batch_column` that `harmony`
+#'   should consider as covariates during integration.
+#' @param return_corrected_expression A boolean indicating whether corrected expression
+#'   values determined by the given integration method should be included in the
+#'   integrated SCE object. Note that `harmony` does not calculate corrected expression,
+#'   so this argument is ignored for this integration method. Default is `TRUE`.
+#' @param seed Random seed to set for integration. This is only set if the value is not `NULL`.
+#' @param ... Any additional parameters to be passed to the given integration method
+#'
+#' @return An update merged SCE object containing a new reduced dimension named
```
allyhawkins, line 24:
```suggestion
#' @return An updated merged SCE object containing a new reduced dimension named
```

R/integrate_sces.R
```diff
@@ -0,0 +1,180 @@
...
+
+
+
+
+#' Integrate a list of SingleCellExperiment (SCE) objects using the harmony
+#' approach from the `harmony` package.
+#' Source: https://cran.r-project.org/web/packages/harmony/index.html
+#'
+#' @param merged_sce A combined SCE object as prepared by `scpcaTools::merge_sce_list()`.
+#' @param batch_column The column in the combined SCE object indicating batches
```
allyhawkins, line 137:
```suggestion
#' @param batch_column The column in the merged SCE object indicating batches
```

R/integrate_sces.R
```diff
@@ -0,0 +1,180 @@
...
+  return(merged_sce)
+}
+
+
+
+#' Integrate a list of SingleCellExperiment (SCE) objects using the fastMNN
+#' approach from the `batchelor` package.
+#' Source: http://www.bioconductor.org/packages/release/bioc/html/batchelor.html
+#'
+#' @param merged_sce A combined SCE object as prepared by `scpcaTools::merge_sce_list()`.
```
allyhawkins, line 105:
```suggestion
#' @param merged_sce A merged SCE object as prepared by `scpcaTools::merge_sce_list()`.
```

R/integrate_sces.R
```diff
@@ -0,0 +1,180 @@
...
+}
+
+
+
+
+#' Integrate a list of SingleCellExperiment (SCE) objects using the harmony
+#' approach from the `harmony` package.
+#' Source: https://cran.r-project.org/web/packages/harmony/index.html
+#'
+#' @param merged_sce A combined SCE object as prepared by `scpcaTools::merge_sce_list()`.
```
allyhawkins, line 136:
```suggestion
#' @param merged_sce A merged SCE object as prepared by `scpcaTools::merge_sce_list()`.
```

allyhawkins (approved):
Just some minor typos, but other than that this look good to me!

R/integrate_sces.R
jashapiro: This is fine. I think I might make the additional argument just `covariate_columns` though, in case we add a future method that also allows covariates, and throw a warning if the method chosen does not support covariates. (I don't think that needs to be a fatal error).

R/integrate_sces.R
```diff
@@ -0,0 +1,180 @@
...
+                                        batch_column,
+                                        ...)
+    # Add corrected expression to merged_sce if specified
+    if (return_corrected_expression) {
+      assay(merged_sce, "fastMNN_corrected") <- assay(integrated_sce, "reconstructed")
+    }
+    # define PCs
+    integrated_pcs <- reducedDim(integrated_sce, "corrected")
+  }
+  if (integration_method == "harmony") {
```
jashapiro, line 83: Minor, but since both are not allowed (at this point...)
```suggestion
  else if (integration_method == "harmony") {
```

R/integrate_sces.R
```diff
@@ -0,0 +1,180 @@
...
+  # Ensure >=2 batches
+  if( length(unique(batches)) < 2 ) {
+    stop("At least two batches are required for integration.")
+  }
+
+
+  # Perform integration with specified method
+  if (integration_method == "fastMNN") {
+
+    if (length(covariate_cols) >= 1) {
```
jashapiro, line 69: I don't know why, but...
```suggestion
    if (length(covariate_cols) > 0) {
```

I mean, we could even do:
```
if (length(covariate_cols)) {
```
which would be more pythonic, but this is R. (Really Pythonic would actually be `if(covariate_cols)`, but that won't work in R)

R/integrate_sces.R
```diff
@@ -0,0 +1,180 @@
...
+#'   integration, the SCE also includes a corrected expression assay (`fastMNN_corrected`)
+#'
+#' @import SingleCellExperiment
+#'
+#' @export
+integrate_sces <- function(merged_sce,
+                           integration_method = c("fastMNN", "harmony"),
+                           batch_column = "sample",
+                           covariate_cols = c(),
+                           return_corrected_expression = TRUE,
```
jashapiro, line 35: Wondering if this is the default we want. I think `FALSE` maybe makes more sense (not sure how often we will use this), and then add a warning for `harmony` if true since this is not an option there.

R/integrate_sces.R
```diff
@@ -0,0 +1,180 @@
...
+  # Check integration_method
+  integration_method <- match.arg(integration_method)
+
+  # Check batch_column
+  if (!(batch_column %in% names(colData(merged_sce)))) {
+    stop("The provided `batch_column` column must be in the `merged_sce` colData.")
+  }
+  # Ensure there are no NAs
+  batches <- colData(merged_sce)[[batch_column]]
+  if( sum(is.na(batches)) > 0 ) {
```
jashapiro, line 57:
```suggestion
  if( any(is.na(batches)) ) {
```

R/integrate_sces.R
```diff
@@ -0,0 +1,180 @@
...
+#'
+#' @import SingleCellExperiment
+integrate_fastmnn <- function(merged_sce,
+                              batch_column,
+                              ...) {
+
+  # Check the merged_sce for logcounts
+  if (!("logcounts" %in% names(assays(merged_sce)))) {
+    stop("The `merged_sce` object requires a `logcounts` assay for fastMNN integration.")
+  }
```
jashapiro, line 119: We need to check that batchelor is installed, since it is not a requirement for scpcaTools in general. Similarly, we will need a check for `harmony`
```suggestion
  # Check that batchelor is installed
  if (!requireNamespace("batchelor", quietly = TRUE)) { 
    stop("The batchelor package must be installed to use fastMNN.") 
  }

  # Check the merged_sce for logcounts
  if (!("logcounts" %in% names(assays(merged_sce)))) {
    stop("The `merged_sce` object requires a `logcounts` assay for fastMNN integration.")
  }
```

R/integrate_sces.R
```diff
@@ -0,0 +1,180 @@
...
+  # Check covariate columns
+  if (!(all(covariate_cols %in% names(colData(merged_sce))))) {
+    stop("The provided covariate columns are not all present in the `merged_sce` colData.")
+  }
+
+  # Setup harmony metadata
+  covariate_cols <- c(batch_column, covariate_cols)
+  harmony_metadata <- tibble::as_tibble(colData(merged_sce)) %>%
+    dplyr::select(
+      dplyr::all_of(c(batch_column, covariate_cols))
```
jashapiro, line 165: I think this is redundant since batch_column was already added?
```suggestion
      dplyr::all_of(covariate_cols)
```

tests/testthat/test-integrate_sces.R
```diff
@@ -0,0 +1,214 @@
+# generate testing data
+set.seed(1665)
+merged_sce <- sim_sce(n_cells = 300, n_genes = 100, n_empty = 0)
+batches <- c(rep("a", 100),
+             rep("b", 100),
+             rep("c", 100))
```
jashapiro, line 6: So very minor, but here we are...
```suggestion
batches <- rep(c("a", "b", "c"), each = 100)
```

tests/testthat/test-integrate_sces.R
```diff
@@ -0,0 +1,214 @@
...
+barcodes <- rownames(colData(merged_sce))
+
+# set up colData rownames as batch-barcode
+new_rownames <- tibble::tibble(
+  batches = batches,
+  barcodes = barcodes,
+  ids = glue::glue("{batches}-{barcodes}")) %>%
+  dplyr::pull(ids)
```
jashapiro, line 14: No need to make the tibble if you aren't saving it
```suggestion
new_rownames <- glue::glue("{batches}-{barcodes}")) 
```

tests/testthat/test-integrate_sces.R
```diff
@@ -0,0 +1,214 @@
...
+rownames(colData(merged_sce)) <- new_rownames
+
+# Add some info to colData:
+# Add "sample" to colData
+colData(merged_sce)$sample <- batches
+# Add in a "covariate" column for testing
+colData(merged_sce)$covariate <- sample(c("patient-a", "patient-b", "patient-c", "patient-d"),
```
jashapiro, line 21: This is not the way I would expect a covariate to work in real data, so it threw me for a second, but I don't think it matters for the test. But maybe not naming it `patient`.

tests/testthat/test-integrate_sces.R
```diff
@@ -0,0 +1,214 @@
...
+  expect_error(
+    integrate_harmony(merged_sce, batch_column)
+  )
+})
+
+test_that("`integrate_harmony` fails when covariate columns are missing", {
+  expect_error(
+    integrate_harmony(merged_sce,
+                      batch_column,
+                      harmony_covariate_cols = "not_a_column"
```
jashapiro, line 90: Noting this will need an update to test what you want it to test... It _does_ fail here, but because an invalid argument is passed to `harmony`.
```suggestion
                      covariate_cols = "not_a_column"
```

jashapiro (commented):
This looks good to me! I have a number of minor comments. Some _very_ minor, forgive me.

The most significant one is adding the checks for `batchelor` and `harmony`, as those are not required by the package. (I suggested one, left the other as an exercise for the submitter).

My other really substantive comment/question is about the "best" default for `return_corrected_expression`.

tests/testthat/test-integrate_sces.R
sjspielman: > but I don't think it matters for the test.

I was all in on this! I can remove `patient`, it's not needed.

jashapiro (approved):
LGTM! (As long as tests pass, which I expect them to given the previous pass and the last commit's changes)

PR #159, opened 2023-01-04
Make UMAP plotting optional

### LINKED ISSUE #158: Account for missing UMAPs in QC Report
We will need to modify the QC report to skip printing out the UMAPs, even in the presence of a `processed` SCE object. In some cases, we may have poor quality libraries and a UMAP may be unable to be calculated as was identified in https://github.com/AlexsLemonade/scpca-nf/issues/260. Right now the QC report always attempts to print a UMAP as long as a processed object exists. 

One approach to this would be to check if the UMAP exists in the object and if it doesn't exist don't include the plots. We previously had been calculating UMAP results if they don't exist, but this did not account for if the UMAP calculation would fail. 

As mentioned in https://github.com/AlexsLemonade/scpca-nf/issues/260#issuecomment-1353820122, we probably also want to include a warning at the top of the QC report if the UMAP is missing and also include the UMAP plots as a separate child report, similar to how we deal with the addition of CITE-seq information.

allyhawkins:
Closes #158 

This PR splits out any of the UMAP plots into their own child report that is only included if the `processed_sce` object that is provided includes UMAP results. I took everything from the `Dimensionality reduction` heading and moved into its own notebook that gets loaded in if UMAP results are present. 

Originally we were checking for if the processed SCE object had UMAP results and if they were not present, we were adding them in. Here I changed to just check for the UMAP being present and if it's not no plots are being produced and we aren't attempting to recalculate it. This makes it a little simpler because we also had some checks for PCA and using variable genes as well. I was torn on if this was the right approach, but having a nice warning that no UMAP plots are present rather than attempting to re-calculate and have it fail again seemed to make the most sense to me. 

I also added two warning messages, one for a low number of cells in the filtered SCE object (n < 100) and one for a low number of cells in the processed SCE object (n < 50). I chose the 50 cutoff because that's when the UMAP will fail and the 100 cutoff because that's when we have issues with normalization, but I'm open to changing them or adding other warnings for different cutoffs if we think those are necessary. 

I tested these changes with a library that didn't have any issues: 
[SCPCL000050_qc_report.html.zip](https://github.com/AlexsLemonade/scpcaTools/files/10348015/SCPCL000050_qc_report.html.zip)

And then the one that failed UMAP: 
[SCPCL000784_qc_report.html.zip](https://github.com/AlexsLemonade/scpcaTools/files/10348016/SCPCL000784_qc_report.html.zip)

### DIFF
--- a/inst/rmd/qc_report.rmd
+++ b/inst/rmd/qc_report.rmd
@@ -88,7 +88,46 @@ has_cellhash <- "cellhash" %in% altExpNames(filtered_sce)
 if(has_cellhash){
   modalities <- c(modalities, "Multiplex")
 }
+
+# check for umap, need to be sure that processed_sce exists first
+if(has_processed){
+  has_umap <- "UMAP" %in% reducedDimNames(processed_sce)
+} else {
+  has_umap <- FALSE
+}
 ```
+
+```{r, results='asis'}
+# check for number of filtered cells
+if(has_filtered){
+  if(ncol(filtered_sce) > 100){
+    glue::glue("
+      <div class=\"alert alert-warning\">
+      
+      This library contains fewer than 100 cells post removal of empty droplets.
+      This may affect the interpretation of results.
+      
+      </div>
+    ")
+  } 
+}
+
+# check for number of cells post processing
+if(has_processed){
+  if(ncol(processed_sce) > 50){
+    glue::glue("
+      <div class=\"alert alert-warning\">
+      
+      This library contains fewer than 50 cells post removal of low quality cells.
+      UMAP is unable to be calculated and plots will not be shown.
+      
+      
+      </div>
+    ")
+  }  
+}
+```
+
 
 # Processing Information for `r library_id`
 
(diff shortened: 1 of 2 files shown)


inst/rmd/umap_qc.rmd
```diff
@@ -0,0 +1,100 @@
...
+                 point_alpha = 0.5,
+                 colour_by = "detected") +
+  scale_color_viridis_c()+
+  theme_bw() + 
+  guides(
+    color = guide_colorbar(title = "Number of \ngenes detected")
+  )
+```
+
+Compromised cells, calculated using `miQC` described above, are filtered and removed from the library prior to downstream analyses.
```
jashapiro, line 16: Should this sentence (and the one after it) appear somewhere else in the report? I feel like we want to include these statements whenever the processed data exists, whether or not the UMAP was calculated.

inst/rmd/umap_qc.rmd
```diff
@@ -0,0 +1,100 @@
...
+  scale_color_viridis_c()+
+  theme_bw() + 
+  guides(
+    color = guide_colorbar(title = "Number of \ngenes detected")
+  )
+```
+
+Compromised cells, calculated using `miQC` described above, are filtered and removed from the library prior to downstream analyses. 
+The raw counts from all remaining cells are then normalized prior to selection of highly variable genes and dimensionality reduction. 
+The above plot shows the UMAP (Uniform Manifold Approximation and Projection) embeddings for each cell, coloring each cell by the total number of genes detected per cell.
```
jashapiro, line 18: I'm wondering (not for this PR, probably) if we want to use the official caption support in Rmarkdown to make these into "proper" captions. 

That would look something like this:
````
```{r message=FALSE}
#| fig.cap: UMAP plot colored by the total number of genes detected per cell.

scater::plotUMAP(processed_sce,
                 ...
````

The advantage is it would save us from referring to plots above and below.

File: inst/rmd/umap_qc.rmd [hunk 3]
```diff
@@ -0,0 +1,100 @@
...
+The plots below show the same UMAP embeddings, coloring each cell by the expression level of the labeled gene.
```
jashapiro, line 19: Should we make this a separate subsection? I think it is worth calling out as "Expression patterns of variable genes" or something like that?

jashapiro (commented):
This looks good. I had a couple of suggestions, one about making the cutoffs easier to change, and the other about moving a couple of sentences about the processing out of the UMAP/dimensionality reduction subdocument. I think we want to say those things whether or not we actually have a UMAP plot to share. 

I also had a future thought about captions, but we can revisit that.

allyhawkins:
I went ahead and moved the beginning few sentences about filtering and normalization to the main report after the filtering plot. I also added a sub heading for the expression of the HVG plots. As far as captions, I think that would be nice to have, but I played around with this for a little bit but it would involve more changes than I initially thought, so probably better for a different PR and more of a "nice to have" feature.

inst/rmd/qc_report.rmd
```diff
@@ -431,136 +470,13 @@ if(has_filtered & has_processed){
...
-    )
-  
-} else {
-  # if processed is missing, set to null to skip next UMAPs
-  has_var_genes <- FALSE
-}
-```
-
 Compromised cells, calculated using `miQC` described above, are filtered and removed from the library prior to downstream analyses. 
 The raw counts from all remaining cells are then normalized prior to selection of highly variable genes and dimensionality reduction.
```
jashapiro, line 474: I started with editing these sentences to refer to the fact that miQC is not always done, but I went back to looking at it as a whole, and I think we can probably just drop these two sentences? Back on line 426, we say "The below plot highlights cells that were removed prior to normalization and dimensionality reduction." So maybe that covers it?


```suggestion
The raw counts from all cells that remain after filtering low quality cells are then normalized prior to selection of highly variable genes and dimensionality reduction. 
```

inst/rmd/umap_qc.rmd
```diff
@@ -0,0 +1,101 @@
...
+    color = guide_colorbar(title = "Number of \ngenes detected")
+  )
+```
+
+### Expression of highly variable genes
+
+The plots below show the same UMAP embeddings, coloring each cell by the expression level of the labeled gene.
+The genes chosen for plotting are the 12 most variable genes identified in the library.
+Gene symbols are used when available to label the UMAP plots.
+If gene symbols are not available, the Ensembl ID will be shown.
```
jashapiro, line 23: Noting here that we are inconsistent about the use of `id` vs `ID`; we should probably clean that up. (We use "id" in sample metrics table, for example). I don't really care which we use, but we should try to be consistent. 

I only noticed this because it came up in a training document, and then I spent too long looking to figure out what is  "correct" and the answer is kind of neither, but probably "ID" is more common, mostly because people transferred ID as an initialism for "identity document" to an abbreviation for "identifier", even though they are not the same. But the training docs (mostly) use "id", because that seems to be what we have been doing...

jashapiro (approved):
LGTM! I just had one thought about whether we even need the sentences that I had you move (filed as a separate comment) and then an extended rabbit hole comment about "ID" vs "id"

PR #163, opened 2023-02-10
Add argument for highly variable genes to integration function

allyhawkins:
In working on adding the steps for integration to the workflow in `scpca-nf`, I noticed that we don't explicitly have an option to indicate using highly variable genes for integration with `fastMNN`. Considering that we pretty much exclusively are running `fastMNN` with only highly variable genes, I think it makes sense to add that in. We do pass additional arguments through using `...`, but the problem is here we are using one function for integration that will either perform `fastMNN` or `harmony`, so any options that are being passed to the function that are not compatible with one of the methods will cause an error. 

This is mostly problematic when writing the script to perform both types of integration in `scpca-nf`, because we only want to provide the highly variable genes with `fastMNN` and not for `harmony`. One way around this would be to call the function differently depending on the integration method being used as input, but that seemed like overkill for one argument that is different. We also have already accounted for dealing with harmony covariates and ignoring them for `fastMNN`, so I decided to take the same approach with the highly variable genes. I set them to be `NULL` by default, but if they are used they will only be passed through to `fastMNN`, otherwise they are ignored and a warning is printed. 

Another thought I had when working on this was getting rid of the wrapper function entirely... so that all of the things that happen in the wrapper function to distinguish `harmony` vs. `fastMNN` would happen in `scpca-nf` instead, but given we are also planning to use integration in other places (like downstream analyses), I talked myself out of that.

### DIFF
--- a/R/integrate_sces.R
+++ b/R/integrate_sces.R
@@ -11,6 +11,9 @@
 #' @param integration_method The integration method to use. One of `fastMNN` or
 #'  `harmony` (case-sensitive)
 #' @param batch_column The column in the merged SCE object indicating batches
+#' @param hv_genes A vector specifying which highly variable features to use for correction.
+#'   This argument will only be used if integrating with `fastMNN` and ignored with `harmony`.
+#'   Default is `NULL`.
 #' @param covariate_cols A vector of additional columns in the merged SCE to
 #'   consider as covariates during integration. Currently, this is used only by
 #'   `harmony`.
@@ -31,6 +34,7 @@
 integrate_sces <- function(merged_sce,
                            integration_method = c("fastMNN", "harmony"),
                            batch_column = "sample",
+                           hv_genes = NULL,
                            covariate_cols = c(),
                            return_corrected_expression = FALSE,
                            seed = NULL,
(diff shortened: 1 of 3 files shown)


jashapiro:
> Another thought I had when working on this was getting rid of the wrapper function entirely... so that all of the things that happen in the wrapper function to distinguish `harmony` vs. `fastMNN` would happen in `scpca-nf` instead, but given we are also planning to use integration in other places (like downstream analyses), I talked myself out of that.

I think the value of the wrapper function is that you always get an SCE out, so keeping it seems reasonable to me.

R/integrate_sces.R
```diff
@@ -131,6 +142,7 @@ integrate_fastmnn <- function(merged_sce,
   # Perform integration
   integrated_sce <- batchelor::fastMNN(merged_sce,
                                        batch = colData(merged_sce)[,batch_column],
+                                       subset.row = hv_genes,
```
jashapiro, line 145: Another question I have is whether we want to use `auto.merge` here?

allyhawkins: We've never done anything with that option, so I think we would want to test it out before using it automatically, but we could add it in as an option so we can more easily use it?

allyhawkins:
> I had a thought that there might be other reasons to call the function differently in the script... if we want to add parallel processing, which I think we do, then we have to do that specifically for fastMNN, but I don't think harmony supports it.

I also was thinking about this and I think if we want to do that then we probably want to break up how we call it in `scpca-nf`.

R/integrate_sces.R
jashapiro: I think we will want to test it, but we can do that in the script. I think we do want to use it, or some method of deciding merge order other than just the order we got things...

PR #164, opened 2023-02-14
Modify generate qc report to include arguments for optional rmd file and extra parameters

### LINKED ISSUE #162: Modify `generate_qc_report` to take a template Rmd file as input
Currently the `generate_qc_report` function does not allow you to alter the Rmd file that is rendered and will always use the `rmd/qc_report.rmd` file stored in the package. We would like to be able to alter this so that we can use the same function to generate other qc or summary reports using any input template Rmd file. The path to the Rmd file will need to be added as an argument to the function and we will also need to figure out how to deal with rendering reports that may have different parameters.

allyhawkins:
Closes #162 

Here I added two arguments to the `generate_qc_report` function, one for passing an `rmd` file to be rendered other than the default qc report that is included in the package, and another parameter for passing extra parameters to that qc report. 

In doing this I also made the decision that extra parameters would only be used if a different rmd file was being used, because I think if we are going to make any changes to the qc report in the package already we would probably add those as optional arguments to the function already. 

I also added some additional tests to make sure that things were failing if the input rmd file didn't exist and that the warnings were being picked up correctly. Having both tests may be overkill, so I'm fine to take them out if others think we don't need them both in.


R/generate_qc_report.R
```diff
@@ -23,6 +27,8 @@ generate_qc_report <- function(library_id,
                                unfiltered_sce,
                                filtered_sce = NULL,
                                processed_sce = NULL,
+                               rmd_file = NULL,
+                               extra_params = c(),
```
jashapiro, line 31: Since you use an `is.null` test later, we should set the default to `NULL`
```suggestion
                               extra_params = NULL,
```

R/generate_qc_report.R
```diff
@@ -66,18 +72,40 @@ generate_qc_report <- function(library_id,
...
+  )
+
+  # define rmd file if none provided
+  if(is.null(rmd_file)){
+    rmd_file <- system.file(file.path("rmd", "qc_report.rmd"), package = "scpcaTools")
+
+    # ignore extra parameters if they exist
+    if(!is.null(extra_params)){
+      warning("`extra_params` will only be used if providing an input rmd file,
+              otherwise these will be ignored.")
```
jashapiro, line 90: Change to a message, as I don't think it rises to a warning, and moving to a single line to prevent odd formatting of line breaks.
```suggestion
      message("`extra_params` are only used if providing a template file, so will be ignored")
```

R/generate_qc_report.R
```diff
@@ -23,6 +27,8 @@ generate_qc_report <- function(library_id,
                                unfiltered_sce,
                                filtered_sce = NULL,
                                processed_sce = NULL,
+                               rmd_file = NULL,
```
jashapiro, line 30: I'd suggest renaming with `template` in the name for clarity. Maybe not exactly this name, but something to say it is a template file.
```suggestion
                               report_template = NULL,
```

R/generate_qc_report.R
```diff
@@ -66,18 +72,40 @@ generate_qc_report <- function(library_id,
...
+    # ignore extra parameters if they exist
+    if(!is.null(extra_params)){
+      warning("`extra_params` will only be used if providing an input rmd file,
+              otherwise these will be ignored.")
+    }
+
+  } else {
+    # rmd file is provided, check that it exists
+    if(!file.exists(rmd_file)){
+      stop("provided rmd file does not exist.")
```
jashapiro, line 96:
```suggestion
      stop("Provided template file does not exist.")
```

tests/testthat/test-generate_qc_report.R
```diff
@@ -5,8 +5,23 @@ filt_sce <- filter_counts(sce)
...
   qc_file <- generate_qc_report(library_id = "TEST",
                                 unfiltered_sce = sce,
-                                filtered_sce = filt_sce, )
+                                filtered_sce = filt_sce)
   expect_true(file.exists(qc_file))
   # clean up
   file.remove(qc_file)
 })
+
+test_that("generating a qc report does not work", {
```
jashapiro, line 14:
```suggestion
test_that("generating a qc report fails with a missing template", {
```

tests/testthat/test-generate_qc_report.R
```diff
@@ -5,8 +5,23 @@ filt_sce <- filter_counts(sce)
...
+  expect_error(generate_qc_report(library_id = "TEST",
+                                unfiltered_sce = sce,
+                                filtered_sce = filt_sce,
+                                rmd_file = "test.rmd"))
```
jashapiro, line 18: Just for a bit of extra clarity?
```suggestion
                                rmd_file = "missing.rmd"))
```

tests/testthat/test-generate_qc_report.R
```diff
@@ -5,8 +5,23 @@ filt_sce <- filter_counts(sce)
...
+})
+
+
+test_that("generating a qc report works with a warning", {
+  expect_warning(generate_qc_report(library_id = "TEST",
+                                    unfiltered_sce = sce,
+                                    filtered_sce = filt_sce,
+                                    extra_params = list(param = "test_param")))
+})
```
jashapiro, line 27: Since I downgraded to a message, I'd probably remove this...

allyhawkins:
Just a note that the failure appears to be a memory issue with trying to update `conda`. In looking at the [error](https://github.com/AlexsLemonade/scpcaTools/actions/runs/4176147487/jobs/7232046356) there's an exit because the conda updates are getting killed. I believe that's because of a memory issue... but I don't know for sure. Just wanted to point that out because based on my googling I think we need to change what we run the GHA on?

tests/testthat/test-generate_qc_report.R
jashapiro: I think we probably also want a tests that _does_ render a report with a template. That test template can probably be extremely short, but it should exist...

jashapiro (commented):
This looks good, but I had a few little suggestions for renaming the argument and maybe swapping out one test for another.

jashapiro:
> Just a note that the failure appears to be a memory issue with trying to update `conda`. In looking at the [error](https://github.com/AlexsLemonade/scpcaTools/actions/runs/4176147487/jobs/7232046356) there's an exit because the conda updates are getting killed. I believe that's because of a memory issue... but I don't know for sure. Just wanted to point that out because based on my googling I think we need to change what we run the GHA on?

I don't think it is a memory issue, but it is hopefully intermittent (I think I have seen failure at that point before). It probably makes sense to update [the action](https://github.com/AlexsLemonade/scpcaTools/blob/main/.github/workflows/R-CMD-check.yaml) to run on `ubuntu-22.04` which might update the version of conda that is used, which might help. At least that is a hope.

tests/testthat/test-generate_qc_report.R
```diff
@@ -5,8 +5,31 @@ filt_sce <- filter_counts(sce)
...
-                                filtered_sce = filt_sce, )
+                                filtered_sce = filt_sce)
   expect_true(file.exists(qc_file))
   # clean up
   file.remove(qc_file)
 })
+
+test_that("generating a qc report works with a different template", {
+
+  test_template <- system.file(file.path("inst", "rmd", "test_report.rmd"), package = "scpcaTools")
```
jashapiro, line 16: I think you actually want to put this file (since it is only for testing) in _this_ directory (`tests/testthat`) and then get it with just:
```suggestion
  test_template <-  "test_report.rmd"
```

allyhawkins:
I think something is killing the workflow. I'm consistently seeing this at the `Running ‘testthat.R’` step. 
`Error: Process completed with exit code 143.`

jashapiro:
I don't think this is related, but when I tried to run tests locally, I found that I could not restore the `renv` due to a package (`spatstats.core`) that seems to have been removed from CRAN & was a dependency of `Seurat`. Installing Seurat separately did the trick, even though that kind of sucks.

But the tests should be using the latest versions, so I'm not quite sure what might be going on with the error...

jashapiro:
My best guess is that it is taking too long to build the dependencies, so github is killing the job... I'm trying to think about if there is a way to speed things up. One might be to do a run where it builds but does not test (comment out the test step) which should allow the dependency installation step to cache its progress.

allyhawkins:
> I don't think this is related, but when I tried to run tests locally, I found that I could not restore the renv due to a package (spatstats.core) that seems to have been removed from CRAN & was a dependency of Seurat. Installing Seurat separately did the trick, even though that kind of sucks.

That is annoying... we might want to document that somewhere. Thanks for catching that! My tests have been passing locally but probably because I didn't need to run the restore step. 

> My best guess is that it is taking too long to build the dependencies, so github is killing the job... I'm trying to think about if there is a way to speed things up. One might be to do a run where it builds but does not test (comment out the test step) which should allow the dependency installation step to cache its progress.

It's worth a shot to at least see if it works.

jashapiro:
> > I don't think this is related, but when I tried to run tests locally, I found that I could not restore the renv due to a package (spatstats.core) that seems to have been removed from CRAN & was a dependency of Seurat. Installing Seurat separately did the trick, even though that kind of sucks.
> 
> That is annoying... we might want to document that somewhere. Thanks for catching that! My tests have been passing locally but probably because I didn't need to run the restore step. 
> 

Just to clarify, I think the correct solution is to update Seurat in the renv.lock... I am now out of sync locally.

allyhawkins:
> Just to clarify, I think the correct solution is to update Seurat in the renv.lock... I am now out of sync locally.

Ah okay, that should be done now!

jashapiro (approved):
LGTM! I expect the job to pass at some point soon (I just retriggered it). The jobs seem to be killed from somewhere outside the job itself (note they get little octogons not a red X), which seems like something unrelated to the code changes here...

PR #165, opened 2023-02-24
Add SingleR to R project and docker

allyhawkins:
This adds SingleR to the lockfile and to the dependencies list for the docker image so that we can use it in `scpca-nf`.


jashapiro (approved):
LGTM!

Issue #167, opened 2023-04-14
Add scverse to the scpcaTools docker image


jashapiro:
A secondary thought: we should probably make `scverse` packages part of the `scpcaTools` docker image: that seems like it may be easiest way to manage that installation for use in workflows, especially if we want to use `reticulate` to handle interoperation between R and python.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/sc-data-integration/pull/215#pullrequestreview-1385917084_

We may want to define the packages to install in a separate file that can be used for local installation as well (e.g. a conda `environment.yaml` file), but that may be more trouble than it is worth.

jashapiro:
completed by #169 (at least in initial form)

PR #169, opened 2023-05-01
Test adding python to renv

jashapiro:
⚠️ Stacked on #171 
In this draft PR, I am testing including python requirements as part of the `renv` setup.  (related to #167)

I did this by creating a local python virtual environment and installing some scverse tools to it, then using `renv` to handle the creation of the `requirements.txt` file.

I created the environment with the command `python3 -m venv .venv` in the base directory of the project, which created the `.venv` directory and filled it with some basic python packages. 
I then ran `source .venv/bin/activate` to activate the environment, followed by `pip install scanpy scvi-tools` to install the packages and their dependencies.

Then, from within R, I ran `renv::use_python(".venv/bin/python")` followed by `renv::snapshot()` to record the python environment into a `requirements.txt` file.

For the docker image, I then moved the `requirements.txt` file into the `docker` directory and replaced it in the base directory with an alias. I then added a step to the Dockerfile to use that requirements file to install the packages from that file.

So far, everything seems to work at my end, but I haven't done extensive testing. It does seem that running `renv::restore()` will automatically create the virtual environment if it doesn't exist, which is good (if it works for others).  (I will note that in RStudio, the snapshot and restore options in the menus seem to skip the python portion, so you will want to run `renv::` commands in the console)

I also added a multiplatform build option to the github action, because we might as well test that; it should make local testing on M1 machines faster/easier: I found that I couldn't install some of the python packages in an emulated environment due to limitations of the emulator, so I assume they won't run either! Luckily rocker now supports AMD images; hopefully it works!)

### Instructions to reviewers

I don't think this is a high priority project, but I was thinking about it, so I wanted to get a draft with some setup in. Note that I also updated some other dependencies/R version. But I did hope to see if this worked for anybody else, and what other problems crop up


allyhawkins:
I went ahead and tested this out and was able to use `renv::restore()` which setup everything for the virtual environment without issue. 
I think the one thing that would be helpful is to include some documentation on how to add python packages to the project, but that also could be something for a different PR.

allyhawkins (approved):
Just one minor typo, but looks good to me and the python environment seems to work for me locally.

PR #171, opened 2023-05-02
Update packages

jashapiro:
The changes here update the packages being used in renv.lock and Docker to the latest versions, but the real goal here it to debug the failing tests

The change I am most interested in is the addition of the `eds` package, which should make loading alevin data faster and maybe not hit memory limits. We'll see if it works and passes tests.


jashapiro:
It does seem that this has fixed the build issues, so this should be reviewed and merged before #170 and #169, which will be updated to include these changes.

allyhawkins (approved):
LGTM 👍

PR #179, opened 2023-07-05
Fix issue release template

sjspielman:
While working on #173, I noticed the issue template itself may be a bit out of date? The suggested line for code testing is missing our internal config (which seemed like it should be there given the subsequent `run_ids` protip), and the linked `run_ids` now points to some technologies, not run ids!

Updated in this PR!


jashapiro (approved):
LGTM

PR #181, opened 2023-07-12
Add functions to calculate silhouette width

### LINKED ISSUE #177: Add function to calculate batch ASW

Add a function for calculating `batch average silhouette width (ASW)` similar to the script in the `sc-data-integration` repo: https://github.com/AlexsLemonade/sc-data-integration/blob/main/scripts/utils/calculate-silhouette-width.R

cbethell:
Closes #177

This PR adds a `downsample_for_metrics.R` script with the `downsample_for_metrics()` function to perform downsampling, which is implemented in the added `calculate_silhouette_width()` function.

I also used `roxygen2` to add the Rd files and update documentation where necessary (as a package was added here)

I locally tested that these functions work, but it would probably be nice to ensure that they work for someone else as well.


R/downsample_for_metrics.R
```diff
@@ -0,0 +1,28 @@
+#' Downsample PCs for use in integration metric calculations 
+#'  
+#' @param integrated_pcs The full set of integrated pcs
```
allyhawkins, line 3:
```suggestion
#' @param pcs The PCs to downsample
```
These can be any type of PCs and don't have to be integrated.

R/downsample_for_metrics.R
```diff
@@ -0,0 +1,28 @@
...
+#' @param pc_name The name that allows access to the PCs. Example: fastMNN_PCA
+#' @param frac_cells The fraction of cells to downsample to
+#'
+#' @return List with two items: `pcs`, the downsampled PCs; `batch_labels`, the 
+#'  corresponding batch labels as integers for the downsampled PCs
+downsample_for_metrics <- function(integrated_pcs, pc_name, frac_cells) {
+  
+  num_cells <- nrow(integrated_pcs)
```
allyhawkins, line 11:
```suggestion
  num_cells <- nrow(pcs)
```

R/downsample_for_metrics.R
```diff
@@ -0,0 +1,28 @@
...
+  
+  # Determines rows to sample
+  downsampled_indices <- sample(1:num_cells,
+                                frac_cells*num_cells,
+                                replace = FALSE) 
+  
+  # Extract PCs for downsample
+  downsampled_integrated_pcs <- integrated_pcs[downsampled_indices, , drop = FALSE]
```
allyhawkins, line 19:
```suggestion
  downsampled_pcs <- pcs[downsampled_indices, , drop = FALSE]
```

R/downsample_for_metrics.R
```diff
@@ -0,0 +1,28 @@
+#' Downsample PCs for use in integration metric calculations 
+#'  
+#' @param integrated_pcs The full set of integrated pcs
+#' @param pc_name The name that allows access to the PCs. Example: fastMNN_PCA
```
allyhawkins, line 4: We don't actually use this argument so let's remove it.

R/downsample_for_metrics.R
```diff
@@ -0,0 +1,28 @@
...
+#' @param frac_cells The fraction of cells to downsample to
+#'
+#' @return List with two items: `pcs`, the downsampled PCs; `batch_labels`, the 
+#'  corresponding batch labels as integers for the downsampled PCs
+downsample_for_metrics <- function(integrated_pcs, pc_name, frac_cells) {
+  
+  num_cells <- nrow(integrated_pcs)
+  
+  # Determines rows to sample
+  downsampled_indices <- sample(1:num_cells,
```
allyhawkins, line 14: We should add a check before this that `frac_cells` is in the 0-1 range.

R/calculate_silhouette_width.R
```diff
@@ -0,0 +1,64 @@
...
+                                       frac_cells = 0.8,
+                                       nreps = 20,
+                                       seed = 2023,
+                                       batch_column = "library_id") {
+
+  # Set the seed for subsampling
+  set.seed(seed)
+
+  # Pull out the PCs or analogous reduction
+  pcs <- reducedDim(integrated_sce, pc_name)
```
allyhawkins, line 33: Before you get to this add a check for if the `pc_name` is present in `reducedDimNames()`.

R/calculate_silhouette_width.R
```diff
@@ -0,0 +1,64 @@
...
+#' @import bluster
+#' @import dplyr
+#'
+#' @export
+calculate_silhouette_width <- function(integrated_sce,
+                                       pc_name,
+                                       frac_cells = 0.8,
+                                       nreps = 20,
+                                       seed = 2023,
+                                       batch_column = "library_id") {
```
allyhawkins, line 27: We don't actually use this in the code you have right now. I think we need to add in this function that removes any NA's for batches and then also sets the rownames of the integrated pcs to batches. Let's add that as a PR stacked on this one? 

https://github.com/AlexsLemonade/sc-data-integration/blob/8152a52977a00f1bbf637053da853f676ad116bb/scripts/utils/integration-helpers.R#L559-L579

R/calculate_silhouette_width.R
```diff
@@ -0,0 +1,64 @@
...
+#'
+#' @import SingleCellExperiment
+#' @import bluster
+#' @import dplyr
+#'
+#' @export
+calculate_silhouette_width <- function(integrated_sce,
+                                       pc_name,
+                                       frac_cells = 0.8,
+                                       nreps = 20,
```
allyhawkins, line 25: We'll need checks that these fit the assumptions we have. `frac_cells` in 0-1 and `nreps` is an integer.

allyhawkins (commented):
Thanks for getting this started! Overall it's looking pretty good, we just need to make sure we incorporate checks. We want to make sure we think about every instance where this could break. Also we haven't yet actually accounted for the `batch_column` here, so we should do that. I would do that in a separate smaller PR though.

cbethell:
Thanks for the review @allyhawkins! I addressed all of your above comments with exception of https://github.com/AlexsLemonade/scpcaTools/pull/181#discussion_r1262956522 -- which will be in an upcoming stacked PR.

jashapiro:
I'm going to update this branch & merge in main to test styling again.

R/calculate_silhouette_width.R
```diff
@@ -0,0 +1,70 @@
...
+      dplyr::select(rep,
+        silhouette_width = width,
+        silhouette_cluster = cluster,
+        other_cluster = other
+      ) |>
+      dplyr::bind_rows()
+  })
+
+  # Add integration method into tibble
+  all_silhouette <- dplyr::mutate(all_silhouette, pc_name = pc_name)
```
allyhawkins, line 66:
```suggestion
  }) |> 
  # Add integration method into tibble
  dplyr::mutate(all_silhouette, pc_name = pc_name)
```
Just some minor simplification

cbethell: Ah, yes, thanks @allyhawkins...I added this line, but also added the `as.data.frame()` function right before due to the following error I received:

```
Error in UseMethod("mutate") : 
  no applicable method for 'mutate' applied to an object of class "list"
```

R/calculate_silhouette_width.R
```diff
@@ -0,0 +1,72 @@
...
+      dplyr::mutate(rep = rep) |>
+      dplyr::select(
+        rep,
+        silhouette_width = width,
+        silhouette_cluster = cluster,
+        other_cluster = other
+      ) |>
+      dplyr::bind_rows()
+    }) |>
+    as.data.frame() |>
```
jashapiro, line 66: I think you just want to move the `bind_rows()` out of the `map()` function here.

```suggestion
      ) 
    }) |>
    [as.data.frame()](dplyr::bind_rows()) |>
```

cbethell: Ah yes that did the trick, thank you!

R/downsample_for_metrics.R
allyhawkins, line 1: Just noticed this as I was starting to work on test functions. Can you update this filename to reflect the function name `downsample_pcs`?

cbethell: Done 👍

R/calculate_silhouette_width.R
```diff
@@ -0,0 +1,70 @@
...
+    stop("The provided `pc_name` cannot be found in the SingleCellExperiment object.")
+  }
+
+  # Check that `nreps` is an integer
+  if (!(is.integer(nreps))) {
+    stop("The provided `nreps` should be an integer.")
+  }
+
+  # Check that frac_cells is in range
+  if (frac_cells < 0 | frac_cells > 1) {
```
allyhawkins, line 41:
```suggestion
  if (frac_cells <= 0 | frac_cells > =1) {
```
We should also fail if it equals 0 or 1. Otherwise it isn't downsampling.

R/calculate_silhouette_width.R
```diff
@@ -0,0 +1,70 @@
...
+  # Set the seed for subsampling
+  set.seed(seed)
+
+  # Check that provided `pc_name` is present in SingleCellExperiment object
+  if (!pc_name %in% reducedDimNames(integrated_sce)) {
+    stop("The provided `pc_name` cannot be found in the SingleCellExperiment object.")
+  }
+
+  # Check that `nreps` is an integer
+  if (!(is.integer(nreps))) {
```
allyhawkins, line 36:
```suggestion
  if (!(nreps == round(nreps))) {
```
`is.integer` won't work sadly.

allyhawkins (commented):
A few more small comments I found.

cbethell:
Okay, now that #184 has been merged into this branch and I've implemented all outstanding comment -- I'm requesting a review from you both to hopefully wrap this up.

jashapiro:
/style @cbethell For styling to work for you from a comment like this, you need to have your membership in ALSF group here: https://github.com/orgs/AlexsLemonade/people to be set to "Public". But you might also need to be an admin of the repo.

jashapiro:
@cbethell But you can always run the styler locally!

R/downsample_pcs.R
```diff
@@ -0,0 +1,31 @@
...
+    frac_cells * num_cells,
+    replace = FALSE
+  )
+
+  # Extract PCs for downsample
+  downsampled_pcs <- pcs[downsampled_indices, , drop = FALSE]
+
+  return(
+    list(
+      pc_name = downsampled_pcs,
```
allyhawkins, line 27:
```suggestion
      pcs = downsampled_pcs,
```
I think we should change this to just `pcs` or `downsampled_pcs`. When you change this, line 61 in the ASW function will also need to be updated.

R/set_pc_rownames.R
```diff
@@ -0,0 +1,23 @@
+#' Label rownames of provided PCs and remove cells with NA batch label from PCs
+#'
+#' @param pcs The PCs with unlabeled rownames
+#' @param batches Vector of cell-wise batch information whose length is the number of
+#'   rows in `pcs`
+#'
+#' @return PCs with labeled rownames and NA batch cells removed
+set_pc_rownames <- function(pcs, batches) {
```
jashapiro, line 8: I think there was some discussion of the naming of this function, but since it also filters based on batch presence, I might rename it to something that reflects that. maybe `filter_pcs`. The fact that it adds names almost seems secondary to the filtering?

R/set_pc_rownames.R
```diff
@@ -0,0 +1,23 @@
...
+  # Check dimensions
+  if (nrow(pcs) != length(batches)) {
+    stop("Incompatable PC and batch information dimensions after removing NAs.")
+  }
+
+  # Remove NA batch cells
+  retain_indices <- which(!is.na(batches))
```
jashapiro, line 15: Do we want to check that this is not length zero and give an error if it is?

jashapiro (commented):
This looks good Chante. I have some minor suggestions, but also a couple that are a bit larger, like changing the return value from `downsample_pcs` to be just the pcs, like you have for `set_pc_rownames`.

As the suggestions are mostly for internal functions they really aren't too critical. If you want to move on to other tasks, that would be just fine too!

jashapiro:
The test failure here seems to be related to an ssh certificate failure. Hoping this resolves on its own.

R/calculate_silhouette_width.R
```diff
@@ -0,0 +1,77 @@
...
+#'  Generally this is either batches or cell types. Default is "library_id".
+#'
+#' @return Tibble with six columns: `rep`, representing the given downsampling replicate;
+#'   `silhouette_width`, the calculated silhouette width for the given `rep`; `silhouette_cluster`,
+#'   the assigned cluster for the cell during silhouette calculation, i.e. the true identity;
+#'   `other_cluster`, the other assigned for the cell during silhouette calculation;
+#'   `pc_name`, the name associated with the pc results
+#'
+#' @import SingleCellExperiment
+#' @import bluster
```
jashapiro, line 18: I think we don't need to import bluster here, as long as you are only calling it with `::` syntax. We generally don't want to add imports we don't use broadly, as they will actually be imported in the whole package, which can have unexpected results.
```suggestion
```

NAMESPACE
```diff
@@ -22,6 +23,7 @@ export(sce_to_anndata)
 export(sce_to_seurat)
 import(SingleCellExperiment)
 import(SummarizedExperiment)
+import(bluster)
```
jashapiro, line 26: Same as the previous comment (I think roxygen/build will handle this automatically, but still)
```suggestion
```

R/calculate_silhouette_width.R
```diff
@@ -0,0 +1,77 @@
...
+    stop("The provided `pc_name` cannot be found in the SingleCellExperiment object.")
+  }
+
+  # Check that `nreps` is an integer
+  if (!(nreps == round(nreps))) {
+    stop("The provided `nreps` should be an integer.")
+  }
+
+  # Check that frac_cells is in range
+  if (frac_cells <= 0 | frac_cells >= 1) {
```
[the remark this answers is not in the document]
cbethell: I added this check in the `downsample_pcs()` function -- let me know if the added modification is what you intended!

R/processing_pcs.R
```diff
@@ -0,0 +1,65 @@
...
+}
+
+
+
+#' Downsample PCs for use in integration metric calculations
+#'
+#' @param pcs The PCs to downsample, these PCs should contain batch labels as rownames
+#' @param frac_cells The fraction of cells to downsample to
+#'
+#' @return List with two items: `pcs`, the downsampled PCs; `batch_labels`, the
```
allyhawkins, line 40: This should be updated to reflect we are only returning the downsampled pcs.

R/processing_pcs.R
```diff
@@ -0,0 +1,65 @@
...
+#'  corresponding batch labels for the downsampled PCs
+downsample_pcs <- function(pcs, frac_cells) {
+
+  # Check that there is a minimum number of cells
+  num_cells <- nrow(pcs)
+  if (!num_cells > 1) {
+    stop("There are not enough cells to perform downsampling.")
+  }
```
jashapiro: I might modify this to check the cell count _after_ downsampling:
```suggestion
  if (frac_cells * num_cells < 1) {
    stop("Downsampling would result in no cells.")
  }
```

1 is probably too conservative, but it is fine for now.

R/processing_pcs.R
```diff
@@ -0,0 +1,65 @@
...
+#' @param pcs The PCs to downsample, these PCs should contain batch labels as rownames
+#' @param frac_cells The fraction of cells to downsample to
+#'
+#' @return List with two items: `pcs`, the downsampled PCs; `batch_labels`, the
+#'  corresponding batch labels for the downsampled PCs
+downsample_pcs <- function(pcs, frac_cells) {
+
+  # Check that there is a minimum number of cells
+  num_cells <- nrow(pcs)
+  if (!num_cells > 1) {
```
allyhawkins, line 46: Do we want to increase this? I feel like 1 is really low but I don't know what a good threshold is? 50? 25?

allyhawkins (approved):
This looks good from my end so I'm going to approve under the condition that any of Josh's comments will also get addressed. Also, that the minimum number of cells for downsampling is increased.

R/processing_pcs.R
cbethell: I am also not sure of a good threshold, I implemented 50 for now unless @jashapiro has objections.

jashapiro (commented):
Looks good! Just a couple little suggestions.

R/processing_pcs.R
```diff
@@ -0,0 +1,64 @@
...
+#' @param frac_cells The fraction of cells to downsample to
+#'
+#' @return The downsampled PCs
+downsample_pcs <- function(pcs, frac_cells) {
+
+  # Check that there is a minimum number of cells
+  num_cells <- nrow(pcs)
+  if (!num_cells > 50) {
+    stop("There are not enough cells to perform downsampling.")
+  }
```
jashapiro, line 47: My suggestion was marked as outdated, so I am re-adding it here... But making `min_cells` a parameter.

```suggestion
  if (frac_cells * num_cells < min_cells) {
    stop("Downsampling would result in fewer cells than the `min_cells` threshold.")
  }
```

R/processing_pcs.R
```diff
@@ -0,0 +1,64 @@
...
+
+
+
+#' Downsample PCs for use in integration metric calculations
+#'
+#' @param pcs The PCs to downsample, these PCs should contain batch labels as rownames
+#' @param frac_cells The fraction of cells to downsample to
+#'
+#' @return The downsampled PCs
+downsample_pcs <- function(pcs, frac_cells) {
```
jashapiro, line 41: Adding the parameter
```suggestion
#' @param pcs The PCs to downsample, these PCs should contain batch labels as rownames
#' @param frac_cells The fraction of cells to downsample to
#' @param min_cells The minimum number of cells after downsampling. Default: 50
#'
#' @return The downsampled PCs
downsample_pcs <- function(pcs, frac_cells, min_cells = 50) {
```

jashapiro (commented):
Updating my comment after it became outdated... Adding a parameter

cbethell:
Okay all comments have been addressed 🚀

R/calculate_silhouette_width.R
```diff
@@ -0,0 +1,76 @@
...
+#'
+#' @import SingleCellExperiment
+#'
+#' @export
+calculate_silhouette_width <- function(integrated_sce,
+                                       pc_name,
+                                       frac_cells = 0.8,
+                                       nreps = 20,
+                                       seed = NULL,
+                                       batch_column = "library_id") {
```
jashapiro, line 25: Just a very minor suggestion to change default order here
```suggestion
                                       batch_column = "library_id",
                                       frac_cells = 0.8,
                                       nreps = 20,
                                       seed = NULL) {
```

R/calculate_silhouette_width.R
```diff
@@ -0,0 +1,76 @@
+#' Calculate silhouette width scores from an integrated SCE object
+#'
+#' @param integrated_sce The integrated SCE object
+#' @param pc_name The name that allows access to the PCs. Example: fastMNN_PCA
+#' @param frac_cells The fraction of cells to downsample to. Default: 0.8
+#' @param nreps The number of times to repeat sub-sampling procedure. Default: 20
+#' @param seed Seed for initializing random sampling
+#' @param batch_column The variable in `integrated_sce` indicating the grouping of interest.
+#'  Generally this is either batches or cell types. Default is "library_id".
```
jashapiro, line 9: Matching the order of my other comment.
```suggestion
#' @param batch_column The variable in `integrated_sce` indicating the grouping of interest.
#'  Generally this is either batches or cell types. Default is "library_id".
#' @param frac_cells The fraction of cells to downsample to. Default: 0.8
#' @param nreps The number of times to repeat sub-sampling procedure. Default: 20
#' @param seed Seed for initializing random sampling
```

R/calculate_silhouette_width.R
```diff
@@ -0,0 +1,76 @@
+#' Calculate silhouette width scores from an integrated SCE object
```
jashapiro, line 1: We should probably expand this description a bit to say that this is performing replicated calculations on downsampled datat

jashapiro (approved):
Looks good, with just a couple tiny ordering suggestions and a request for a bit more detail in the function description.

R/calculate_silhouette_width.R
```diff
@@ -0,0 +1,77 @@
+#' Calculate silhouette width scores from an integrated SCE object.
+#' This function performs replicated calculations on downsampled data.
```
jashapiro, line 2: It looks like the documentation function isn't working properly (and the test failure is for documentation mismatches): Can you rerun documentation on this @cbethell?

I'm also separating the "title" from "description" here by adding a blank line in the documentation script.

```suggestion
#'
#' This function performs replicated calculations on downsampled data.
```

jashapiro:
Hmmm... the checks don't seem to be running after text commands. I know why, but I probably won't fix it today. Gonna push a whitespace change to make the action run so you can merge this.

jashapiro:
Better idea: @allyhawkins can merge in her test functions?

allyhawkins:
> Better idea: @allyhawkins can merge in her test functions?

Done! 🤞

R/calculate_silhouette_width.R
```diff
@@ -0,0 +1,78 @@
...
+    downsampled <- downsample_pcs(labeled_pcs, frac_cells)
+    # Calculate batch ASW and add into final tibble
+    bluster::approxSilhouette(downsampled, rownames(downsampled)) |>
+      tibble::as_tibble() |>
+      dplyr::mutate(rep = rep) |>
+      dplyr::select(
+        rep,
+        silhouette_width = width,
+        silhouette_cluster = cluster,
+        other_cluster = other
```
jashapiro, line 69: A really minor thing that I noticed looking at the test results: 

When using variables in the tidyverse like this, R checks will quietly complain that they are not defined. To solve this we can use `.data` within functions like this:
```suggestion
      dplyr::select(
        .data$rep,
        silhouette_width = .data$width,
        silhouette_cluster = .data$cluster,
        other_cluster = .data$other
```

R/calculate_silhouette_width.R
```diff
@@ -0,0 +1,78 @@
...
+#' @param nreps The number of times to repeat sub-sampling procedure. Default: 20
+#' @param seed Seed for initializing random sampling
+#'
+#' @return Tibble with five columns: `rep`, representing the given downsampling replicate;
+#'   `silhouette_width`, the calculated silhouette width for the given `rep`; `silhouette_cluster`,
+#'   the assigned cluster for the cell during silhouette calculation, i.e. the true identity;
+#'   `other_cluster`, the other assigned for the cell during silhouette calculation;
+#'   `pc_name`, the name associated with the pc results
+#'
+#' @import SingleCellExperiment
```
jashapiro, line 19: This doesn't really matter as it is imported elsewhere in the package but it is good practice to add it where it is used: 
```suggestion
#' @import SingleCellExperiment
#' @importFrom rlang .data
```

R/calculate_silhouette_width.R
```diff
@@ -0,0 +1,79 @@
...
+    downsampled <- downsample_pcs(labeled_pcs, frac_cells)
+    # Calculate batch ASW and add into final tibble
+    bluster::approxSilhouette(downsampled, rownames(downsampled)) |>
+      tibble::as_tibble() |>
+      dplyr::mutate(rep = rep) |>
+      dplyr::select(
+        .data$rep,
+        silhouette_width = .data$width,
+        silhouette_cluster = .data$cluster,
+        other_cluster = .data$other
```
jashapiro, line 70: Ah! it turns out I was wrong about this! This method is deprecated, so now we get a new warning!

The new solution is apparently just to quote the variable name (which was my first inclination, but other places in the code use `.data` so I thought that was the way.
```suggestion
      dplyr::select(
        "rep",
        silhouette_width = "width",
        silhouette_cluster = "cluster",
        other_cluster = "other"
```

cbethell: Ah okay, I'll test this one out! Thank you!

PR #186, opened 2023-07-18
Add ontoProc

### LINKED ISSUE #185: Add ontoProc to image
From https://github.com/AlexsLemonade/scpca-nf/pull/382#discussion_r1266805645

> I'm also noting that `ontoProc` does not seem to be in our image yet, so I think it will need to be added to scpcaTools?

Indeed, add to the image (`renv`, `docker`).

sjspielman:
Closes #185 

This PR adds `ontoProc`, after much 🎢 , to the image. `renv` decided to also add a couple of bioc repos to the lockfile which didn't seem like too much of a problem, but I can remove them if we want!


jashapiro (approved):
LGTM, and to the robots!

sjspielman:
🤖

PR #187, opened 2023-07-18
Add test functions for ASW 

### LINKED ISSUE #182: Add test functions for calculating ASW
Now that we are adding a function to calculate ASW in #181, we will also need to add test functions for the new additions. We only need test functions for any exported functions added in that PR.

allyhawkins:
Closes #182 
Stacked on #181 

This PR adds in test functions for all of the functions being added in #181. To set up actually running the test function I followed a similar format to the tests for the integration functions. I created a merged object using the `sim_sce` function and then added in PCs and batch labels. 

Overall I think this was mostly straight forward, but are there any other edge cases that we should test that I missed?


tests/testthat/test-integration_metrics.R
```diff
@@ -0,0 +1,109 @@
...
+# set up colData rownames as batch-barcode
+new_rownames <- glue::glue("{batches}-{barcodes}")
+rownames(colData(merged_sce)) <- new_rownames
+
+# Add "sample" to colData
+colData(merged_sce)$sample <- batches
+
+# add PCs for testing (exact numbers don't matter)
+# make a 300x100 matrix
+reducedDim(merged_sce, "PCA") <- matrix(runif(300 * 100, min = 0, max = 100), nrow = 300)
```
jashapiro, line 16: I wonder if we might set these up to get specific ASW values? Probably not necessary.

tests/testthat/test-integration_metrics.R
```diff
@@ -0,0 +1,109 @@
...
+
+test_that("`downsample_pcs` works as expected", {
+
+  downsampled <- downsample_pcs(pcs, frac_cells = 0.8)
+  downsampled_pcs <- downsampled$pcs
+  downsampled_batch_labels <- downsampled$batch_labels
+
+  # check that the correct number of cells were downsampled
+  expect_true(length(rownames(downsampled_pcs)) == 0.8*length(rownames(pcs)))
+  expect_true(length(downsampled_batch_labels) == 0.8*length(batches))
```
jashapiro, line 55: Since sampling does not allow fractional samples...

We might want to use a test matrix with a non-multiple-of-5 number of rows to make this aspect of the test more applicable.
```suggestion
  expect_true(length(rownames(downsampled_pcs)) == floor(0.8*length(rownames(pcs))))
  expect_true(length(downsampled_batch_labels) == floor(0.8*length(batches)))
```

tests/testthat/test-integration_metrics.R
```diff
@@ -0,0 +1,109 @@
...
+
+  # check that the columns in downsampled were in the original pcs
+  expect_true(length(colnames(downsampled_pcs)) == length(colnames(pcs)))
```
jashapiro, line 58: Is there any way to make sure the labels are not jumbled during the downsampling? I'm not sure what that would be, but it seems an important check.

tests/testthat/test-integration_metrics.R
```diff
@@ -0,0 +1,109 @@
...
+    "other_cluster", "pc_name"
+  )
+  # check column names
+  expect_true(all(expected_cols %in% colnames(asw)))
+
+  # check that pc name is PCA
+  expect_true(all(asw$pc_name == "PCA"))
+
+  # check that nreps equal 1-20
+  expect_true(all(asw$rep %in% 1:20))
```
jashapiro, line 87: Make this more specific?

```suggestion
  expect_true(all(sort(unique(asw$rep)) == 1:nreps))
```

I also replaced 20 with `nreps`, because I think we should specify this, rather than relying on the default value here. (we probably don't need 20 reps for the test either!)

tests/testthat/test-integration_metrics.R
```diff
@@ -0,0 +1,109 @@
...
+  short_batches <- c("a", "b", "c")
+  expect_error(scpcaTools:::set_pc_rownames(pcs, short_batches))
+
+})
+
+test_that("`downsample_pcs` works as expected", {
+
+  downsampled <- downsample_pcs(pcs, frac_cells = 0.8)
+  downsampled_pcs <- downsampled$pcs
+  downsampled_batch_labels <- downsampled$batch_labels
```
jashapiro, line 51: Just noting that I suggested elsewhere that we might just return the pc matrix here with the appropriate rownames, which would allow us to reduce this to just 
```
downsampled_batch_labels <- rownames(downsampled)
```

tests/testthat/test-integration_metrics.R
```diff
@@ -0,0 +1,109 @@
...
+test_that("`set_pc_rownames` fails without batch labels", {
+
+  short_batches <- c("a", "b", "c")
+  expect_error(scpcaTools:::set_pc_rownames(pcs, short_batches))
+
+})
+
+test_that("`downsample_pcs` works as expected", {
+
+  downsampled <- downsample_pcs(pcs, frac_cells = 0.8)
```
jashapiro, line 49: Just to avoid later potential errors, we should probably set this as a variable and that rather than worrying about having to change numbers. 
```suggestion
  test_frac = 0.8
  downsampled <- downsample_pcs(pcs, frac_cells = test_frac)
```

jashapiro (commented):
Looks good overall. I had a few suggestions that are not too major. The big one is maybe changing the test matrix (or `frac_cells`) to make some errors more likely. 

Probably also needs a quick styling run, but whenver.

tests/testthat/test-integration_metrics.R
allyhawkins: I think this is actually okay, because we have 300 cells and downsampling to 80% of those should give you 240 cells? Unless I'm missing something.

tests/testthat/test-integration_metrics.R
allyhawkins: Do you mean the cell barcodes or the PC labels? There actually are no PC column names which is why I checked the length instead of using `all.equal`.

tests/testthat/test-integration_metrics.R
jashapiro: Yes, in this case it is fine, but I think we might want to check that it works as expected with something like 303 cells.

tests/testthat/test-integration_metrics.R
allyhawkins: > It would be nice to have some kind of check that the rows in the downsampled data are still correctly named, but I don't have a good thought about how to do that.

That's what I'm doing in line 26 and 36 - checking that the rownames of the downsampled PCs are the same as the batch labels, which is what is expected.

tests/testthat/test-integration_metrics.R
allyhawkins: Ah I mis-read non-multiple of 5 as multiple of 5. That makes sense.

tests/testthat/test-integration_metrics.R
jashapiro: Right, that works for the labeling, but it is harder to do for the downsampling. I don't think it is likely given the way it is implemented now, but when we were returning two separate objects for the matrix and the labels, it was possible that you could mess up the proper pairings, which would be bad.

allyhawkins:
I incorporated most of the suggestions here including increasing the number of cells to be 303. I did not alter the matrix values to get specific ASW values... If we really think it's necessary we could do that. Another thought that just came to me is we could check the range of ASW values. They should be between -1 and 1.

tests/testthat/test-integration_metrics.R
```diff
@@ -0,0 +1,116 @@
...
+  # expect rownames to be equal to only the batches without NA
+  named_pcs <- filter_pcs(pcs, batches)
+  expect_true(all.equal(rownames(named_pcs), batches[!is.na(batches)]))
+  expect_false(length(rownames(named_pcs)) == length(batches))
+})
+
+test_that("`filter_pcs` fails without batch labels", {
+
+  all_na_batches <- NA
+  expect_error(scpcaTools:::filter_pcs(pcs, all_na_batches))
```
jashapiro, line 43: I don't think you need to name `scpcaTools` here?
```suggestion
  expect_error(filter_pcs(pcs, all_na_batches))
```

jashapiro (approved):
LGTM!

tests/testthat/test-integration_metrics.R
allyhawkins: No I don't, I must have forgotten to remove it from when I was testing things locally.

cbethell (approved):
Works and LGTM!

PR #188, opened 2023-07-19
Add function to perform clustering on SCE

### LINKED ISSUE #178: Function for performing clustering on a single SCE object
> I may be a little confused here, but I think we could simplify this function if was assume that clustering has already been performed on both the integrated and unintegrated data sets? I'm not sure this function should care about the PCs at all.

Assuming this, we need a separate function solely for clustering a single SCE object. That function must be run on every individual SCE and integrated SCE before calculating metrics. That may be the way to go. Either way, I agree that we should add clustering parameters as inputs. If you agree with this, the input you described makes more sense.

_Originally posted by @allyhawkins in https://github.com/AlexsLemonade/scpcaTools/issues/176#issuecomment-1611891620_

cbethell:
Closes #178 

This PR adds an internal function and documentation to perform kmeans clustering on a single SCE object, across a range of k values. 

I included just kmeans clustering here to keep things as simple as possible at this time -- but I am also wondering if we want to include the option to specify the desired cluster type (kmeans vs graph-based walktrap and louvain). What do folks think?


jashapiro:
I would have expected this method to be agnostic to the clustering method. I do not expect we will use k-means clustering much in practice, so I would have wanted to see this implemented to use `bluster` allowing the user to use whatever clustering method they wanted via passing in a `BLUSPARAM` object.

cbethell:
> I would have expected this method to be agnostic to the clustering method. I do not expect we will use k-means clustering much in practice, so I would have wanted to see this implemented to use `bluster` allowing the user to use whatever clustering method they wanted via passing in a `BLUSPARAM` object. This might specify the output column name that would get added to the SCE object.
> 
> To get the k-means range, you could then write a separate function that calls the `cluster_sce` object some number of times to get the range of outputs stored in the SCE object

Ah okay so we want this function to work on a single SCE with a single nearest neighbors or k value -- and within this function we would implement `bluster::clusterRows()`

jashapiro:
> Ah okay so we want this function to work on a single SCE with a single nearest neighbors or k value -- and within this function we would implement `bluster::clusterRows()`

Yup! I would also set it up so it has a `...` argument that would pass anything the user might want (like multiprocessing) along to `clusterRows()` without us having to add anything.

One other thing is that the function here (and elsewhere in the package) should use `seed = NULL` as the default, so that a user can set a single seed in their script and not have to worry about it being reset unexpectedly. See for example how we implemented it in [`add_miQC.R`](https://github.com/AlexsLemonade/scpcaTools/blob/1d3a05221c37ebe8f9cfaff79f7e566830ab1b54/R/add_miQC.R#L17C1-L40C17)

cbethell:
> > Ah okay so we want this function to work on a single SCE with a single nearest neighbors or k value -- and within this function we would implement `bluster::clusterRows()`
> 
> Yup! I would also set it up so it has a `...` argument that would pass anything the user might want (like multiprocessing) along to `clusterRows()` without us having to add anything.
> 
> One other thing is that the function here (and elsewhere in the package) should use `seed = NULL` as the default, so that a user can set a single seed in their script and not have to worry about it being reset unexpectedly. See for example how we implemented it in [`add_miQC.R`](https://github.com/AlexsLemonade/scpcaTools/blob/1d3a05221c37ebe8f9cfaff79f7e566830ab1b54/R/add_miQC.R#L17C1-L40C17)

Okay I did the above and also added a `cluster_type` argument to help specify whether to use `KMeansParam` or `NNGraphParam` in the `clusterRows` function.

R/cluster_sce.R
```diff
@@ -0,0 +1,45 @@
...
+        cluster.fun = cluster_type,
+      )
+    )
+  } else {
+    stop("Clustering type is not valid. Please specify 'kmeans', 'walktrap' or 'louvain' clustering.")
+  }
+
+  # Set column name to store results in
+  cluster_name <- paste(cluster_type, k, sep = "_")
+  sce[[cluster_name]] <- clustering_results
```
jashapiro, line 42: I was thinking something even simpler!
```suggestion
cluster_sce <- function(sce,
                        pca_name,
                        BLUSPARAM,
                        cluster_name,
                        seed = NULL
                        ...) {
  # Set the seed
  set.seed(seed)

  # Extract PCs
  pcs <- reducedDim(sce, pca_name)
  clustering_results <- bluster::clusterRows(pcs, BLUSPARAM, ...)
  sce[[cluster_name]] <- clustering_results
```

This relies on the outer function to define the `cluster_name`, but it makes the internals much simpler, and more flexible. 

Note that I also changed `pca_column` to `pca_name`, as it is not really a column.

R/cluster_sce.R
```diff
@@ -0,0 +1,27 @@
+#' Perform clustering on a SingleCellExperiment object
+#'
+#' @param sce SingleCellExperiment object that requires clustering
+#' @param pca_name The name used to access the PCA results
+#' @param BLUSPARAM A BlusterParam object specifying the clustering algorithm to
+#' use - can be `KmeansParam` or `NNGraphParam`
+#' @param cluster_name The type of clustering method to be performed - can be
```
allyhawkins, line 7: I think I would change this variable to reflect that it's just being used to add a column with clustering results to the SCE. Maybe `cluster_column_name`?

R/cluster_sce.R
```diff
@@ -0,0 +1,27 @@
...
+#' @return SingleCellExperiment object containing clustering results
+cluster_sce <- function(sce,
+                        pca_name,
+                        BLUSPARAM,
+                        cluster_name,
+                        seed = NULL,
+                        ...) {
+  # Set the seed
+  set.seed(seed)
+
```
allyhawkins, line 20: You will also need to add in checks here that the pcs exist in the SCE object and then we will probably want to check the `BLUSPARAM` object too. I'm not quite sure how we would do that though...

jashapiro: I would probably leave the `BLUSPARAM` checks to `bluster` in this case.

R/cluster_sce.R
```diff
@@ -0,0 +1,27 @@
+#' Perform clustering on a SingleCellExperiment object
+#'
+#' @param sce SingleCellExperiment object that requires clustering
+#' @param pca_name The name used to access the PCA results
+#' @param BLUSPARAM A BlusterParam object specifying the clustering algorithm to
+#' use - can be `KmeansParam` or `NNGraphParam`
```
jashapiro, line 6: We can be open to any kind of clustering here, if we want to be creative later.
```suggestion
#' @param BLUSPARAM A BlusterParam object specifying the clustering algorithm and parameters to use
```

R/cluster_sce.R
jashapiro: cluster_column_prefix maybe? Again here we probably don't need to have the function say what is allowed.

R/cluster_sce.R
```diff
@@ -0,0 +1,35 @@
+#' Perform clustering on a SingleCellExperiment object
+#'
+#' @param sce SingleCellExperiment object that requires clustering
+#' @param pca_name The name used to access the PCA results
+#' @param BLUSPARAM A BlusterParam object specifying the clustering algorithm to
+#' use
```
allyhawkins, line 6:
```suggestion
#'   use and any additional clustering options
```

R/cluster_sce.R
```diff
@@ -0,0 +1,35 @@
...
+#' @param cluster_column_name The name of the column to store the clustering
+#' results in the SCE; for naming you may want to include the type of clustering
+#' and k value of centers, e.g. `kmeans_10`
```
allyhawkins, line 9:
```suggestion
#'   results in the SCE; for naming you may want to include the type of clustering
#'   and k value of centers, e.g. `kmeans_10`
```
Just spacing

allyhawkins (approved):
This looks good to me. Just a few docs updates that need to be addressed. 
Additionally, can you file an issue to add tests for this function?

R/cluster_sce.R
```diff
@@ -0,0 +1,35 @@
...
+#' @param seed Seed for reproducibility of clustering results
+#' @param ... Allows for additional arguments to be provided to `clusterRows()`
```
sjspielman, line 11:
```suggestion
#' @param ... Additional arguments to provide to `bluster::clusterRows()`
```

R/cluster_sce.R
```diff
@@ -0,0 +1,35 @@
...
+#'
```
sjspielman, line 12: We probably need this line in roxygen comments - 

```suggestion
#' @import SingleCellExperiment
```

R/cluster_sce.R
```diff
@@ -0,0 +1,35 @@
+#' Perform clustering on a SingleCellExperiment object
+#'
+#' @param sce SingleCellExperiment object that requires clustering
+#' @param pca_name The name used to access the PCA results
```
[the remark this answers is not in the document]
sjspielman: I also wonder if we want this to default to `PCA`? If so, add to docs here.

R/cluster_sce.R
```diff
@@ -0,0 +1,35 @@
...
+#' use
+#' @param cluster_column_name The name of the column to store the clustering
+#' results in the SCE; for naming you may want to include the type of clustering
+#' and k value of centers, e.g. `kmeans_10`
+#' @param seed Seed for reproducibility of clustering results
+#' @param ... Allows for additional arguments to be provided to `clusterRows()`
+#'
+#' @return SingleCellExperiment object containing clustering results
+cluster_sce <- function(sce,
+                        pca_name,
```
sjspielman, line 15: If we want a default - 

```suggestion
                        pca_name = "PCA",
```

sjspielman (commented):
Looks great to me! Just a few comments for function documentation and imports, as well as a possible default we might want to use.

R/cluster_sce.R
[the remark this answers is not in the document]
cbethell: We do expect values like "fastMNN_PCA" and "harmony_PCA" when we implement this function in downstream analyses but I do not think making `PCA` the default here would hurt so I made that modification!

cbethell:
> This looks good to me. Just a few docs updates that need to be addressed. Additionally, can you file an issue to add tests for this function?

Filed #192 👍

allyhawkins (approved):
Good to go 👍

R/cluster_sce.R
```diff
@@ -0,0 +1,42 @@
...
+#' @param seed Seed for reproducibility of clustering results
+#' @param ... Additional arguments to provide to `bluster::clusterRows()`
+#'
+#' @import SingleCellExperiment
+#'
+#' @return SingleCellExperiment object containing clustering results
+#'
+#' @examples
+#'   Perform K-means clustering with 10 centers
+#'   cluster_sce(sce, "PCA", bluster::KmeansParam(centers = 10), "kmeans_10")
```
allyhawkins, line 20: This is missing a `\dontrun{}` around it and also a `#` before the comment. 
```suggestion
#' @examples
#'  \dontrun{
#'   # Perform K-means clustering with 10 centers
#'   cluster_sce(sce, "PCA", bluster::KmeansParam(centers = 10), "kmeans_10")
#'  }
```
Also to test out checks locally I would encourage you to run `devtools::check()` while in the project if you are not doing that already. It will help catch things like these.

sjspielman: good catch, @allyhawkins! I definitely second running `devtools::check()` which I should have done as part of review but did not (happily the 🤖 does though!) - protip, use `cmd+shift+e` shortcut in RStudio.

PR #189, opened 2023-07-20
Remove rprofile before setup

jashapiro:
So it looked like having an Rprofile made the documentation part of the GHA fail, so I am adding a step to remove that file before setting up R and dependencies. Note that the deletion of that file is not part of the changes that are added before the commit, so that should be just a temporary change local to the action run.

Will try to  test to see if this works before requesting a reviewer, but may run into challenges.

### DIFF
--- a/.github/workflows/pr-commands.yaml
+++ b/.github/workflows/pr-commands.yaml
@@ -3,15 +3,12 @@
 on:
   issue_comment:
     types: [created]
-  push:
-    branches:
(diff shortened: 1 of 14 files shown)


jashapiro:
Now tested, and test commits reverted. I left in the automatic commit that updated the documentation to match the example code changes that seem to have happened since the last time we ran a full documentation command

allyhawkins (approved):
👍

PR #191, opened 2023-07-24
Add within-batch ARI function

### LINKED ISSUE #176: Add function to calculate within-batch ARI

Add a function for calculating `within-batch ARI` similar to the script in the `sc-data-integration` repo: https://github.com/AlexsLemonade/sc-data-integration/blob/main/scripts/utils/calculate-ARI.R

cbethell:
Closes #176

This PR adds functionality to calculate within-batch ARI values given a list of individual SCEs and the integrated SCE object.

This is currently in draft form due to the following error I am receiving at the `pairwiseRand()` step:

```
Error in (function (classes, fdef, mtable)  : 
  unable to find an inherited method for function ‘NSBS’ for signature ‘"CompressedLogicalList"’
```

I am currently working to solve this error but feedback is welcome in the meantime!

**Edit:** I was able to solve the error by isolating the two vectors in the `pairwiseRand()` function per https://github.com/AlexsLemonade/scpcaTools/pull/191#discussion_r1272721910

For reviewers, please let me know if there is any refactoring that you believe should be done here.


R/calculate_within_batch_ari.R
```diff
@@ -0,0 +1,89 @@
...
+          BLUSPARAM = bluster::NNGraphParam(cluster.fun = "louvain", type = "jaccard"),
+          cluster_column_name = "individual_clusters"
+        )
+
+      # Extract clusters from integrated clustering for batch
+      clusters_to_keep <- grep(batch, integrated_clustering_result[[batch_column]])
+      batch_integrated_clusters <- integrated_clustering_result[clusters_to_keep]
+
+      # Calculate ARI between pre-integrated clustering and post-integrated clustering for the given batch
+      ari <- bluster::pairwiseRand(individual_clustering_result,
```
allyhawkins, line 73:
```suggestion
      ari <- bluster::pairwiseRand(individual_clustering_result[["individual_clusters"]],
```
I didn't look at everything in detail, but it looks like you are trying to pass an sce object to `pairwiseRand`. You will want to pass in just the two vectors that have clustering results.

sjspielman (commented):
Overall this looks really good! But, while I was testing this code out, it seems we have conflicting `pc_name` values being used:
- Line 49 expects the PC name associated with the integrated SCE
- Line 63 expects the PC name associated with _individual_ SCEs

So, I think we're going to need two arguments to this function:
- `integrated_pc_name` (no default value)
- `individual_pc_name` (maybe use "PCA" default value here, @allyhawkins any thoughts?)

Once that is all set, I'll come back for a more thorough review!

Also, I am realizing something about the `cluster_sce()` function that was previously reviewed... It's argument name is `pca_name`, but we seem to use `pc_name` more commonly in `scpcaTools` (here and in the ASW function), so we might want to open a separate issue to have `cluster_sce()` also use `pc_name`. This can potentially be taken care of during #192 also.

R/calculate_within_batch_ari.R
```diff
@@ -2,7 +2,10 @@
 #'
 #'
 #' @param individual_sce_list A list of individual SCE objects
-#' @param pc_name The name that allows access to the PCs. Example: fastMNN_PCA
+#' @param individual_pc_name The name that allows access to the PCs in the
+#'  individual SCE objects. Default is "PCA".
+#' @param integrated_pc_name The name that allows access to the PCs in the
+#'  integrated SCE object.  Example: "fastMNN_PCA".
```
allyhawkins, line 8: I think what we actually want to do here is just have one `pc_name` argument that is only used for the integrated object. For the individual object, I think it should be required that it has a `PCA` stored in it. There shouldn't be any other options for the individual object. 
Maybe what will make this more straight forward is referring to `integrated_sce` as `merged_sce`? And I would do that across the board for other functions that use that argument name. These should be able to be used with a merged object regardless of if there is batch corrected embeddings stored or not. 
```suggestion
#' @param pc_name The name that allows access to the PCs in the
#'  integrated SCE object.  Example: "fastMNN_PCA".
```

R/calculate_within_batch_ari.R
```diff
@@ -2,7 +2,10 @@
...
 #' @param integrated_sce The integrated SCE object
```
allyhawkins, line 9:
```suggestion
#' @param merged_sce The merged SCE object containing data from multiple batches
```

R/calculate_within_batch_ari.R
```diff
@@ -60,7 +64,7 @@ calculate_within_batch_ari <- function(individual_sce_list,
       # Cluster pc matrix for specified batch
       individual_clustering_result <- individual_sce_list[[batch]] |>
         cluster_sce(
-          pca_name = pc_name,
+          pca_name = individual_pc_name,
```
allyhawkins, line 67:
```suggestion
          pca_name = "PCA",
```

R/calculate_within_batch_ari.R
```diff
@@ -0,0 +1,94 @@
...
+    if (!(identical(
+      sort(batch_ids),
+      sort(unique(colData(integrated_sce)[, batch_column]))
+    ))
+    ) {
+      stop("Names of provided SCE objects included in the individual SCE object
+           do not match batch IDs present in the batch_column of the integrated object")
+    }
+  }
+
```
allyhawkins, line 46: You need a check here for the `pc_name` being present in the `merged_sce` object (or integrated object as it is in the code now).

R/calculate_within_batch_ari.R
```diff
@@ -0,0 +1,94 @@
...
+                                       individual_pc_name,
+                                       integrated_pc_name,
+                                       integrated_sce,
+                                       batch_column = "library_id",
+                                       seed = NULL) {
+  # Set the seed for subsampling and clustering
+  set.seed(seed)
+
+  # Check that list of SCE objects is named
+  batch_ids <- names(individual_sce_list)
```
allyhawkins, line 32: Before this you need to check that the `batch_column` exists in the colnames.

R/calculate_within_batch_ari.R
```diff
@@ -0,0 +1,94 @@
...
+
+  # Cluster integrated pcs only one time
+  integrated_clustering_result <- integrated_sce |>
+    cluster_sce(
+      pca_name = integrated_pc_name,
+      BLUSPARAM = bluster::NNGraphParam(cluster.fun = "louvain", type = "jaccard"),
+      cluster_column_name = "integrated_clusters"
+    )
+
+  rownames(integrated_pcs) <- integrated_clustering_result[["integrated_clusters"]] # make sure to set the names with the batch ids
```
allyhawkins, line 58: Do you need this here? I'm not sure what it's doing since it's adding the cluster assignments as the rownames?

R/calculate_within_batch_ari.R
```diff
@@ -0,0 +1,94 @@
...
+      stop("Names of provided SCE objects included in the individual SCE object
+           do not match batch IDs present in the batch_column of the integrated object")
+    }
+  }
+
+  # Pull out the PCs or analogous reduction from integrated object
+  integrated_pcs <- reducedDim(integrated_sce, integrated_pc_name)
+
+  # Cluster integrated pcs only one time
+  integrated_clustering_result <- integrated_sce |>
```
allyhawkins, line 51:
```suggestion
  merged_sce <- merged_sce |>
```
I don't think we need to save an entirely new object. This function is adding in a new column `integrated_clusters` to an existing SCE so we can use the same object. I would then later pull out the cluster column so it's easier to follow what's happening. 
This change depends on you taking my suggestion to change to `merged_sce`.

R/calculate_within_batch_ari.R
```diff
@@ -0,0 +1,94 @@
...
+    cluster_sce(
+      pca_name = integrated_pc_name,
+      BLUSPARAM = bluster::NNGraphParam(cluster.fun = "louvain", type = "jaccard"),
+      cluster_column_name = "integrated_clusters"
+    )
+
```
allyhawkins, line 57:
```suggestion
merged_clusters <- merged_sce$merged_clusters |>
  purrr::set_names(merged_sce[[batch_column]])
```
Then you can directly use the `merged_clusters` vector below.

R/calculate_within_batch_ari.R
```diff
@@ -0,0 +1,94 @@
...
+      individual_clustering_result <- individual_sce_list[[batch]] |>
+        cluster_sce(
+          pca_name = individual_pc_name,
+          BLUSPARAM = bluster::NNGraphParam(cluster.fun = "louvain", type = "jaccard"),
+          cluster_column_name = "individual_clusters"
+        )
+
+      # Extract clusters from integrated clustering for batch
+      clusters_to_keep <- grep(batch, integrated_clustering_result[[batch_column]])
+      batch_integrated_clusters <- integrated_clustering_result[clusters_to_keep]
```
allyhawkins, line 74:
```suggestion
      clusters_to_keep <- grep(batch, merged_sce[[batch_column]])
      batch_integrated_clusters <- merged_clusters[clusters_to_keep]
```
This should be directly subsetting the vector with the merged clustering results, not the SCE object.

R/calculate_within_batch_ari.R
```diff
@@ -0,0 +1,94 @@
...
+
+      # Calculate ARI between pre-integrated clustering and post-integrated clustering for the given batch
+      ari <- bluster::pairwiseRand(individual_clustering_result[["individual_clusters"]],
+        batch_integrated_clusters[["integrated_clusters"]],
+        mode = "index"
+      )
```
allyhawkins, line 80:
```suggestion
      ari <- bluster::pairwiseRand(individual_clustering_result[["individual_clusters"]],
        batch_integrated_clusters,
        mode = "index"
      )
```

R/calculate_within_batch_ari.R
cbethell: I made the change from `integrated_sce` to `merged_sce` in the silhouette width script here as well 👍

R/calculate_within_batch_ari.R
```diff
@@ -0,0 +1,100 @@
...
+                                       pc_name,
+                                       merged_sce,
+                                       batch_column = "library_id",
+                                       seed = NULL) {
+  # Set the seed for subsampling and clustering
+  set.seed(seed)
+
+  # Check that `batch_column` is in colData of SCE
+  if (!batch_column %in% colnames(colData(merged_sce))) {
+    stop("The specified batch column is missing from the colData of the SingleCellExperiment object.")
```
allyhawkins, line 30:
```suggestion
    stop("The specified batch column is missing from the colData of the `merged_sce`.")
```
Since there are two objects being input, let's be specific.

R/calculate_within_batch_ari.R
```diff
@@ -0,0 +1,100 @@
...
+    ))
+    ) {
+      stop("Names of provided SCE objects included in the individual SCE object
+           do not match batch IDs present in the batch_column of the merged object")
+    }
+  }
+
+  # Check that `pc_name` is in merged SCE object
+  if (!pc_name %in% reducedDimNames(merged_sce)) {
+    stop("The provided `pc_name` cannot be found in the SingleCellExperiment object.")
```
allyhawkins, line 51:
```suggestion
    stop("The provided `pc_name` cannot be found in the `merged_sce`.")
```

R/calculate_within_batch_ari.R
```diff
@@ -0,0 +1,100 @@
...
+          BLUSPARAM = bluster::NNGraphParam(cluster.fun = "louvain", type = "jaccard"),
+          cluster_column_name = "individual_clusters"
+        )
+
+      # Extract clusters from merged clustering for batch
+      clusters_to_keep <- grep(batch, merged_sce[[batch_column]])
+      batch_merged_clusters <- merged_clusters[clusters_to_keep]
+
+      # Calculate ARI between pre-merged clustering and post-merged clustering for the given batch
+      ari <- bluster::pairwiseRand(individual_clustering_result[["individual_clusters"]],
```
allyhawkins, line 84:
```suggestion
      ari <- bluster::pairwiseRand(individual_clustering_result$individual_clusters,
```

allyhawkins (approved):
This looks much cleaner! I just have a few minor comments, but I don't need to see it again.

R/calculate_within_batch_ari.R
```diff
@@ -0,0 +1,100 @@
+#' Calculate within-batch ARI scores from a merged SCE object.
+#'
+#'
+#' @param individual_sce_list A list of individual SCE objects
```
sjspielman, line 4: I might add a bit more here (but you will probably need to fix spacing locally!!)
```suggestion
#' @param individual_sce_list A named list of individual SCE objects. It is assumed these have a reduced dimension slot with principal components named "PCA".
```

sjspielman (approved):
👍 
Left one quick docs suggestion, and you'll want to pull this down and check spacing in the code is ok (github suggestions are very bad at spacing...). Then, you'll also want to re-run `devtools::document()` to update the associated `.Rd` file with these docs changes. Should be all good then!

PR #195, opened 2023-07-27
Add tests for `cluster_sce` function

### LINKED ISSUE #192: Add test functions for clustering SCE
Once we add the function to cluster a SCE object in #188, we will also want to add test functions for the added R script. Noting here for reference that the `cluster_sce()` function in that PR is currently an internal function (not being exported).

sjspielman:
Closes #192 

This PR adds (passing :tada:) tests for the new `cluster_sce()` function, including tests that output is as expected when function is correctly used, and tests that the function errors when it should.
Some context - these tests were written as paired programming session w/ @cbethell. During this, we discussed how writing tests can help you find areas where the code itself can be made more robust. As an example of how one could do that, we added another check to `cluster_sce` that the input SCE is indeed an SCE, and wrote a unit test accordingly.

I also took this time to change the argument `pca_name` -> `pc_name`, which was mentioned here: https://github.com/AlexsLemonade/scpcaTools/pull/191#pullrequestreview-1547979343. 

Let me know if there are other tests that would be good to add!

### DIFF
--- a/R/cluster_sce.R
+++ b/R/cluster_sce.R
@@ -29,6 +29,11 @@ cluster_sce <- function(sce,
   # Set the seed
   set.seed(seed)
 
+  # Check that provided sce is indeed an SCE
+  if (!is(sce, "SingleCellExperiment")) {
+    stop("Expected a `SingleCellExperiment` object for `sce` argument.")
+  }
+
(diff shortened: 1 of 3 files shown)


allyhawkins (approved):
LGTM. Thank you for doing this!

Issue #197, opened 2023-07-28
Add iLISI tests


sjspielman:
Following the addition of a function to calculate `iLISI` scores in #174, we should add associated unit tests.

sjspielman:
Just went to do this, but looks like @allyhawkins beat me to it while I was at chopshop! I'll go ahead and close this out now.

Issue #196, opened 2023-07-28
Add tests for `processing_pcs.R` functions


sjspielman:
We have a couple helper functions in https://github.com/AlexsLemonade/scpcaTools/blob/62ef2e9ab3893a5f1a1cd0e5f2ec4a53a0753cad/R/processing_pcs.R that are currently untested. This issue tracks adding (small!) unit tests for those functions - `filter_pcs()` and `downsample_pcs()`.

allyhawkins:
I already wrote those - https://github.com/AlexsLemonade/scpcaTools/blob/62ef2e9ab3893a5f1a1cd0e5f2ec4a53a0753cad/tests/testthat/test-integration_metrics.R#L22 and https://github.com/AlexsLemonade/scpcaTools/blob/62ef2e9ab3893a5f1a1cd0e5f2ec4a53a0753cad/tests/testthat/test-integration_metrics.R#L42

sjspielman:
Ah! I looked for a file `test-processing_pcs.R` and since it wasn't there I assumed there were no tests, but I was super wrong! I'll just go ahead and update the `filter_pcs()` test as part of #198 (except pending discussion, since I might not need to given what I wrote in the PR!).

PR #198, opened 2023-07-28
Add iLISI function

### LINKED ISSUE #174: Add function to calculate iLISI

Add a function for calculating `iLISI` similar to the script in the `sc-data-integration` repo: https://github.com/AlexsLemonade/sc-data-integration/blob/main/scripts/utils/calculate-LISI.R

sjspielman:
Closes #174 

This PR adds a function `calculate_ilisi()`. 

- The iLISI function as I wrote it is not general to any LISI metric, but specific to `iLISI` (e.g., not `cLISI`). This might be something to circle back to if we want to use the other metric for cell typing, but then again we won't usually have cell type "ground truths" in non-simulated data, so I thought the chances of needing `cLISI` were probably low enough to not consider it for now. Do we agree with this choice?
- (edit) I added `lisi` to the image.
- Along the way, I updated the `filter_pcs()` function, as discussed in #174, to only relabel PC rownames to batches if specified.
  - When writing the iLISI function, I therefore assumed that the PC rownames going _into_ the iLISI function at least contain barcodes, aka some kind of cell-specific ID. I did a little digging through our existing code to see how reasonable of an assumption this was, and I ended up discovered something else, as follows....
  - In the `integrate_sce.R` script, we actually have already enforced no `NA`s in batches:
     https://github.com/AlexsLemonade/scpcaTools/blob/62ef2e9ab3893a5f1a1cd0e5f2ec4a53a0753cad/R/integrate_sces.R#L54-L58
  - So, if we are ever calculating a metric from an integrated SCE (as opposed to a merged SCE), we can be reasonably sure that there are no `NA`s in batches! This might mean we can actually revert the `filter_pcs()` changes that we discussed + I made here, and just not even call it at all from the iLISI function. It probably doesn't hurt to keep the function call, but it also won't really have an effect. Curious to hear thoughts on this!
  - This also has the same implication for the ASW function - 
     https://github.com/AlexsLemonade/scpcaTools/blob/62ef2e9ab3893a5f1a1cd0e5f2ec4a53a0753cad/R/calculate_silhouette_width.R#L55-L56

Note also I opened a subsequent issue #197 for adding associated tests for iLISI.


sjspielman:
Noting that we'll want to update `filter_pcs()` tests, but I'm not going to do so until we chat a little first about whether we actually want to keep the `filter_pcs()` tests I made in this PR.
Note: https://github.com/AlexsLemonade/scpcaTools/issues/196#issuecomment-1655886248

R/calculate_ilisi.R
```diff
@@ -0,0 +1,61 @@
+#' Calculate iLISI (integration Local Inverse Simpson's Index) scores
+#'
+#'
+#' @param integrated_sce The integrated SCE object
```
allyhawkins, line 4: Given the changes that are in #191, let's use `merged_sce` throughout. These could be either unintegrated and merged objects or objects with integration. 
```suggestion
#' @param merged_sce The merged SCE object containing data from multiple batches
```

R/calculate_ilisi.R
```diff
@@ -0,0 +1,61 @@
...
+#' @param pc_name The name that allows access to the PCs. Example: fastMNN_PCA
+#' @param batch_column The variable in `integrated_sce` indicating the grouping of interest.
```
allyhawkins, line 6:
```suggestion
#' @param batch_column The variable in `merged_sce` indicating the grouping of interest.
```

R/calculate_ilisi.R
```diff
@@ -0,0 +1,61 @@
...
+#'  Generally this is either batches or cell types. Default is "library_id".
+#'
+#'
+#' @return Data frame with three columns, one row per cell. Columns are `ilisi_score`,
+#'   `cell_barcode`, and `batch_id`
+#'
+#' @import SingleCellExperiment
+#'
+#' @export
+calculate_ilisi <- function(integrated_sce,
```
allyhawkins, line 16:
```suggestion
calculate_ilisi <- function(merged_sce,
```

R/calculate_ilisi.R
```diff
@@ -0,0 +1,61 @@
...
+    stop("The provided `pc_name` cannot be found in the SingleCellExperiment object.")
+  }
+
+  # Check that `batch_column` is in colData of SCE
+  if (!batch_column %in% colnames(colData(integrated_sce))) {
+    stop("The specified batch column is missing from the colData of the SingleCellExperiment object.")
+  }
+
+  # Pull out the PCs or analogous reduction
+  pcs <- reducedDim(integrated_sce, pc_name)
```
allyhawkins, line 31:
```suggestion
  pcs <- reducedDim(merged_sce, pc_name)
```

allyhawkins (commented):
This looks good to me, my only suggestion is about the renaming to `merged_sce`. Also we should keep in the NA removal because we could use unintegrated objects. Also we had used that in data integration in the case of having missing cell type information and calculating the metrics for cell types. In case we do that for any of these metrics I think we should keep it. 

Also feel free to update the tests for `filter_pcs`.

sjspielman:
@allyhawkins thanks for the review, this should be ready for another look! I swapped all `integrated_sce` -> `merged_sce`, and I added an additional test for the new `filter_pcs()` argument.

R/processing_pcs.R
```diff
@@ -3,9 +3,11 @@
 #' @param pcs The PCs to be filtered and labeled
 #' @param batches Vector of cell-wise batch information whose length is the number of
 #'   rows in `pcs`
+#' @param rename_pcs Whether to rename the PC rownames with the cell's batch label.
+#'   Default is `TRUE`.
 #'
 #' @return PCs with NA batch cells removed and labeled rownames
-filter_pcs <- function(pcs, batches) {
+  filter_pcs <- function(pcs, batches, rename_pcs = TRUE) {
```
allyhawkins, line 10:
```suggestion
filter_pcs <- function(pcs, batches, rename_pcs = TRUE) {
```
I don't think we want the spaces here?

allyhawkins (approved):
This looks good to me. We should probably run the styler that we have set up before merging this. If you haven't seen it you just need to comment `/style` after going to https://github.com/orgs/AlexsLemonade/people and making sure that your membership is set to "public".

R/processing_pcs.R
sjspielman: no we do not, thanks!

sjspielman:
Not sure why this CI check is hanging. There's also nothing indicating "in progress" on the actions status page, so I'm going to just tweak a super minor thing just to have something to push and maybe trigger this action.

PR #199, opened 2023-07-31
Accept list of PCs in within-batch function

### LINKED ISSUE #193: Accept list of PC names in clustering and within-batch functions
We will now want to modify the `cluster_sce()` and `calculate_within_batch_ari()` functions to accept a list of PC names, rather than just a single PC name, for use in downstream analyses.

cbethell:
Closes #193 

This PR modifies the `calculate_within_batch_ari()` function to accept a list of PC names, rather than a single PC name for use in downstream analyses (i.e. a scenario where we want to compare `fastMNN_PCA` to `harmony_PCA`). 

Note that the issue #193 mentions modifying the `cluster_sce()` function as well, but with the modifications here that was unnecessary as we look through a list of PC names include the cluster function within the loop. Let me know if you still believe the cluster function should be modified here as well.


R/calculate_within_batch_ari.R
```diff
@@ -3,7 +3,7 @@
 #'
 #' @param individual_sce_list A named list of individual SCE objects. It is
 #'  assumed these have a reduced dimension slot with principal components named "PCA".
-#' @param pc_name The name that allows access to the PCs in the merged SCE
+#' @param pc_list A list of names that allow access to the PCs in the merged SCE
 #'  object. Example: "fastMNN_PCA".
```
allyhawkins, line 7:
```suggestion
#'  object. Example: c("PCA", "fastMNN_PCA").
```

R/calculate_within_batch_ari.R
```diff
@@ -48,54 +48,61 @@ calculate_within_batch_ari <- function(individual_sce_list,
...
-  # Cluster merged pcs only one time
-  merged_sce <- merged_sce |>
-    cluster_sce(
-      pca_name = pc_name,
-      BLUSPARAM = bluster::NNGraphParam(cluster.fun = "louvain", type = "jaccard"),
-      cluster_column_name = "merged_clusters"
-    )
+  for (pc_name in pc_list) {
+    # Pull out the PCs or analogous reduction from merged object
+    merged_pcs <- reducedDim(merged_sce, pc_name)
```
allyhawkins, line 60: So what I think we should do is take everything inside this `for` loop and put into a non-exported function that takes as input `merged_sce`, `pc_name`, and `batch_column`. 

Then you can easily change the `for` loop to a `purrr::map` call instead.

allyhawkins (commented):
I think my biggest comment is that with this addition I think we want to take a lot of the steps that are inside the `for` loop and create a new function. Then switch the `for` loop to `purrr::map`. Also we probably need to do the same thing for ASW, but that can be in a different PR.

R/calculate_within_batch_ari.R
cbethell: Thanks for that tip @allyhawkins! I refactored things as you suggested above -- let me know if the changes look reasonable to you!!

R/calculate_within_batch_ari.R
```diff
@@ -48,54 +48,20 @@ calculate_within_batch_ari <- function(individual_sce_list,
...
-      return(ari)
-    })
-
-  # Create tibble with ARI and batch id
-  within_batch_ari_tibble <- tibble::tibble(
-    ari = all_ari,
-    batch_id = batch_ids,
-    pc_name = pc_name
-  )
+    )
```
allyhawkins, line 64:
```suggestion
    ) |>
    dplyr::bind_rows()
```
The `purrr::map` will return a list of data frames, so you will need to combine those into a single dataframe.

R/processing_pcs.R
```diff
@@ -61,3 +61,68 @@ downsample_pcs <- function(pcs, frac_cells, min_cells = 50) {
 
   return(downsampled_pcs)
 }
+
+#' Calculate within-batch ARI using provided PCs for use in integration metric calculations
```
allyhawkins, line 65: Because this function is only being used inside the other within batch ari function, I would move this to be within the same script, just below the main function.

R/processing_pcs.R
```diff
@@ -61,3 +61,68 @@ downsample_pcs <- function(pcs, frac_cells, min_cells = 50) {
...
+#' @param batch_column The variable in `merged_sce` indicating the grouping of interest.
+#'  Generally this is either batches or cell types. Default is "library_id".
+#'
+#' @return Tibble with three columns: `ari`, representing the calculated ARI values;
+#'   `batch_id`, the batch id associated with each `ari`; `pc_name`, the name
+#'   associated with the pc results
+
+calculate_within_batch_ari_pcs <-
+  function(merged_sce, pc_name, batch_column = "library_id") {
+    # Pull out the PCs or analogous reduction from merged object
```
allyhawkins, line 79: For good practice, we should add the check for `pc_name` being in the `merged_sce` and same with the `batch_column`.

allyhawkins (commented):
Thanks for doing the reorganization here. I added a few comments, one of which is about moving the internal function into the same R script as the wrapper function for within batch ARI. Make sure you update the documentation after doing that.

cbethell:
Thanks for those refactoring suggestions @allyhawkins! I moved that internal function into the same script as the wrapped function that calls it, updated the documentation, and refactored the `purrr::map()` step. I believe this is ready for another look!

R/calculate_within_batch_ari.R
```diff
@@ -48,54 +48,92 @@ calculate_within_batch_ari <- function(individual_sce_list,
...
+#' @return Tibble with three columns: `ari`, representing the calculated ARI values;
+#'   `batch_id`, the batch id associated with each `ari`; `pc_name`, the name
+#'   associated with the pc results
+
+within_batch_ari_from_pcs <-
+  function(merged_sce, pc_name, batch_column = "library_id") {
+    # Check that `pc_name` is in merged SCE object
+    if (!pc_name %in% reducedDimNames(merged_sce)) {
+      stop("The provided `pc_name` cannot be found in the `merged_sce` object.")
+    }
```
allyhawkins, line 88: Add one more check here for `batch_column` in `colnames(merged_sce)`

allyhawkins (approved):
This looks good. Just one missing check and then make sure you remove the extra `.Rd` file for when you previously had them separate. And run the styler. Otherwise good to go from me.

R/calculate_within_batch_ari.R
```diff
@@ -3,8 +3,8 @@
 #'
 #' @param individual_sce_list A named list of individual SCE objects. It is
 #'  assumed these have a reduced dimension slot with principal components named "PCA".
-#' @param pc_name The name that allows access to the PCs in the merged SCE
-#'  object. Example: "fastMNN_PCA".
+#' @param pc_names A list of names that allow access to the PCs in the merged SCE
```
jashapiro, line 6:
```suggestion
#' @param pc_names A list of names of the PCs in the merged SCE `reducedDim` slot 
```

R/calculate_within_batch_ari.R
```diff
@@ -48,54 +48,96 @@ calculate_within_batch_ari <- function(individual_sce_list,
   }
 
   # Check that `pc_name` is in merged SCE object
-  if (!pc_name %in% reducedDimNames(merged_sce)) {
-    stop("The provided `pc_name` cannot be found in the `merged_sce`.")
+  if (!any(pc_names %in% reducedDimNames(merged_sce))) {
```
jashapiro, line 51: If we want to check that they are all present, we need "not all"
```suggestion
  if (!all(pc_names %in% reducedDimNames(merged_sce))) {
```

R/calculate_within_batch_ari.R
```diff
@@ -48,54 +48,96 @@ calculate_within_batch_ari <- function(individual_sce_list,
...
+#'
+#' @return Tibble with three columns: `ari`, representing the calculated ARI values;
+#'   `batch_id`, the batch id associated with each `ari`; `pc_name`, the name
+#'   associated with the pc results
+within_batch_ari_from_pcs <-
+  function(merged_sce, pc_name, batch_column = "library_id") {
+    # Check that `pc_name` is in merged SCE object
+    if (!pc_name %in% reducedDimNames(merged_sce)) {
+      stop("The provided `pc_name` cannot be found in the `merged_sce` object.")
+    }
```
jashapiro, line 87: Since this function is not exported, and this was already checked in the wrapper function, do we need the check here again?

R/calculate_within_batch_ari.R
```diff
@@ -48,54 +48,96 @@ calculate_within_batch_ari <- function(individual_sce_list,
...
+
+    # Check that `batch_column` is in colData of SCE
+    if (!batch_column %in% colnames(colData(merged_sce))) {
+      stop("The specified batch column is missing from the colData of the `merged_sce`.")
+    }
```
jashapiro, line 92: This check should probably be in the wrapper function instead? We only need to check this once, not for each PC.

R/calculate_within_batch_ari.R
```diff
@@ -48,54 +48,96 @@ calculate_within_batch_ari <- function(individual_sce_list,
...
+      cluster_sce(
+        pc_name = pc_name,
+        BLUSPARAM = bluster::NNGraphParam(cluster.fun = "louvain", type = "jaccard"),
+        cluster_column_name = "merged_clusters"
       )
 
-      return(ari)
-    })
+    merged_clusters <- merged_sce$merged_clusters |>
+      purrr::set_names(merged_sce[[batch_column]])
```
jashapiro, line 106: The clustering parameters (as `BLUSPARAM`) should probably be an option to the `within_batch_ari_from_pcs()` and `calculate_within_batch_ari()` functions. This will ensure that we are using the same ones always, but could change them more easily if needed.

R/calculate_within_batch_ari.R
```diff
@@ -48,54 +48,96 @@ calculate_within_batch_ari <- function(individual_sce_list,
...
+#' @param pc_name The name used to access the PCA results stored in the
+#'   SingleCellExperiment object
+#' @param batch_column The variable in `merged_sce` indicating the grouping of interest.
+#'  Generally this is either batches or cell types. Default is "library_id".
+#'
+#' @return Tibble with three columns: `ari`, representing the calculated ARI values;
+#'   `batch_id`, the batch id associated with each `ari`; `pc_name`, the name
+#'   associated with the pc results
+within_batch_ari_from_pcs <-
+  function(merged_sce, pc_name, batch_column = "library_id") {
```
jashapiro, line 83: Implementing passing in the clustering parameters.
```suggestion
  function(merged_sce, 
           pc_name,
           batch_column = "library_id",
           BLUSPARAM = bluster::NNGraphParam(cluster.fun = "louvain", type = "jaccard")
           ) {
```

R/calculate_within_batch_ari.R
```diff
@@ -19,7 +19,7 @@
 #'
 #' @export
 calculate_within_batch_ari <- function(individual_sce_list,
-                                       pc_name,
+                                       pc_names,
                                        merged_sce,
                                        batch_column = "library_id",
```
jashapiro, line 24: Allowing the clustering parameter to be set.
```suggestion
                                       batch_column = "library_id",
                                       BLUSPARAM = bluster::NNGraphParam(cluster.fun = "louvain", type = "jaccard")
```

R/calculate_within_batch_ari.R
```diff
@@ -48,54 +48,96 @@ calculate_within_batch_ari <- function(individual_sce_list,
...
+  # Calculate within-batch ARI values across list of PCs
+  within_batch_ari_tibble <-
+    purrr::map(
+      pc_names,
+      \(pcs)
+      within_batch_ari_from_pcs(
+        merged_sce = merged_sce,
+        batch_column = batch_column,
+        pc_name = pcs
+      )
```
jashapiro, line 64: Pass the `BLUSPARAM` to the internal function
```suggestion
        pc_name = pcs,
        BLUSPARAM = BLUSPARAM
      )
```

R/calculate_within_batch_ari.R
```diff
@@ -48,54 +48,96 @@ calculate_within_batch_ari <- function(individual_sce_list,
...
+          cluster_sce(
+            pc_name = "PCA",
+            BLUSPARAM = bluster::NNGraphParam(cluster.fun = "louvain", type = "jaccard"),
+            cluster_column_name = "individual_clusters"
+          )
 
-  return(within_batch_ari_tibble)
-}
+        # Extract clusters from merged clustering for batch
+        clusters_to_keep <- grep(batch, merged_sce[[batch_column]])
```
jashapiro, line 121: I don't think you need grep here? And you are keeping cells, right, not whole clusters?
```suggestion
        cells_to_keep <- which(merged_sce[[batch_column]] == batch)
```

jashapiro: Looking at this, I think I am actually a bit concerned that there is no guarantee that the order of the cells in these two SCE objects is the same. While we would not expect merging to alter the order, we don't guarantee that. 
Can we check that the barcodes all match between the individual clustering result and the batch clusters?

On line 106 you pull out the merged clusters, but I would move that down here to closer to where it will be used. But also there you set the names of the object to the batch names, which I'm not sure we need anymore? If not, maybe set it to the barcodes so we can check order?

R/calculate_within_batch_ari.R
jashapiro, line 106: Do we need to pull this out or can we just leave it? I don't think I understand why we are setting the names here.

jashapiro (commented):
The overall structure is good, but I suggested a couple things: One is making the `bluster` parameter set by a argument, in case we want to change it later (and to prevent recreating the same object for every clustering run).

The other thing is that we probably want to check that the cells are in the same order for the merged and individual batches. The way we merge, this is not likely to be an issue, but it seems worth checking where we can. The easiest would be to check that the barcodes all match, I think.

cbethell:
Thanks for the review @jashapiro! I implemented the changes you mentioned above and added the cell barcode check as you suggested. I've tested that things successfully work so I think this is ready for another look!

jashapiro (approved):
LGTM!

The only thing I might add is that we should probably allow passing in `BPPARAM` and passing that along to `cluster_sce` functions for potential speedups. We could do that either with an explicit `BPPARAM` argument, or by adding `...` as we have in `cluster_sce` already.

cbethell:
> LGTM!
> 
> The only thing I might add is that we should probably allow passing in `BPPARAM` and passing that along to `cluster_sce` functions for potential speedups. We could do that either with an explicit `BPPARAM` argument, or by adding `...` as we have in `cluster_sce` already.

Noting here that I added `...` to keep things consistent and flexible (in case there's another argument we may want to pass in addition to `BPPARAM`) -- and also updated the docs 🚀

PR #200, opened 2023-07-31
Accept list of PCs in silhouette width function

cbethell:
Related to #193 

This PR modifies the `calculate_silhouette_width()` function to accept a list of PC names, rather than a single PC name for use in downstream analyses (i.e. a scenario where we want to compare `fastMNN_PCA` to `harmony_PCA`).

In order to make this modification, a non-exported function named `calculate_silhouette_width_pcs()` was added to the `processing_pcs.R` script.


R/calculate_silhouette_width.R
```diff
@@ -44,36 +45,21 @@ calculate_silhouette_width <- function(merged_sce,
...
-    dplyr::bind_rows() |>
-    # Add integration method into tibble
-    dplyr::mutate(pc_name = pc_name)
+  # Calculate the silhouette width values across list of PCs
+  all_silhouette_df <- purrr::map_df(
+    pc_list,
+    ~ calculate_silhouette_width_pcs(
+      merged_sce = merged_sce,
+      batch_column = batch_column,
+      pc_name = .
```
allyhawkins, line 59:
```suggestion
    pc_list,
    \(pcs)
    silhouette_width_from_pcs(
      merged_sce = merged_sce,
      batch_column = batch_column,
      pc_name = pcs
```

R/processing_pcs.R
```diff
@@ -126,3 +126,47 @@ calculate_within_batch_ari_pcs <-
 
     return(within_batch_ari_tibble)
   }
+
+#' Calculate silhouette width scores from a merged SCE object.
```
allyhawkins, line 130: Same comment from #199 to move this into the same script where you have the wrapper function.

R/processing_pcs.R
```diff
@@ -126,3 +126,47 @@ calculate_within_batch_ari_pcs <-
...
+#' @param batch_column The variable in `merged_sce` indicating the grouping of interest.
+#'  Generally this is either batches or cell types. Default is "library_id".
+#'
+#' @return Tibble with five columns: `rep`, representing the given downsampling replicate;
+#'   `silhouette_width`, the calculated silhouette width for the given `rep`; `silhouette_cluster`,
+#'   the assigned cluster for the cell during silhouette calculation, i.e. the true identity;
+#'   `other_cluster`, the other assigned for the cell during silhouette calculation;
+#'   `pc_name`, the name associated with the pc results
+calculate_silhouette_width_pcs <- function(merged_sce, pc_name, batch_column = "library_id") {
+  # Pull out the PCs or analogous reduction
```
allyhawkins, line 146: Same comment as #199, add some checks here to be safe!

R/calculate_silhouette_width.R
```diff
@@ -44,36 +45,21 @@ calculate_silhouette_width <- function(merged_sce,
...
+    )
+  )
```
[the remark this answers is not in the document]
jashapiro: Just a quick note to say that this should be accompanied by changing `purrr::map_df()` to `purrr::map()` as the former is deprecated.

cbethell:
I propagated those changes that applied here from #199 and also updated argument naming in the test script 👍

R/calculate_silhouette_width.R
```diff
@@ -3,7 +3,8 @@
 #' This function performs replicated calculations on downsampled data.
 #'
 #' @param merged_sce The merged SCE object containing data from multiple batches
-#' @param pc_name The name that allows access to the PCs. Example: fastMNN_PCA
+#' @param pc_list A list of names that allow access to the PCs in the merged SCE
+#'  object. Example: c("PCA", "fastMNN_PCA").
```
sjspielman, line 7: This is not a list, but a vector! The argument name should better reflect the actual type. That said, `pc_vector` is not a very appealing argument name, so perhaps `pc_names`? This will require updating code in several places, and re-running document.

R/calculate_silhouette_width.R
```diff
@@ -30,8 +31,8 @@ calculate_silhouette_width <- function(merged_sce,
   set.seed(seed)
```
sjspielman, line 31: I am not sure that we still need to set a seed here, given that it is set fresh every time `silhouette_with_from_pcs` is called. I do think it is better to set the seed from that latter function, to ensure that seed is being used for all random sampling and not something later in the random number generator chain.
Anyways, since we set the seed later, this doesn't _hurt_ anything to stick around, but it also doesn't _do_ anything. So, it can go.

R/calculate_silhouette_width.R
```diff
@@ -44,36 +45,95 @@ calculate_silhouette_width <- function(merged_sce,
...
+#' in integration metric calculations
+#'
+#' @param merged_sce The merged SCE object containing data from multiple batches
+#' @param pc_name The name used to access the PCA results stored in the
+#'   SingleCellExperiment object
+#' @param batch_column The variable in `merged_sce` indicating the grouping of interest.
+#'  Generally this is either batches or cell types. Default is "library_id".
+#' @param frac_cells The fraction of cells to downsample to.
+#' @param nreps The number of times to repeat sub-sampling procedure.
+#' @param seed Seed for initializing random sampling
```
sjspielman, line 83:
```suggestion
#' @param seed Seed for initializing random sampling. Default is NULL.
```

sjspielman (commented):
Returning some initial review comments to start! After these changes are incorporated, I'll go ahead and test the code out a little more in depth.

R/calculate_silhouette_width.R
cbethell: I went with `pc_names`, updated the docs, and on my way to propagate this change to #199!

R/calculate_silhouette_width.R
```diff
@@ -44,36 +42,95 @@ calculate_silhouette_width <- function(merged_sce,
...
+        "The specified batch column is missing from the colData of the SingleCellExperiment object."
+      )
+    }
+
+    # Pull out the PCs or analogous reduction
+    pcs <- reducedDim(merged_sce, pc_name)
+
+    # Remove batch NAs from PCs and label rownames
+    labeled_pcs <-
+      filter_pcs(pcs, colData(merged_sce)[, batch_column])
```
jashapiro, line 114: Just for consistency with the other PR I just reviewed (#199), we should decide if we want to use `merged_sce[[batch_column]]` or `colData(merged_sce)[, batch_column]` or colData(merged_sce)[[batch_column]]` and try to be consistent about it. I think I like the last one best, even if it is a bit more verbose.

R/calculate_silhouette_width.R
```diff
@@ -44,36 +42,95 @@ calculate_silhouette_width <- function(merged_sce,
...
+#'   `silhouette_width`, the calculated silhouette width for the given `rep`; `silhouette_cluster`,
+#'   the assigned cluster for the cell during silhouette calculation, i.e. the true identity;
+#'   `other_cluster`, the other assigned for the cell during silhouette calculation;
+#'   `pc_name`, the name associated with the pc results
+silhouette_width_from_pcs <-
+  function(merged_sce,
+           pc_name,
+           batch_column = "library_id",
+           frac_cells,
+           nreps,
```
jashapiro, line 92: Pretty minor, but it is generally good practice to put required parameters first, with the ones that have default values following.

Alternatively, you could leave the order as is, but add defaults to `frac_cells` and `nreps` as above.

jashapiro (approved):
This looks good to me with only a few minor comments.

sjspielman (approved):
LGTM, thanks for making the changes here!
Just want to note that this PR won't run tests automatically since base isn't `main`, but tests run fine locally on my end (and they'll get run in the other PR branch anyways) 👍

PR #203, opened 2023-08-08
Fix within-batch ARI function

cbethell:
Upon an attempt to implement the `calculate_within_batch_ari()` function in downstream analyses, an error regarding the `batch_ids` object not being found was encountered.

This is because we currently refer to batch ids in the [outer/wrapper function](https://github.com/AlexsLemonade/scpcaTools/blob/47644e14efbb9d37db37c8e08324f53dc769763e/R/calculate_within_batch_ari.R#L41) without passing them to the internal function [where we use them](https://github.com/AlexsLemonade/scpcaTools/blob/47644e14efbb9d37db37c8e08324f53dc769763e/R/calculate_within_batch_ari.R#L113).

This PR fixes that error by moving the `batch_ids` definition to the internal function. The documentation has also been updated to reflect the change.

### DIFF
--- a/R/calculate_within_batch_ari.R
+++ b/R/calculate_within_batch_ari.R
@@ -37,22 +37,6 @@ calculate_within_batch_ari <- function(individual_sce_list,
     stop("The specified batch column is missing from the colData of the `merged_sce`.")
   }
 
-  # Check that list of SCE objects is named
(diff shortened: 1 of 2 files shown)


allyhawkins (approved):
As long as this passes checks you are good to go.

PR #204, opened 2023-08-09
Integration metrics test functions

### LINKED ISSUE #194: Add test functions for calculating within-batch ARI
Now that we are adding a function to calculate within-batch ARI in PR https://github.com/AlexsLemonade/scpcaTools/pull/191, we will also want to add relevant test functions.

allyhawkins:
Closes #194 and #197 

This PR adds test functions for both within batch ARI and the iLisi functions we recently added. In doing this I noticed that we had hard coded the column name that contains the original cell barcode in the `merged_sce`. So I went ahead and removed that hard coding and added a `barcode_column` argument. 

Are there any tests that I missed or should be checking?

### DIFF
--- a/R/calculate_within_batch_ari.R
+++ b/R/calculate_within_batch_ari.R
@@ -10,7 +10,7 @@
 #' @param batch_column The variable in `merged_sce` indicating the grouping of interest.
 #'  Generally this is either batches or cell types. Default is "library_id".
 #' @param barcode_column The variable in `merged_sce` indicating the original cell barcode.
-#'  Default is "barcode".
+#'  Default is "cell_id".
 #' @param BLUSPARAM A BlusterParam object specifying the clustering algorithm to
 #'   use and any additional clustering options. Default is
 #'   `bluster::NNGraphParam(cluster.fun = "louvain", type = "jaccard")`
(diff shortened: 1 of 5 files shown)


R/calculate_within_batch_ari.R
```diff
@@ -26,6 +29,7 @@ calculate_within_batch_ari <- function(individual_sce_list,
                                        pc_names,
                                        merged_sce,
                                        batch_column = "library_id",
+                                       barcode_column = "barcode",
```
cbethell, line 32: Will we mostly want to use `barcode` here instead of `cell_id`?

My reason for asking is that I checked the `barcode` column of our example integrated SCE object and it seems to be filled with NAs:

Screen Shot 2023-08-09 at 11 02 05 AM

Whereas, `cell_id` gives us:

Screen Shot 2023-08-09 at 11 03 19 AM

allyhawkins: Good catch! We have an argument in `merge_sce_list()` that sets the default cell column to be `cell_id`. I went ahead and changed the argument name and default to match what we already use there: 
https://github.com/AlexsLemonade/scpcaTools/blob/7982096108d031510eb925d1d6d600caf91a4c7f/R/merge_sce_list.R#L59

cbethell (approved):
Looks like all necessary tests are covered here so LGTM! Thanks for tackling this @allyhawkins!

PR #206, opened 2023-08-11
Fix iLISI calculation and test

### LINKED ISSUE #205: Revisit iLISI calculation and tests
While looking through some iLISI code that went in for #197 when I was at chopshop, I noticed something is off, so opening this issue to address things!

- First, the test includes this logic, which is probably _meant_ to test for the inclusive range `[0,1]`, but actually just tests that numbers are positive since an "or" is used. We probably want to change the logical statement here to something like  `dplyr::between(ilisi$ilisi_scores, 0, 1)`
https://github.com/AlexsLemonade/scpcaTools/blob/39dbed38b3b906b132783f45b3e3a3a7a85bc729/tests/testthat/test-integration_metrics.R#L253-L255

- Second, as a consequence of that test being a little off, the test passes but it actually _should have failed_. These are the ranges for the ilisi scores going into the test:
```
> summary(ilisi$ilisi_score)
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
  7.299  12.039  13.555  13.452  15.082  19.901 
```

So what is happening?! I went back to the `sc-data-integration` repo, and realized that we did some more post-processing hiding in a plotting function!! So, that post-processing didn't make it into the iLISI function in _this_ repository. 

https://github.com/AlexsLemonade/sc-data-integration/blob/140ce34ecf15e2af2b3c7d7efdc1a32435859eae/analysis_templates/01-single-group-integration-check-template.Rmd#L460
```
The range of iLISI scores will be from 1 to the number of batches, however we have normalized the score following the same method utilized by the `scIB` benchmarking tools: (ilisi - 1)/(num batches -1)
```

https://github.com/AlexsLemonade/sc-data-integration/blob/140ce34ecf15e2af2b3c7d7efdc1a32435859eae/scripts/utils/plotting-functions.R#L475-L486

```
  if (lisi_type == "iLISI") {
    # normalize following the scIB method
    # https://github.com/theislab/scib/blob/067eb1aee7044f5ce0652fa363ec8deab0e9668d/scib/metrics/lisi.py#L98-L100
    lisi_df_updated <- lisi_df %>%
      dplyr::mutate(lisi_score_norm = (lisi_score-1)/(num_batches - 1))
  } else {
    # normalize following the scIB method
    # https://github.com/theislab/scib/blob/067eb1aee7044f5ce0652fa363ec8deab0e9668d/scib/metrics/lisi.py#L157-L159    
    lisi_df_updated <- lisi_df %>%
      dplyr::mutate(lisi_score_norm = (num_batches - lisi_score)/(num_batches - 1))
  }
```

### Suggested next steps

1. Add `dplyr::mutate(lisi_score_norm = (lisi_score-1)/(num_batches - 1))` to the `calculate_ilisi` function
2. Use `dplyr::between()` as I suggested above for the test

I'm going to go ahead and do this today, FYI @allyhawkins !

sjspielman:
Closes #205 

This PR fixes iLISI as explained in the linked issue.

One choice I made here is to keep _both_ the raw ilisi score and the normalized version, saved in `ilisi_score_norm`. Then in the tests, I test that `ilisi_score_norm` is in the correct range. So, a question is do we actually foresee needing to keep the raw `ilisi_score` itself? It can be back-calculated from the normalized score if we wanted, so technically we don't _need_ both columns. What do reviewers think?

### DIFF
--- a/R/calculate_ilisi.R
+++ b/R/calculate_ilisi.R
@@ -8,7 +8,8 @@
 #'
 #'
 #' @return Data frame with three columns, one row per cell. Columns are `ilisi_score`,
-#'   `cell_barcode`, and `batch_id`
+#'  `ilisi_score_norm`, `cell_barcode`, and `batch_id`. The column `ilisi_score_norm`
+#'  is a normalized version if `ilisi_score` where values range from [0,1], inclusive
 #'
 #' @import SingleCellExperiment
 #'
@@ -44,6 +45,9 @@ calculate_ilisi <- function(merged_sce,
   # Create data frame for input to lisi calculation
   batch_df <- data.frame(batch = rownames(labeled_pcs))
 
+  # calculate number of unique batches for later normalization
+  num_batches <- length(unique(batch_df$batch))
+
   # Calculate iLISI score
   ilisi_df <- lisi::compute_lisi(
     labeled_pcs,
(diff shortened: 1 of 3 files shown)


allyhawkins (approved):
LGTM, thanks so much for catching this!

PR #207, opened 2023-08-11
Update ARI function and ARI test

allyhawkins:
Inspired by #205, I decided to check up on the ARI tests and also change to using `dplyr::between()` for that. In doing that I realized we were returning a list for each ari in the dataframe, when we want each entry to be its own value. I changed the function to use `purrr::map_dbl` instead of `purrr::map`. 

@cbethell once this is fixed, you may want to make sure the hash is correct for downstream analyses and that nothing is broken in the report.


cbethell (approved):
LGTM!

Issue #208, opened 2023-08-11
Function for adding sample metadata to sce object


allyhawkins:
So the idea would be we would 
1. Make a `add_sample_metadata()` function that would add a `sample_metadata` data frame to `metadata(sce)`
2. `merge_sce_list()` would account for the possibility of `metadata(sce)$sample_metadata` tables in the `sce` objects to merge, and combine those in the merged sce metadata.
3. We could also add a function to copy data from the `sample_metadata` to `colData` by sample id, which might be useful.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/issues/403#issuecomment-1674982343_

PR #213, opened 2023-08-15
Add and combine sample metadata

allyhawkins:
Partially addresses #208 

This PR adds a new function to add a sample metadata data frame to an existing SCE object. It's pretty simple and takes an SCE and a dataframe as input and then adds the data frame to the metadata with the name `sample_metadata`. Is this what you had in mind @jashapiro? I feel like I might be missing something. 

The other thing I did here was account for sample metadata being present in the object metadata when merging objects. When the objects merge we get duplicates of the same named item in the metadata list. So after merging an object there are multiple `sample_metadata` slots filled with 1 row data frames. I decided that if there are `sample_metadata` slots in the metadata list, then we just combine them together into a single data frame and replace the existing metadata with the modified metadata containing the combined data frame. 

When working on this, I also noticed a lot of duplicate information in our merged object metadata and no indication of what information belonged to which original object. To account for that, I combined relevant library and processing information into a data frame that is stored in `metadata(sce)$library_metadata` in the `prepare_sce_for_merge()` function. This allows us to mimic how we deal with sample metadata and combine it into a single data frame with information about all the libraries stored in the merged object. 

I did add an argument that sets the default columns that should be included in that library metadata. I excluded any adt information, miQC results, and then the HVGs. One thought I had was to name the rest of the metadata with a library indicator, so instead of `highly_variable_genes` it becomes `<library_id>_highly_variable_genes`.


R/add_sample_metadata.R
```diff
@@ -0,0 +1,28 @@
...
+#'
+#' @param sce SingleCellExperiment object to add metadata too
+#' @param metadata_df A dataframe with metadata to add to the sce.
+#'
+#' @return SingleCellExperiment object with a `metadata_df` added to the `sample_metadata`
+#'   slot of the metadata list
+#' @export
+#'
+add_sample_metadata <- function(sce,
+                                metadata_df){
```
jashapiro, line 11: I think I was expecting this function to also check that the samples are present in the SCE, if we know that. I think we do with `metadata(sce)$sample_id`?

R/merge_sce_list.R
```diff
@@ -105,19 +131,59 @@ merge_sce_list <- function(sce_list = list(),
...
+  metadata_list <- c(
+    library_metadata = list(all_library_metadata),
+    metadata_to_keep
+  )
+
+  # if object has sample metadata then combine into a single data frame
+  if("sample_metadata" %in% names(metadata_list)){
+    all_sample_metadata <- metadata_list[names(metadata_list) == "sample_metadata"] |>
+      purrr::map(as.data.frame) |>
+      dplyr::bind_rows()
```
jashapiro, line 177: Do we want to add `unique()` here in case we are combining multiple runs of the same sample?

R/merge_sce_list.R
```diff
@@ -105,19 +131,59 @@ merge_sce_list <- function(sce_list = list(),
...
+
+    metadata_list <- c(
+      metadata_list,
+      sample_metadata = list(all_sample_metadata)
+    )
```
jashapiro, line 182: I'm not sure why you are converting to a list here?

Could we just do:
```
metadata_list[[sample_metadata]] <- all_sample_metadata
```

allyhawkins:
Thanks for taking a look at this! I added a new argument to the `add_sample_metadata()` function for the `sample_id_column`. This is the column name that contains the sample ids in the `metadata_df`. I then check that the contents of that match the `metadata(sce)$sample_id` contents. This should be ready for another look.

R/merge_sce_list.R
```diff
@@ -190,6 +259,23 @@ prepare_sce_for_merge <- function(sce,
   # Add `sce_name` to colnames so cell ids can be mapped to originating SCE
   colnames(sce) <- glue::glue("{sce_name}-{colnames(sce)}")
 
+  # merge library metadata
+  metadata_list <- metadata(sce)
+
+  # create a data frame with all the library specific information
+  library_metadata <- metadata_list[names(metadata_list) %in% combine_metadata_cols] |>
```
jashapiro, line 266: After a bit of thought, while I think the idea of combining library metadata into a data frame like for samples is a noble one, I worry that it will be brittle. I think a simpler path might be to just combine _all_ of the metadata into a `library_metadata` list of lists. So much of this might reduce down to something like:

```
library_metadata <- list(metadata(sce)$library_id = metadata(sce)[names(metadata(sce)) != "sample_metadata"]
sample_metadata <- metadata(sce)[["sample_metadata"]]

metadata_list <- c(
    library_id = metadata(sce)[["library_id"]]
    sample_id =  metadata(sce)[["sample_id"]]
    library_metadata = library_metadata,
    sample_metadata = sample_metadata
  )

```

The main thing I am thinking here is that we would not have to maintain a list of `combine_metadata_cols` or deal with collapsing values where there might have been a list. We would, however, maintain the `library_id` and `sample_id` values, which should be combined into vectors (followed by `unique()`). If there are other values that should be maintained at the top level, we could do that as well. But I think maintaining a list of things we want to retain at that top level makes more sense than a list of things we want to combine.



One potential problem I see here is if there is dealing with the case of combining two already merged sce objects, which is probably something we want to account for (and add as a test case?). 

Also, there is probably other metadata that is specific to multi-sample libraries that we would want to maintain if merging merged libraries... I'm not quite sure what we would do there, but maybe dropping anything like that and recreating it would make sense anyway? (I'm thinking things like any integration algorithm data, which I assume we are storing).

R/add_sample_metadata.R
```diff
@@ -0,0 +1,45 @@
...
+#' @param sample_id_column Name of column in `metadata_df` that contains the sample ids.
+#'   The contents of this column should be equivalent to the `sample_id` in `metadata(sce)`
+#'
+#' @return SingleCellExperiment object with a `metadata_df` added to the `sample_metadata`
+#'   slot of the metadata list
+#' @export
+#'
+add_sample_metadata <- function(sce,
+                                metadata_df,
+                                sample_id_column){
```
jashapiro, line 14: I made this comment elsewhere, but maybe we should just require that there be a column be named `sample_id`? I think it is just as easy for the user to make that change outside the function, and maybe more sustainable for downstream applications where we might want those columns to match.

jashapiro (commented):
Thanks for tackling this, and in particular noticing the problem of combining metadata. I proposed below a different solution for the combination though: I worry that data frames will be a bit too brittle, so I think we should probably just combine most of the metadata as a named list, which will allow us to retain all of the metadata for each library (basically, what you originally were doing for sample metadata, I think?). I would pull the `sample_metadata` out to have only one data frame there, and maybe a few other metadata items (those specific to merged libraries, like integration methods)

So the idea would be you would be able to access everything with calls like: 
```
my_library ="SCPCL000000"
metadata(merged_sce)$library_metadata[[my_library]]$filtering_method
```

A bit ugly, but not really worse than 
`metadata(merged_sce)$library_metadata[library_id == my_library, "filtering_method"]`

What do you think?

R/merge_sce_list.R
allyhawkins: So part of the reason I didn't do this earlier is because the SCE objects don't seem to want to store lists of lists? So I have the following code in the function for preparing the individual objects prior to merging: 
```
# separate library and sample metadata
  metadata_list <- metadata(sce)
  library_metadata <- metadata_list[names(metadata_list) != "sample_metadata"]
  sample_metadata <- metadata_list$sample_metadata

  # combine into one list
  metadata_list <- c(
    library_id = metadata(sce)[["library_id"]],
    sample_id = metadata(sce)[["sample_id"]],
    library_metadata = library_metadata,
    sample_metadata = sample_metadata
  )
```
But then that gives a list of 26 elements instead of a list of 4. And then `names(metadata_list` gives: 

```
[1] "library_id"                               "sample_id"                                "library_metadata.library_id"             
 [4] "library_metadata.sample_id"               "library_metadata.salmon_version"          "library_metadata.reference_index"        
 [7] "library_metadata.total_reads"             "library_metadata.mapped_reads"            "library_metadata.mapping_tool"           
[10] "library_metadata.alevinfry_version"       "library_metadata.af_permit_type"          "library_metadata.af_resolution"          
[13] "library_metadata.usa_mode"                "library_metadata.af_num_cells"            "library_metadata.tech_version"           
[16] "library_metadata.transcript_type"         "library_metadata.include_unspliced"       "library_metadata.filtering_method"       
[19] "library_metadata.miQC_model"              "library_metadata.prob_compromised_cutoff" "library_metadata.scpca_filter_method"    
[22] "library_metadata.min_gene_cutoff"         "library_metadata.adt_scpca_filter_method" "library_metadata.normalization"          
[25] "library_metadata.adt_normalization"       "library_metadata.highly_variable_genes"  
```

So then if I combine objects, I get the same problem that I was trying to avoid which is a bunch of metadata entries in the list that are named the same way.

allyhawkins: You can ignore that comment... I just need to convert the `library_metadata` and `sample_metadata` vectors to lists before adding them.

jashapiro: My bad for saying list and using `c()`! Should be:
```
metadata_list <- list(
    library_id = metadata(sce)[["library_id"]]
    sample_id =  metadata(sce)[["sample_id"]]
    library_metadata = library_metadata,
    sample_metadata = sample_metadata
  )
```

allyhawkins:
Thanks for the help with this @jashapiro. I went ahead and converted the library metadata to a list that then becomes a list of library metadata lists in the merged object. I also made the changes to the `add_sample_metadata` function to specify that the `metadata_df` should have a `sample_id` column. 

Right now, I didn't add any tests for the merged object other than making sure the right metadata names are present. If you have ideas for more tests I can add them once we solidify how the function will actually work.

R/merge_sce_list.R
```diff
@@ -118,6 +129,41 @@ merge_sce_list <- function(sce_list = list(),
   # Create the merged SCE from the processed list ------------------
   merged_sce <- do.call(cbind, sce_list)
 
+  # combine library metadata into a single data frame
+  metadata_list <- metadata(merged_sce)
+
+  # create a single list of library metadata
+  library_metadata <- metadata_list[names(metadata_list) == "library_metadata"]
```
jashapiro, line 136: Can we do these operations based on the original `sce_list`? It somehow feels safer to me, because still I don't think I understand how the metadata works after `cbind()`...

So my thought would be that you would do something like this for the libraries:

```suggestion
# create a single list of library metadata lists
library_metadata  <- sce_list |> purrr::map("library_metadata")
```
(this is a really nice shorthand for pulling a single element out of each of a list of lists).

R/merge_sce_list.R
```diff
@@ -118,6 +129,41 @@ merge_sce_list <- function(sce_list = list(),
...
+
+  # get a single list of all library ids
+  library_ids <- metadata(merged_sce)[names(metadata_list) == "library_id"] |>
+    unlist() |>
+    unique()
+
+  # get a single list of all sample ids
+  sample_ids <- metadata(merged_sce)[names(metadata_list) == "sample_id"] |>
+    unlist() |>
+    unique()
```
jashapiro, line 146:
```suggestion
  # get a vector of all library ids
  library_ids <- sce_list |> 
    purrr::map("library_id") |>
    unlist() |>
    unique()

  # get a vector of all sample ids
  sample_ids <- sce_list |> 
    purrr::map("sample_id") |>
    unlist() |>
    unique() 
```

R/merge_sce_list.R
```diff
@@ -118,6 +129,41 @@ merge_sce_list <- function(sce_list = list(),
...
+
+  # combine into one metadata list
+  metadata_list <- list(
+    library_id = library_ids,
+    sample_id = sample_ids,
+    library_metadata = library_metadata
+  )
+
+  # if object has sample metadata then combine into a single data frame
+  if("sample_metadata" %in% names(metadata_list)){
```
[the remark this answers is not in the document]
jashapiro: Also see comment about doing this from the original `sce_list` data?

jashapiro (commented):
Okay, getting there! I am still a bit worried about how `cbind()` handles metadata in ways that make me uncomfortable, so I had a suggestion to do things a bit differently there. 

I'm thinking that at this point we simply don't support merging merged sces, but we might want to make that explicit somehow. Maybe testing if there is a `library_metadata` slot in `metadata(sce)` and returning an error if so?

As for tests, I think we can check that all of the library_ids we expect are in `metadata(result_sce)$library_id` for example, and that `metadata(result_sce)$library_metadata` is a list with the same names? And that `metadata(result_sce)$sample_id` and `$sample_metadata` match?

R/merge_sce_list.R
allyhawkins: I just want to confirm that what you mean is below. We want to pull from the `metadata_list` and not the `sce_list`? 
```
  library_ids <- metadata_list |> 
    purrr::map("library_id") |>
    unlist() |>
    unique()
```

R/merge_sce_list.R
allyhawkins: Just to try and add some clarity into how the metadata looks after `cbind()`: 

```
$library_id
[1] "SCPCL000050"

$sample_id
[1] "SCPCS000053"

$salmon_version
[1] "1.9.0"

$reference_index
[1] "Homo_sapiens.GRCh38.104.spliced_intron.txome"
....

$library_id
[1] "SCPCL000051"

$sample_id
[1] "SCPCS000054"

$salmon_version
[1] "1.9.0"

$reference_index
[1] "Homo_sapiens.GRCh38.104.spliced_intron.txome"
...
```

So the lists are just concatenated together. Which is why I was pulling out anything with `library_metadata` as the name. Once we combine, we'll have a list named `library_metadata` for each object that was included in the merge.

R/merge_sce_list.R
jashapiro: As I was writing what I really wanted, I just realized we can do most of this in a single step with `purrr::transpose()` to turn the list of lists "inside out". At least I think so.

```
# get a list of just metadata and invert it so each library is an element within the metadata 
# components from `prepare_sce_for_merge()`
metadata_list <- sce_list |>
  purrr::map(metadata) |>
  purrr::transpose()

# flatten and reduce the library ids and sample ids
metadata_list$library_id <- metadata_list$library_id |>
  unlist() |>
  unique()
metadata_list$sample_id<- metadata_list$sample_id |>
  unlist() |>
  unique()

# row bind the sample metadata (I didn't add the if statement here)
metadata_list$sample_metadata <- metadata_list$sample_metadata |>
  dplyr::bind_rows() |>
  unique()
```
Does that make sense?

Then you can remove lines 150 - 155 as they are already combined into a list

allyhawkins: Yes this makes sense! Thank you for this, I think this really cleans it up! I implemented this in https://github.com/AlexsLemonade/scpcaTools/pull/213/commits/6024762765bfc2c96a28c189f2bee7b2a38d132b.

allyhawkins:
Okay I think we are getting there. I added a few more checks and simplified the metadata combination based on the suggestion to go right from the `sce_list`. This should be ready for another look.

tests/testthat/test-merge_sce_list.R
```diff
@@ -20,7 +28,7 @@ set.seed(1665)
 total_cells <- 24 # divisible by 3
 total_genes <- 12 # number of months intentionally.
 sce <- add_sce_data(
-  sim_sce(n_cells = total_cells, n_genes = total_genes, n_empty = 0)
+  scpcaTools:::sim_sce(n_cells = total_cells, n_genes = total_genes, n_empty = 0)
```
jashapiro, line 31: The `scpcaTools:::` should not be needed within the test suite, is it? (You might need it for running the code in the console, but not for running via testing tools)

jashapiro (commented):
I think this looks good overall, but it still needs tests for the metadata merging. 

In `test_that("merging SCEs with matching genes works as expected")` (or an extra test set) we should be checking that the metadata has the format that we would expect with the changes that we have here: There should still be the 4 metadata objects, some containing a list of libraries, etc.

Add those tests, and I think we should be all set.

tests/testthat/test-merge_sce_list.R
allyhawkins: No sorry that's just leftover from testing on my own.

allyhawkins:
> In test_that("merging SCEs with matching genes works as expected") (or an extra test set) we should be checking that the metadata has the format that we would expect with the changes that we have here: There should still be the 4 metadata objects, some containing a list of libraries, etc.

Added in https://github.com/AlexsLemonade/scpcaTools/pull/213/commits/4385f7040197fabb7bae31068043e779e099452c

tests/testthat/test-merge_sce_list.R
```diff
@@ -172,6 +201,32 @@ test_that("merging SCEs with matching genes works as expected", {
...
+  # library metadata should contain a list of library metadata
+  expect_true(
+    length(metadata(merged_sce)$library_metadata) == 3
+  )
+
+  # check that contents of library id and sample id are correct
+  # since they all have the same library/sample id there should just be 1 rather than the list
+  expect_true(
+    metadata(merged_sce)$library_id == "library1"
+  )
```
jashapiro, line 224: I think we may want to test merging with different libraries here to be sure we are doing it right for that (more common) case as well. And then you can use `expect_setequal()` to check that they are the right set of libraries.

tests/testthat/test-merge_sce_list.R
```diff
@@ -172,6 +201,32 @@ test_that("merging SCEs with matching genes works as expected", {
...
+
+  expect_true(
+    metadata(merged_sce)$sample_id == "sample1"
+  )
```
jashapiro, line 228:
```suggestion
  expect_equal(
    metadata(merged_sce)$sample_id, "sample1"
  )
  
  # Test that all the samples are in the sample metadata and vice versa
  expect_setequal(
    metadata(merged_sce)$sample_id, 
    metadata(merged_sce)$sample_metadata$sample_id
  )
```

jashapiro (commented):
Almost there... I updated a few tests to use the more specific `testthat` functions, as the errors you get from them are often a bit more helpful than the ones from `expect_true()` I only did this for one set of tests, so the other still needs updating. 

I also think we should probably update the test merge to include more than one library id and maybe more than one sample (though some duplicates might be good) for better testing.

allyhawkins:
> I also think we should probably update the test merge to include more than one library id and maybe more than one sample (though some duplicates might be good) for better testing.

I went ahead and added a batch id to the library/sample id being stored in the metadata to account for this. 

> I updated a few tests to use the more specific testthat functions, as the errors you get from them are often a bit more helpful than the ones from expect_true() I only did this for one set of tests, so the other still needs updating.

I removed using `expect_true()` and I believe I got all of them.

tests/testthat/test-merge_sce_list.R
```diff
@@ -31,10 +43,10 @@ preserve_rowdata_cols <- "gene_names"
 expected_coldata_cols <- sort(c("sum", "detected", batch_column, cell_id_column))
 
 # Generate some shared data for testing `merge_sce_list()` ------
-sce1 <- sim_sce(n_cells = total_cells / 3, n_genes = total_genes, n_empty = 0)
-sce2 <- sim_sce(n_cells = total_cells / 3, n_genes = total_genes, n_empty = 0)
-sce3 <- sim_sce(n_cells = total_cells / 3, n_genes = total_genes, n_empty = 0)
-sce_list <- purrr::map(
+sce1 <- scpcaTools:::sim_sce(n_cells = total_cells / 3, n_genes = total_genes, n_empty = 0)
+sce2 <- scpcaTools:::sim_sce(n_cells = total_cells / 3, n_genes = total_genes, n_empty = 0)
+sce3 <- scpcaTools:::sim_sce(n_cells = total_cells / 3, n_genes = total_genes, n_empty = 0)
```
jashapiro, line 48: I think what I meant was to change the library ids of these to be different... `library-1`, `library-2`, etc. Or just `sce1`, `sce2`, etc. which would be more in keeping with how sces are likely to be named.

Then something like the following below for the test.
```
  expect_setequal(
    metadata(result_sce)$library_id,
    c("sce1", "sce2", "sce3")
  ) 
```

allyhawkins: So I did that... In the `add_sce_data()` at the beginning of the function, I added a `batch` argument. That gets used to create the unique `library_id` and `sample_id`. So when that function is run on the `sce_list`, I pass the name of the sce object to the `batch` option. 

https://github.com/AlexsLemonade/scpcaTools/blob/292aa8a4bb0f3baaf9877e1f8178f058c055b0fb/tests/testthat/test-merge_sce_list.R#L13-L21

jashapiro: Ah, I see that now. I was just also expecting an explicit test for the expected `library_ids`, in _addition_ to checking that it matched the `sample_metadata` table. (I was seeing the test in the prepare function which only (correctly) expected a single value).


One more note while we are in there: you could update the following chunks to use `expect_setequal` instead of sorting... https://github.com/AlexsLemonade/scpcaTools/blob/292aa8a4bb0f3baaf9877e1f8178f058c055b0fb/tests/testthat/test-merge_sce_list.R#L178-L193

allyhawkins: Okay. I added a check for the contents of the `library_id` containing exactly the ids we want and then converted those and any others I saw to `expect_setequal`.

jashapiro (approved):
LGTM

PR #216, opened 2023-08-17
Update to R 4.3 & Bioconductor 3.17

### LINKED ISSUE #214: Update to R 4.3 and Bioconductor 3.17
From @jashapiro in slack: 
> Speaking of bluster: I just saw this when looking at the latest release notes:

```
Corrected calculation of the ratio matrix in pairwiseRand()
	when adjusted=TRUE, so that ratios have an upper bound of 1
```

We have seen this issue previously, which is why we had been using `pdfCluster::adj.rand.index()` in most cases. However here we use `pariwiseRand` to calculate within batch ARI. We do have a test implemented to make sure that ARI is in the correct range right now, but ultimately we should make this update.

jashapiro:
Closes #214 
Here I updated the `renv` and requirements for scpcaTools to use R 4.3 and Bioconductor 3.17. That is the bulk of the changes here, but there were also a number of smaller changes that I made to quiet test warnings and messages.

I am filing this as a draft for now because I have not yet tested it on the server. Doing so would require updating R on the server since we can only have one version there at a time (conda environments notwithstanding). If the Docker image builds properly, we can take that as some confirmation that things are working, but I did not want to make major assumptions on that front. 

Reviewers should checkout and test the `renv` installation and that the package builds and tests properly on their own machine, using R 4.3. One niggle is that on my machine the `anndata` tests were failing repeatedly if I triggered them through RStudio until I ran `basilisk.utils::installConda()` outside of the test environment. If you run into that problem, maybe give that a try?

You may also see in testing some messages about `rgdal` and such. These are related to dependencies of a `Seurat` dependency that are being removed from CRAN, but the packages that are warned about are not installed or used so it is just an annoying warning, and I couldn't find an easy way to quiet them, so we just deal for now. 

I also bumped the version number here. I think the fact that we are changing the required versions of R & Bioconductor will warrant a full top number bump, so I went for it, partly just as a marker.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM rocker/r-ver:4.2.3
+FROM rocker/r-ver:4.3.1
 LABEL maintainer="ccdl@alexslemonade.org"
 LABEL org.opencontainers.image.source https://github.com/AlexsLemonade/scpcaTools
 
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -1,10 +1,20 @@
-^requirements\.txt$
+# Github and docs
+^\.github$
+^CONTRIBUTING\.md$
+
+# user settings
+^.*\.Rproj$
+^\.Rproj\.user$
+
+# renv files
 ^renv$
 ^renv\.lock$
 ^dependencies\.R$
-^.*\.Rproj$
-^\.Rproj\.user$
-^\.github$
-^README\.Rmd$
+^requirements\.txt$
+^\.venv$
+
+# rendered notebooks
 ^inst/rmd/.*\.html$
+
+# Docker folder
 ^docker$
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -16,7 +16,7 @@ env:
 
 jobs:
   build_and_push:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     permissions:
       contents: read
       packages: write
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: scpcaTools
 Type: Package
 Title: Single cell data tools for ScPCA
-Version: 0.2.3
+Version: 0.3.0
 Authors@R: c(
     person("Ally", "Hawkins",
            email = "ally.hawkins@ccdatalab.org",
@@ -27,10 +27,10 @@ Encoding: UTF-8
 LazyData: true
 RoxygenNote: 7.2.3
 Depends:
-    R (>= 4.2.0)
+    R (>= 4.3.0)
 Imports:
     BiocGenerics,
-    bluster,
+    bluster (>= 1.10),
     dplyr,
     DropletUtils,
     fishpond,
@@ -39,7 +39,7 @@ Imports:
     jsonlite,
     Matrix,
     methods,
-    miQC (>= 1.2.0),
+    miQC,
     purrr,
     rlang,
     S4Vectors,
@@ -66,6 +66,5 @@ Suggests:
     testthat (>= 3.0.0),
     zellkonverter
 Remotes:
-    AlexsLemonade/scpcaData,
-    github::immunogenomics/lisi
+    AlexsLemonade/scpcaData
 Config/testthat/edition: 3
--- a/R/calculate_ilisi.R
+++ b/R/calculate_ilisi.R
@@ -63,7 +63,7 @@ calculate_ilisi <- function(merged_sce,
       batch_id = batch_df$batch,
       # add normalized iLISI score as in in scib approach:
       # https://github.com/theislab/scib/blob/067eb1aee7044f5ce0652fa363ec8deab0e9668d/scib/metrics/lisi.py#L98-L100
-      ilisi_score_norm = (.data$ilisi_score - 1)/(num_batches - 1)
+      ilisi_score_norm = (.data$ilisi_score - 1) / (num_batches - 1)
     )
 
   # Return data frame with cell-wise iLISI scores and associated cell & batch identifiers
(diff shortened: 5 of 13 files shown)


docker/Dockerfile
```diff
@@ -20,6 +20,7 @@ RUN Rscript -e "renv::consent(provided = TRUE); \
       rm -rf /tmp/Rtmp*
 
 # Complete installation of zellkonverter conda env
+ENV BASILISK_EXTERNAL_DIR /usr/local/renv/basilisk
```
sjspielman, line 23: How does a path like this work if someone doesn't have permissions to be working in `/usr/local`, e.g. on a server?

jashapiro (commented):
Thanks for having a look Stephanie!

Don't know why `basilisk` still installs a rather old `conda` version, but I'm not doing anything about that.

Responses to your other comments below:

docker/Dockerfile
sjspielman: Ah, yes, Dockerfile! Indeed.

allyhawkins:
Unfortunately this didn't work so cleanly for me... 

After updating to `rig 0.5.3`, I was able to run `rig install 4.1.3 --arch x86_64`. I then opened the project and ran `renv::restore()`, which resulted in this error: 

```
Installing zlibbioc [1.46.0] ...
	FAILED
Error installing package 'zlibbioc':
====================================

* installing *source* package ‘zlibbioc’ ...
** using staged installation
configure: creating ./config.status
config.status: creating src/Makevars
** libs
xcrun: error: invalid active developer path (/Library/Developer/CommandLineTools), missing xcrun at: /Library/Developer/CommandLineTools/usr/bin/xcrun
xcrun: error: invalid active developer path (/Library/Developer/CommandLineTools), missing xcrun at: /Library/Developer/CommandLineTools/usr/bin/xcrun
ERROR: compilation failed for package ‘zlibbioc’
* removing ‘/Users/allyhawkins/Documents/ALSF/git_repos/scpcaTools/renv/staging/4/zlibbioc’
Error: install of package 'zlibbioc' failed [error code 1]
In addition: Warning messages:
1: could not retrieve available packages for url 'https://bioconductor.org/packages/3.17/bioc/bin/macosx/contrib/4.1' 
2: could not retrieve available packages for url 'https://bioconductor.org/packages/3.17/books/bin/macosx/contrib/4.1' 
3: could not retrieve available packages for url 'https://bioconductor.org/packages/3.17/data/annotation/bin/macosx/contrib/4.1' 
4: could not retrieve available packages for url 'https://bioconductor.org/packages/3.17/data/experiment/bin/macosx/contrib/4.1' 
5: could not retrieve available packages for url 'https://bioconductor.org/packages/3.17/workflows/bin/macosx/contrib/4.1' 
6: In system2(R(), args, stdout = TRUE, stderr = TRUE) :
  running command ''/Library/Frameworks/R.framework/Versions/4.1/Resources/bin/R' CMD config CC 2>&1' had status 1
7: In if (eval(cond, envir = environment(dot))) return(eval(expr, envir = environment(dot))) :
  the condition has length > 1 and only the first element will be used
8: In if (eval(cond, envir = environment(dot))) return(eval(expr, envir = environment(dot))) :
  the condition has length > 1 and only the first element will be used
```

This is interesting because in trying to restore the environment that was updated in `sc-data-integration` I saw this similar error `In if (eval(cond, envir = environment(dot))) return(eval(expr, envir = environment(dot))) :
  the condition has length > 1 and only the first element will be used` and a failure to install `Matrix`. 
I feel like this might be a personal problem...

sjspielman:
@allyhawkins are you sure about the `rig` line? We want R **4.3.1** not **4.1.3**, easy thing to flip by accident! Also, I had failures with the `x86_64` arch and did `arm64` instead, but I don't remember if you are on an M1 or Intel. If M1, you probably want `arm64`?

allyhawkins:
> @allyhawkins are you sure about the rig line? We want R 4.3.1 not 4.1.3, easy thing to flip by accident! Also, I had failures with the x86_64 arch and did arm64 instead, but I don't remember if you are on an M1 or Intel. If M1, you probably want arm64?

Yea I was just coming to comment that I must have installed 4.1.3 🤦‍♀️ And I'm on intel, I don't have an M1.

jashapiro:
If you are on intel, you shouldn't have to specify the architecture for `rig` (and the default should have changed on M1 to `arm64`, so you should only _need_ that argument if you want the other version)

jashapiro:
Side note: `In if (...) : the condition has length > 1 and only the first element will be used` is a warning that will never occur in R 4.3 or higher. If you give an `if` a vector with more than one element it is now a full error. I'm glad they finally made this change.

allyhawkins:
Okay I am definitely using R 4.3 now and I open up the lock file and try to restore, but I keep getting this message about activating the project? 

```
> renv::restore()
It looks like you've called renv::restore() in a project that hasn't been activated yet.
How would you like to proceed? 

1: Activate the project and use the project library.
2: Do not activate the project and use the current library paths.
3: Cancel and resolve the situation another way.

Selection: 1

Restarting R session...
```

jashapiro:
> Okay I am definitely using R 4.3 now and I open up the lock file and try to restore, but I keep getting this message about activating the project?
> 

When you say you keep getting it, I assume you mean that after you choose option 1 and R restarts you get the same thing?

If so, I might try 2 things: One is running `renv::activate()` manually before `renv::restore()`. If that doesn't work, maybe deleting any folders from your `renv` directory? Leave `activate.R` and `settings.json`.

allyhawkins:
Note to self: If updating mac software before doing anything with renv, you need to re-install command line tools and gfortran and that will solve all your problems. 

This worked successfully after I figured out my own personal mac issues!

jashapiro:
I made a few updates to the way we handle repos here, defaulting to true CRAN, but restoring from PPM on linux (or more accurately not Mac or Windows). I also updated to the lastest versions of everything as of today.

sjspielman (approved):
🚀  Launched R 4.3.1 from terminal, ran `renv::restore()`, and it restored without a hitch, including both R and python packages!

PR #217, opened 2023-08-24
Function to incorporate sample metadata into colData of SCE object

### LINKED ISSUE #208: Function for adding sample metadata to sce object

allyhawkins:
This PR closes #208 and adds a new function to incorporate the `sample_metadata` from the metadata of an SCE object into the `colData`. This function will mainly be utilized before converting SCE objects to AnnData objects in `scpca-nf`. 

- This takes as input a SCE object that contains `sample_metadata` in the metadata slot. It also requires a `batch_column`, which is the name of the column in _both_ the `colData(sce)` and `metadata(sce)$sample_metadata` that corresponds to the batch or library ID from which the library(s) in the SCE object originates from. This means individual SCE objects should have that column added to the `colData` before using this function. 
- This works on either objects with a single library or merged objects. The only requirement for the object is that a `sample_metadata` data frame exists in the metadata and a column with the corresponding library id information is present in the `colData`. This column is automatically added when merging objects. Doing this also allows us to explicitly tell the function where to find the library ID information -  rather than assuming where it lives in the `metadata(sce)` (which is different for individual and merged objects). 
- I added test functions here that test both an individual library and a merged object. In generating the merged object, I noticed that we require `rowData` to be present in the `merge_sce_list()` function and don't have a check for it. So I added a slight modification to account for when the objects in the `sce_list` don't have any `rowData`. 
- I plan on following a follow-up PR to `scpca-nf` showing how I incorporated this function into the workflow before converting to AnnData for reference. 
- One last note is that this assumes we want to add the entirety of the `sample_metadata` to the `colData`. I think this is the most generalizable way to do it rather than providing a list of column names to keep. If there are any columns that we don't want to include, we probably just want to remove those from the `sample_metadata` before adding that to the metadata in the first place (I'm thinking about columns like `upload_date`).


jashapiro (commented):
I think this looks good for the single sample case (with a few comments below), but sadly we have a few datasets with multiplexing (really one, but more probably coming), and we probably need to think about how we will deal with that here. 

I'm assuming we are doing this conversion late in the process, when we might have already assigned samples, in which case we can probably assume that there is a `sample_id` column that we have arrived at somehow and added to the `colData`. In that case the `left_join` step should be fine, but some of the earlier checks will need to be reworked. 

Alternatively, we could check whether there are multiple samples for a given library, and fail in that case, but I feel like we ought to be able to handle it without too much reworking.

R/metadata_to_coldata.R
```diff
@@ -0,0 +1,59 @@
...
+#' @param batch_column A character value giving the colData column name
+#'  to differentiate originating SingleCellExperiment objects. Often these values
+#'  are unique library IDs. Default value is `library_id`.
+#'
+#' @return SingleCellExperiment object with the contents of `metadata(sce)$sample_metadata`
+#'   added as new columns to the `colData(sce)`.
+#' @export
+#'
+metadata_to_coldata <- function(sce,
+                                batch_column = "library_id") {
```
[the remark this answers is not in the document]
allyhawkins: The reason I used `batch_column` is just because we use it in other places in `scpcaTools`. Should I change it just here and in `merge_sce_list`? Or just here? 

Also, I thought we weren't sure how we planned to handle cell hashing data with the AnnData objects. So I hadn't been thinking about those here. But I do like the idea of being able to join by multiple columns if desired.

tests/testthat/test-metadata_to_coldata.R
```diff
@@ -0,0 +1,136 @@
...
+sce$library_id <- "library_id"
+
+# create a merged object for testing the function still works with a merged object
+set.seed(1665)
+total_cells <- 24
+total_genes <- 12
+# create three individual objects
+sce1 <- scpcaTools:::sim_sce(n_cells = total_cells / 3, n_genes = total_genes, n_empty = 0)
+sce2 <- scpcaTools:::sim_sce(n_cells = total_cells / 3, n_genes = total_genes, n_empty = 0)
+sce3 <- scpcaTools:::sim_sce(n_cells = total_cells / 3, n_genes = total_genes, n_empty = 0)
```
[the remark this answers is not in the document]
allyhawkins: I really thought I had deleted those this time 😢

R/metadata_to_coldata.R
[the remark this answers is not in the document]
allyhawkins: Actually based on the fact that we would be allowing multiple columns, I think I'm going to change the variable name to `join_columns`.

allyhawkins:
I went ahead and changed the `batch_column` to be named `join_columns` and allowed for more than one column to be used there. This should account for instances where we have multiplexed data. 

I updated the checks to account for multiple columns, modified the check for the `sample_metadata` so that we just get a warning and return the original object, and updated the checks for matching the contents between the sce and the sample metadata. Now we allow for samples that are not present in the sce object and just give a warning. This should be ready for another look @jashapiro.

R/merge_sce_list.R
```diff
@@ -205,11 +205,14 @@ prepare_sce_for_merge <- function(sce,
   #  present in `preserve_rowdata_cols`
   original_colnames <- colnames(rowData(sce))
 
-  colnames(rowData(sce)) <- ifelse(
-    original_colnames %in% preserve_rowdata_cols,
-    original_colnames,
-    glue::glue("{sce_name}-{original_colnames}")
-  )
+  # only rename rowData if there is any rowData to begin with
+  if(length(original_colnames) > 0){
```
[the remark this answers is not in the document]
jashapiro: This seems like an error in the DataFrame implementation: a matrix with a zero dimension will happily let you assign a zero length thing to the colnames... But that's what we've got I guess!

tests/testthat/test-metadata_to_coldata.R
```diff
@@ -0,0 +1,160 @@
...
+merged_sce <- merge_sce_list(sce_list = sce_list,
+                             retain_coldata_cols = "sum")
+
+
+test_that("`metadata_to_coldata` works as expected with a single object", {
+
+  sce_with_metadata <- metadata_to_coldata(sce,
+                                           join_columns = "library_id")
+
+  # check that colData now contains sample id column
```
jashapiro, line 67:
```suggestion
  # check that colData now contains sample id and diagnosis columns
```

tests/testthat/test-metadata_to_coldata.R
```diff
@@ -0,0 +1,160 @@
+# establish single test sce
+sce <- sim_sce(n_genes = 20, n_cells = 100, n_empty = 0)
+metadata(sce)$sample_id <- "sample_id"
+metadata(sce)$library_id <- "library_id"
+
+# create sample data frame
+sample_metadata_df <- data.frame(
+  sample_id = "sample_id",
+  library_id = "library_id",
+  diagnosis = "diagnosis"
```
jashapiro, line 10: Just because I started to get confused here, can we change these to different values from the column name? (Note that tests will need to be changed too)
```suggestion
  sample_id = "sample1",
  library_id = "library1",
  diagnosis = "diagnosis1"
```

R/metadata_to_coldata.R
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    # pull out sample metadata
+    sample_metadata_df <- metadata_list$sample_metadata
+    # check that batch column exists
+    if(!all(join_columns %in% colnames(sample_metadata_df))){
+      stop("One or more of the specified `join_columns` are not a column in `metadata(sce)$sample_metadata.")
+    }
+
+    # check that batches match in colData and sample metadata
```
jashapiro, line 41:
```suggestion
    # check that join columns match in colData and sample metadata
```

R/metadata_to_coldata.R
```diff
@@ -0,0 +1,69 @@
...
+
+    if(!is.null(mismatching_columns)){
+      warning(
+        glue::glue("Contents of {mismatching_columns} do not match the `metadata(sce)$sample_metadata.`")
+        )
+    }
+
+    # join coldata with sample metadata
+    coldata_df <- as.data.frame(colData(sce)) |>
+      dplyr::left_join(sample_metadata_df, by = join_columns)
```
jashapiro, line 58: In the event that there are multiple matching rows, `left_join()` will duplicate the `colData` rows. We would want to catch that: it would mean that there are multiple sample ids per library, or something like that. 

So we should add a check for `nrow(coldata_df) == nrow(colData(sce))` and fail if that is not the case.

This might fail anyway when trying to replace the `colData`, but we should probably give a better error if we can.

We should also add a test for this: where the column we are joining on is not sufficient for a unique match.

R/metadata_to_coldata.R
```diff
@@ -0,0 +1,69 @@
...
+  if (!is(sce, "SingleCellExperiment")) {
+    stop("`sce` must be a SingleCellExperiment object.")
+  }
+
+  # check that sample_metadata present in metadata of sce object
+  # if it's not present, just skip everything and return the original object
+  metadata_list <- metadata(sce)
+  if(!"sample_metadata" %in% names(metadata_list)){
+    warning("No `sample_metadata` in SCE object to add.")
+  } else{
```
jashapiro, line 27: Rather than indenting/nesting everything below, we can just short circuit the function here:
```suggestion
  if(!"sample_metadata" %in% names(metadata_list)){
    warning("No `sample_metadata` in SCE object to add.")
    return(sce)
  }
```

Since this is more parallel what happens with `stop()` calls, I find this to be more intuitive and easier to track than nesting the rest in an `else{}` block.

jashapiro (commented):
Getting close, but a few more little cases I want to account for. 

The main one is non-unique matches. If we join a table like:

|library_id|sample_id|
|--|--|
|library1|sample1|
|library1|sample2|

on only `library_id`, we could get multiple rows per original row, which would be bad. 
So we should check in general that the number of rows after the join is equal to before the join. Both in the function itself, and a test...

My other suggestions are about naming things, mostly because I got confused. 
I also caught a few comments that hadn't been updated, but there are likely more.

allyhawkins:
I went ahead and added a check to make sure the merging is unique, and we don't add more rows to `colData`. I also adjusted the tests to test for that scenario. 

@sjspielman do you mind taking a look at this since Josh is out for a while?

R/metadata_to_coldata.R
```diff
@@ -0,0 +1,75 @@
...
+#'
+#' @param sce SingleCellExperiment object to add sample metadata to colData.
+#'   Must contain the `batch_column` as a column in `colData`.
+#' @param join_columns A character value giving the names of the columns in colData
+#'  to use for joining with the `sample_metadata`. Default is `library_id`
+#'
+#' @return SingleCellExperiment object with the contents of `metadata(sce)$sample_metadata`
+#'   added as new columns to the `colData(sce)`.
+#' @import SingleCellExperiment
+#'
```
sjspielman, line 11: Tests are failing for me in part because `DataFrame` isn't in the environment, so add:
```suggestion
#' @importFrom S4Vectors DataFrame
```

R/metadata_to_coldata.R
```diff
@@ -0,0 +1,75 @@
+#' Add sample metadata to colData of SCE object
+#'
+#' @param sce SingleCellExperiment object to add sample metadata to colData.
+#'   Must contain the `batch_column` as a column in `colData`.
+#' @param join_columns A character value giving the names of the columns in colData
```
sjspielman, line 5:
```suggestion
#' @param join_columns A character vector giving the name(s) of the column(s) in colData
```

R/metadata_to_coldata.R
```diff
@@ -0,0 +1,75 @@
...
+  # check that the number of columns in colData is still the same
+  # make sure that nothing has been duplicated before adding it back into the colData
+  if(nrow(coldata_df) != nrow(colData(sce))){
+    stop("The specified `join_columns` cannot produce multiple matches.")
+  }
+
+  # replace existing coldata
+  colData(sce) <- DataFrame(coldata_df,
+                            row.names = rownames(coldata_df))
+
```
sjspielman, line 70: I think we might want to add a check here that the rownames are not NULL?

tests/testthat/test-metadata_to_coldata.R
```diff
@@ -0,0 +1,189 @@
...
+  # add a new column that is only in the colData and not in the sample metadata
+  sce$batch_column <- "library1"
+  expect_error(metadata_to_coldata(sce,
+                                   join_columns = "batch_column"))
+
+  # no sample metadata present
+  metadata(sce) <- metadata(sce)[!names(metadata(sce)) %in% "sample_metadata"]
+  expect_warning(metadata_to_coldata(sce,
+                                   join_columns = "library_id"))
+
```
sjspielman, line 143: I would move this to a different testing block, since this condition is not actually a failure!

sjspielman (commented):
What a fun journey you've been on, I see! 🎢 

Mostly things seem to me where they should be! The main thing imo is dependencies - in addition to adding `DataFrame` in the environment as I commented, tests were also failing for me because of the `testthat` version I had - it seems the function `testthat::expect_contains()` only was released with [`3.1.9`](https://github.com/r-lib/testthat/blob/29018e067f87b07805e55178f387d2a04ff8311f/NEWS.md?plain=1#L26-L30), but our `DESCRIPTION` has only `testthat (>= 3.0.0)` and similarly `renv.lock` is at `3.1.7`. We'll need to bump to `>=3.1.9` in `DESCRIPTION` and we can probably use that same version in `renv` (`renv::record("testthat@3.1.9")`, and then restore from the updated lockfile).

allyhawkins:
> Mostly things seem to me where they should be! The main thing imo is dependencies - in addition to adding DataFrame in the environment as I commented, tests were also failing for me because of the testthat version I had - it seems the function testthat::expect_contains() only was released with [3.1.9](https://github.com/r-lib/testthat/blob/29018e067f87b07805e55178f387d2a04ff8311f/NEWS.md?plain=1#L26-L30), but our DESCRIPTION has only testthat (>= 3.0.0) and similarly renv.lock is at 3.1.7. We'll need to bump to >=3.1.9 in DESCRIPTION and we can probably use that same version in renv (renv::record("testthat@3.1.9"), and then restore from the updated lockfile).

This is interesting to me, because I would think that the tests would fail in GHA if the dependencies weren't set up correctly? I had used `expect_contains` in other tests and things are passing GHA?

sjspielman:
> This is interesting to me, because I would think that the tests would fail in GHA if the dependencies weren't set up correctly? I had used expect_contains in other tests and things are passing GHA?

You are certainly not wrong, and now I am confused...
Poking round a bit, 

The `renv.lock` file in the repo currently has this:
https://github.com/AlexsLemonade/scpcaTools/blob/0f4f3eaaf861efb0564813616bfb926f7306dd94/docker/renv.lock#L4300-L4303


And yet during the GHA, this was installed:
Screenshot 2023-08-28 at 2 46 22 PM

No immediate thoughts other than.."hmm"!

R/metadata_to_coldata.R
allyhawkins: So I added this check, but then when I was testing it, I was noticing that the `rownames` will actually never be `NULL`. If there are no row names provided, each row is numbered so `is.null(rownames(colData(sce))` is actually false. I don't think we need this check because if the input SCE doesn't have rownames then the ouptut one also won't. I think that's okay.

allyhawkins:
Thanks for taking a look at this. I went ahead and incorporated your suggestions, other than the null row names check. I also updated to use `testthat 3.1.9`, although I'm still not quite sure why GHA is passing. This should be ready for another look.

sjspielman (approved):
LGTM! I'd run a `/style` first, but otherwise I think this is set!

allyhawkins:
> LGTM! I'd run a `/style` first, but otherwise I think this is set!

Thank you for the reminder!

PR #219, opened 2023-09-05
Add argument for assay name to use for X in AnnData object

### LINKED ISSUE #218: Modify sce_to_anndata to specify assay to save to X
We will need to be able to choose which assay gets saved as the `X` layer in AnnData when converting objects in `scpca-nf`. (see AlexsLemonade/scpca-nf#429). 

To do this we should modify `sce_to_anndata()` to add a new argument for the layer to save to X (`x_assay`) instead of hard coding converting the `counts` assay as we do now. 

https://github.com/AlexsLemonade/scpcaTools/blob/d7b11a2e096d48d7fe7dcea1845a7e682b05f3b0/R/sce_to_anndata.R#L46

allyhawkins:
Closes #218 

This PR adds a new argument to the function for converting SCE objects to AnnData, x_assay_name. This was pretty straight forward and I just passed the name to the zellkonverter function. 

I also updated the tests to check that it works with a different assay name and failed with an incorrect name. 

The only other thing I thought of here was trying to save an object with `logcounts` to the X assay and then `counts` to the `raw.X`. The only way to save things to `raw.X` is to use `raw=TRUE` with the zellkonverter function. When I try that, it doesn't save anything to the `X` assay even if you specify what to save. So I think for saving things to `raw.X` we will have to go in and directly modify the AnnData object after conversion.

### DIFF
--- a/R/sce_to_anndata.R
+++ b/R/sce_to_anndata.R
@@ -2,7 +2,7 @@
 #'
 #' @param sce SingleCellExperiment object to be converted to AnnData as an HDF5 file
 #' @param anndata_file Path to output AnnData file. Must be an `.h5` or `.hdf5`
-#' @param x_assay_name Name of assay in SCE object to save as X in AnnData
+#' @param x_assay_name Name of assay in SCE object to save as X in AnnData. Default is "counts".
 #'
 #' @return original SingleCellExperiment object used as input (invisibly)
 #' **Note that any columns present in the `rowData` of an SCE object that contains
--- a/R/sce_to_anndata.R
+++ b/R/sce_to_anndata.R
@@ -49,7 +49,8 @@ sce_to_anndata <- function(sce, anndata_file, x_assay_name = "counts") {
   }
 
   # export SCE object as AnnData to HDF5 file
-  zellkonverter::writeH5AD(sce_to_convert, file = anndata_file, X_name = x_assay_name)
-
+  zellkonverter::writeH5AD(sce_to_convert,
+                           file = anndata_file,
+                           X_name = x_assay_name)
   invisible(sce)
 }
--- a/R/sce_to_anndata.R
+++ b/R/sce_to_anndata.R
@@ -2,6 +2,7 @@
 #'
 #' @param sce SingleCellExperiment object to be converted to AnnData as an HDF5 file
 #' @param anndata_file Path to output AnnData file. Must be an `.h5` or `.hdf5`
+#' @param x_assay_name Name of assay in SCE object to save as X in AnnData
 #'
 #' @return original SingleCellExperiment object used as input (invisibly)
 #' **Note that any columns present in the `rowData` of an SCE object that contains
@@ -19,7 +20,7 @@
 #'   anndata_file = "test_anndata.h5"
 #' )
 #' }
-sce_to_anndata <- function(sce, anndata_file) {
+sce_to_anndata <- function(sce, anndata_file, x_assay_name = "counts") {
   if (!requireNamespace("zellkonverter", quietly = TRUE)) {
     stop("The zellkonverter package must be installed to convert objects to AnnData. No output file written.")
   }
@@ -33,6 +34,11 @@ sce_to_anndata <- function(sce, anndata_file) {
     stop("`anndata_file` must end in either '.hdf5' or '.h5'")
   }
 
+  # make sure assay is found in sce object
+  if(!x_assay_name %in% assayNames(sce)){
+    stop("`x_assay_name` is not an assay in `sce`")
+  }
+
   # assign SCE to new variable to avoid modifying input SCE
   sce_to_convert <- sce
 
@@ -43,7 +49,7 @@ sce_to_anndata <- function(sce, anndata_file) {
   }
 
   # export SCE object as AnnData to HDF5 file
-  zellkonverter::writeH5AD(sce_to_convert, file = anndata_file, X_name = "counts")
+  zellkonverter::writeH5AD(sce_to_convert, file = anndata_file, X_name = x_assay_name)
 
   invisible(sce)
 }
--- a/tests/testthat/test-sce_to_anndata.R
+++ b/tests/testthat/test-sce_to_anndata.R
@@ -26,7 +26,7 @@ test_that("Conversion of SCE to AnnData works as expected", {
   # test that H5 file is created with new assay name
   # add logcounts
   logcounts(sce) <- counts(sce)
-  new_anndata_file <- "test_anndata_logcounts.h5"
+  new_anndata_file <- tempfile(fileext = ".h5")
   expect_snapshot_file({
     sce_to_anndata(sce, new_anndata_file, x_assay_name = "logcounts")
     new_anndata_file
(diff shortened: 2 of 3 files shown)


R/sce_to_anndata.R
```diff
@@ -2,6 +2,7 @@
 #'
 #' @param sce SingleCellExperiment object to be converted to AnnData as an HDF5 file
 #' @param anndata_file Path to output AnnData file. Must be an `.h5` or `.hdf5`
+#' @param x_assay_name Name of assay in SCE object to save as X in AnnData
```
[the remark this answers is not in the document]
sjspielman: but see my next comment..

sjspielman (commented):
I looked a bit into zellkonverter re- "The only other thing I thought of here was trying to save an object with logcounts to the X assay and then counts to the raw.X." I found this comment in a recent issue, and I expect the situation is still the same - they don't have functionality at this time to write raw to hd5 (https://github.com/theislab/zellkonverter/issues/97#issuecomment-1597237380)

> We are currently able to read the raw slot from a file but not write it to a new .h5ad. There may have been a technical reason for this but more likely it was just overlooked. If this is a feature you would like to have we can discuss adding it (or a PR would be even better).

Is there an issue open for this next step of getting the raw data in? I'd imagine that goes in `scpca-nf`.

R/sce_to_anndata.R
```diff
@@ -43,7 +49,7 @@ sce_to_anndata <- function(sce, anndata_file) {
   }
 
   # export SCE object as AnnData to HDF5 file
-  zellkonverter::writeH5AD(sce_to_convert, file = anndata_file, X_name = "counts")
+  zellkonverter::writeH5AD(sce_to_convert, file = anndata_file, X_name = x_assay_name)
 
   invisible(sce)
```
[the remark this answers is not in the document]
jashapiro: I think the convention of `return()` or `invisible()` is fine: I don't see a need to wrap `invisible()` as there is pretty much no other context where it is used.

R/sce_to_anndata.R
```diff
@@ -19,7 +20,7 @@
 #'   anndata_file = "test_anndata.h5"
 #' )
 #' }
-sce_to_anndata <- function(sce, anndata_file) {
+sce_to_anndata <- function(sce, anndata_file, x_assay_name = "counts") {
```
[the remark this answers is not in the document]
jashapiro: I think the default of saving `counts` to X makes sense, mostly as that preserves the default `zellconverter` behavior unless we start to mess around with the the order of assays in the SCE. (If we wanted to perfectly follow the behavior we would pass `NULL`, which would take the first assay. But I like the explicit name.)

tests/testthat/test-sce_to_anndata.R
```diff
@@ -3,10 +3,13 @@ sce <- sim_sce(n_cells = 100, n_genes = 200, n_empty = 0)
 colData(sce) <- DataFrame("test_column" = sample(0:10, 100, rep = TRUE))
 rowData(sce) <- DataFrame("test_row" = sample(0:10, 200, rep = TRUE))
 
+# define anndata output
+anndata_file <- "test_anndata.h5"
+# quiet messages that will come with zellkonverter
+suppressPackageStartupMessages(library(zellkonverter))
```
jashapiro, line 9: For convention, we should probably move the package loading to the top of the file.

R/sce_to_anndata.R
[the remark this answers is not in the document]
allyhawkins: We also will be using `counts` the majority of the time. The only time we wouldn't be using `counts` is if we have `logcounts`, which is just the processed object.

tests/testthat/test-sce_to_anndata.R
```diff
@@ -20,7 +23,23 @@ test_that("Conversion of SCE to AnnData works as expected", {
   expect_equal(colnames(colData(sce)), colnames(colData(converted_sce)))
   expect_equal(colnames(rowData(sce)), colnames(rowData(converted_sce)))
 
+  # test that H5 file is created with new assay name
+  # add logcounts
+  logcounts(sce) <- counts(sce)
+  new_anndata_file <- "test_anndata_logcounts.h5"
+  expect_snapshot_file({
```
jashapiro, line 30: This test is probably fine for most cases, but we could reload the file and check its contents as well for a more complete test. Since we already have the package loaded, I would reload the file with `zellconverter::AnnData2SCE(new_anndata_file)` directly and check the contents of the `X` assay are equal to `logcounts(sce)`.

tests/testthat/test-sce_to_anndata.R
```diff
@@ -20,7 +23,23 @@ test_that("Conversion of SCE to AnnData works as expected", {
   expect_equal(colnames(colData(sce)), colnames(colData(converted_sce)))
   expect_equal(colnames(rowData(sce)), colnames(rowData(converted_sce)))
 
+  # test that H5 file is created with new assay name
+  # add logcounts
+  logcounts(sce) <- counts(sce)
+  new_anndata_file <- "test_anndata_logcounts.h5"
```
jashapiro, line 29: Better to use temporary files in tests. I don't know how extensively we hard code file names that we are creating elsewhere, but it is probably something we should look to clean up as we see it.
```suggestion
  new_anndata_file <- tempfile(fileext = ".h5")
```

allyhawkins:
> Is there an issue open for this next step of getting the raw data in? I'd imagine that goes in scpca-nf.

Both https://github.com/AlexsLemonade/scpca-nf/issues/429 and https://github.com/AlexsLemonade/scpca-nf/issues/436 are filed to track the next steps from here. 

I fixed the comment you had suggested, but am keeping the default as `counts` and the `invisible`. This should be ready for another look.

tests/testthat/test-sce_to_anndata.R
```diff
@@ -20,7 +23,23 @@ test_that("Conversion of SCE to AnnData works as expected", {
...
+    sce_to_anndata(sce, new_anndata_file, x_assay_name = "logcounts")
+    new_anndata_file
+  })
+
+})
+
+test_that("Conversion of SCE to AnnData fails as expected", {
+
   # check that inputting an improper file name causes a failure
   anndata_bad_file <- "test_anndata"
```
jashapiro, line 40: Probably should use a tempfile here too, with a bad extension... maybe: 
```suggestion
  anndata_bad_file <- tempfile(fileext = ".rds")
```

jashapiro (approved):
This looks good to me, with a couple little suggestions including a slightly more explicit test that things went the way they were expected to on write.

tests/testthat/test-sce_to_anndata.R
allyhawkins: I thought about this, but when you read the object back in it just uses the stored name for the `X_assay` so the result is an sce object with both `logcounts` and `counts` as assays, regardless of what is stored in `X`.

jashapiro: what happens when you use `zellconverter::AnnData2SCE(new_anndata_file, X_name = "X")`?

allyhawkins: I used `zellkonverter::readH5AD` since `AnnData2SCE` requires `reticulate` and an `AnnData` object actually stored in R. That function also has the `X_name` argument, so I used that to add these tests in https://github.com/AlexsLemonade/scpcaTools/pull/219/commits/30f2569a06451d417976f5c613ed2a3d44e8b5c3

sjspielman (approved):
lgtm!

PR #220, opened 2023-09-13
Add some more metadata terms as arguments when building an SCE

allyhawkins:
Related to https://github.com/AlexsLemonade/scpca-nf/issues/419 and https://github.com/AlexsLemonade/scpca-nf/issues/430

This PR continues the steps needed to get the `assay_ontology_term_id`, `seq_unit`, and title (`project_id`) into the SCE object and ultimately the AnnData object. Right now we use the `read_alevin` function to read in the output from alevin-fry and build the unfiltered SCE object. This includes adding things like `sample_id` and`tech_version` to the metadata with the `read_alevin_metadata` function. I thought it would make sense to add the new library-specific metadata terms we need in the same way and have included them as optional arguments here.

### DIFF
--- a/R/read_alevin.R
+++ b/R/read_alevin.R
@@ -17,7 +17,10 @@
 #' @param library_id Optional library identifier
 #' @param sample_id Optional sample identifier.
 #'   If multiplexed samples are included in a library, this may be a vector.
+#' @param project_id Optional project identifier.
 #' @param tech_version Technology or kit used to process library (i.e. 10Xv3, 10Xv3.1).
+#' @param assay_ontology_term_id Optional Experimental Factor Ontology term associated with the provided `tech_version`
+#' @param seq_unit Optional sequencing unit associated with the library (i.e. cell, nucleus)
 #'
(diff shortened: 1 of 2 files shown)


sjspielman (approved):
Makes sense to me!

R/read_alevin.R
```diff
@@ -169,18 +179,24 @@ read_tximport <- function(quant_dir) {
...
+#' @param project_id Optional project identifier.
 #'
 #' @return A list containing alevin run metadata,
 #'   with NULL values for missing elements.
 #'
 #' @noRd
 read_alevin_metadata <- function(quant_dir,
                                  tech_version,
+                                 assay_ontology_term_id = NULL,
+                                 seq_unit = NULL,
```
jashapiro, line 196: Before merging, can we be more consistent about argument order between these functions?

allyhawkins: Sorry I can reopen it... I was just following the order that was there and grouping the seq unit/ tech/ ontology id together and then the library id/ sample id/ project id together.

jashapiro: I guess I'm too late.

PR #222, opened 2023-09-14
Add handling of NA_character_ columns for anndata export

jashapiro:
As seen in https://github.com/AlexsLemonade/scpca-nf/pull/453#issue-1896967770, we sometimes run into trouble with Anndata exports when a column is listed as a character column but is all `NA`.

This code should transform any columns in `colData` or `rowData` that are all NA to logical types before saving. I added tests to be sure that it is working as expected.

I confirmed that without the added code, the tests fail, but they pass with it.

In testing also realized that by using `tempfile()` we lose the snapshot features of `testthat`, so I am now using a temporary directory, but set file names.

### DIFF
--- a/R/sce_to_anndata.R
+++ b/R/sce_to_anndata.R
@@ -35,7 +35,7 @@ sce_to_anndata <- function(sce, anndata_file, x_assay_name = "counts") {
   }
 
   # make sure assay is found in sce object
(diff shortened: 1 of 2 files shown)


allyhawkins (approved):
Looks good, thanks for doing this!

PR #224, opened 2023-09-18
Fail conversion to AnnData if SCE object doesn't have enough cells

allyhawkins:
Since we recently discovered you need at least 2 cells to create an AnnData object using zellkonverter, I thought it would be a good idea to have a check for the number of cells in `sce_to_anndata`. Here, if an object has < 2 cells, then an error is produced.


jashapiro (approved):
LGTM!
(Also, the saddest error. I almost want it to say: "I know we said single cell experiment, but not like that!")

allyhawkins:
> (Also, the saddest error. I almost want it to say: "I know we said single cell experiment, but not like that!")

I got a good laugh out of this, thank you 😂

PR #226, opened 2023-10-16
Add ggforce to docker image

allyhawkins:
In working on https://github.com/AlexsLemonade/scpca-nf/issues/446, I was getting errors rendering the supplemental report in Nextflow. It looks like we are missing `ggforce` from the `scpcaTools` docker image. So I added it here to dependencies and to the lock file. It also had a lot of dependencies that got updated.


sjspielman (commented):
Ran `renv::restore()` with these changes, and all worked as expected!

We should probably update the repos here though - in https://github.com/AlexsLemonade/scpcaTools/pull/216 we solidified everything to use `CRAN` ([see commit here](https://github.com/AlexsLemonade/scpcaTools/pull/216/commits/eed9663d7ae4a8a36a4481bd909a04e107f7132e)), not `RSPM` which seems to have gotten used here. I'm not sure why `CRAN` wasn't used automatically since we have `CRAN` as the default repo in the lockfile, but here we are I suppose.. So, I guess change it manually with `ctl+F` to replace `"Repository": "RSPM"` with` "Repository": "CRAN"` in the lockfile directly, and then we'll make sure it still works as expected!

allyhawkins:
> Actually one other comment - I recall when adding ComplexHeatmap into renv, I actively decided we didn't need ggridges because it's also listed as one of Seurat's requirements. So, I guess it isn't also installed along with Seurat..?

I don't think it was ever part of Seurat? It's not in the diff as being removed from that list. But either way it's in the lock file now.

allyhawkins:
`ggridges` also appears to have already been there.

allyhawkins:
Just noting that I'm seeing the test fail here, but I can't reproduce it locally. This looks like the problem, but even when I run the harmony functions locally I'm not able to re-generate this error... 

I also am seeing a ton of `HarmonyMatrix` deprecated warnings. But the version we are using (v0.1.1) shouldn't have these warnings, they just recently were added in [1.0.3](https://cran.r-project.org/web/packages/harmony/index.html). 

```
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure ('test-integrate_sces.R:217:3'): `integrate_sces` works as expected with a harmony covariate ──
Expected `integrate_sces(merged_sce, "harmony", batch_column, covariate_cols = "covariate")` to run without any errors.
i Actually got a <simpleError> with text:
  You specified a lambda value for each covariate but the number of lambdas specified (1) and the number of covariates (2) mismatch.
```

Tagging @jashapiro and @sjspielman to see if anyone else has any helpful insight.

jashapiro:
The remote tests are done with whatever versions of packages are current now, in order to catch these kinds of errors. Which is to say that remote tests do not use `renv`. This was intentional, so that we would catch some of the errors that a new user might run into. So to properly debug this locally, you would want to create a fresh testing environment. 

It looks like here we are getting an error that may be the result of a change in behavior or stricter checking in new versions.

So I think there are two things here: we might want to upgrade to Harmony 1.0, and we might also want to separate out the **required** testing action to use `renv`, which is what we really care about most of the time, but separately test using the latest package versions (the current test) and maybe make that a flag.

sjspielman:
Noting I also don't see this error locally when running `devtools::check()` from my R `4.3.1` at the most recent commit, I get this _single_ warning - 

```
Warning (test-metadata_to_coldata.R:171:3): `metadata_to_coldata` works as expected with multiplexed libraries
Detected an unexpected many-to-many relationship between `x` and `y`.
i Row 1 of `x` matches multiple rows in `y`.
i Row 1 of `y` matches multiple rows in `x`.
i If a many-to-many relationship is expected, set `relationship = "many-to-many"` to silence this warning.
Backtrace:
     ▆
  1. ├─testthat::expect_error(metadata_to_coldata(sce, join_columns = "library_id")) at test-metadata_to_coldata.R:171:2
  2. │ └─testthat:::expect_condition_matching(...)
  3. │   └─testthat:::quasi_capture(...)
  4. │     ├─testthat (local) .capture(...)
  5. │     │ └─base::withCallingHandlers(...)
  6. │     └─rlang::eval_bare(quo_get_expr(.quo), quo_get_env(.quo))
  7. ├─scpcaTools::metadata_to_coldata(sce, join_columns = "library_id")
  8. │ ├─dplyr::left_join(...) at scpcaTools/R/metadata_to_coldata.R:58:2
  9. │ └─dplyr:::left_join.data.frame(...)
 10. │   └─dplyr:::join_mutate(...)
 11. │     └─dplyr:::join_rows(...)
 12. │       └─dplyr:::dplyr_locate_matches(...)
 13. │         ├─base::withCallingHandlers(...)
 14. │         └─vctrs::vec_locate_matches(...)
 15. ├─vctrs:::warn_matches_relationship_many_to_many(...)
 16. │ └─vctrs:::warn_matches_relationship(...)
 17. │   └─vctrs:::warn_matches(...)
 18. │     └─vctrs:::warn_vctrs(...)
 19. │       └─rlang::warn(...)
 20. │         └─base::warning(cnd)
 21. │           └─base::withRestarts(...)
 22. │             └─base (local) withOneRestart(expr, restarts[[1L]])
 23. │               └─base (local) doWithOneRestart(return(expr), restart)
 24. └─dplyr (local) `<fn>`(`<vc______>`)
 25.   └─dplyr:::rethrow_warning_join_relationship_many_to_many(cnd, error_call)
 26.     └─dplyr:::warn_join(...)
 27.       └─dplyr:::warn_dplyr(...)
```

In the GHA log though, looks like harmony `1.0.3` is being installed in docker: https://github.com/AlexsLemonade/scpcaTools/actions/runs/6537993058/job/17752974579?pr=226#step:6:2651

I recall this happening previously, where a package version different from our lockfile got installed in Docker but I don't remember exactly why - hunting for that old PR now.

sjspielman:
Jumping off @jashapiro's suggestions above, I wonder if we want to lock down some versions in `DESCRIPTION`, too. https://github.com/AlexsLemonade/scpcaTools/blob/139a001cd590dc85df25db39597cd3d26c48c047/DESCRIPTION#L57

More generally, bumping `harmony` seems like something to do either way (in a separate PR).

allyhawkins:
> More generally, bumping harmony seems like something to do either way (in a separate PR).

Yea, I think we want to bump `harmony`, but they made quite a few changes and I would agree that should be a separate PR. 

> So I think there are two things here: we might want to upgrade to Harmony 1.0, and we might also want to separate out the required testing action to use renv, which is what we really care about most of the time, but separately test using the latest package versions (the current test) and maybe make that a flag.

I also like the idea of running the required tests with `renv` and creating a separate test to catch errors with newer versions of packages.

jashapiro:
> Noting I also don't see this error locally when running `devtools::check()` from my R `4.3.1` at the most recent commit, I get this _single_ warning -
> 
> ```
> Warning (test-metadata_to_coldata.R:171:3): `metadata_to_coldata` works as expected with multiplexed libraries
> Detected an unexpected many-to-many relationship between `x` and `y`.
> i Row 1 of `x` matches multiple rows in `y`.
> i Row 1 of `y` matches multiple rows in `x`.
> i If a many-to-many relationship is expected, set `relationship = "many-to-many"` to silence this warning.

Noting that this warning is still there (and new to whatever version of dplyr)... it is for a place where we _do_ expect a many to many mapping, because we are testing that we handle it properly (and produce an error). I think the thing to do may be to suppress warnings (with `suppressWarnings()`) in the following chunk: 
https://github.com/AlexsLemonade/scpcaTools/blob/139a001cd590dc85df25db39597cd3d26c48c047/tests/testthat/test-metadata_to_coldata.R#L170-L173

I don't think we want to modify the code that is called, as that warning could be helpful to somebody who does encounter the error and the fact is that a many-to-many mapping is a bad thing in this case! We just want **more** than a warning, which we have implemented.

docker/dependencies.R
```diff
@@ -9,6 +9,7 @@ library(eds)
 library(eisaR)
 library(ensembldb)
 library(fishpond)
+library(ggforce)
 library(ggplot2)
```
jashapiro, line 13: I'm not sure how `ggplot2` got into this file, but we should not need it as `ggplot2` is already a dependency of required packages. We also made `fishpond` a required dependency at some point, so it should be installed as well without the need to be in this file).
```suggestion
library(ggforce)
```

jashapiro (approved):
LGTM!

sjspielman:
> [spell suppress correctly](https://github.com/AlexsLemonade/scpcaTools/pull/226/commits/b4545a33b0092698109e455f202b3186b137cc38)

I feel this commit message in my soul.

PR #229, opened 2023-10-16
Use Harmony 1.0

allyhawkins:
Stacked on #226 so that we avoid conflicts with package versions. 

Based on comments in #226, this updates the lock file to use [Harmony 1.0](https://cran.r-project.org/web/packages/harmony/harmony.pdf) and updates the integration functions to incorporate any changes. 

Harmony no longer includes `HarmonyMatrix`, which is the function that we were previously using to generate results. Instead, they now have just the main `RunHarmony` that can take as input the pca matrix, SCE, or Seurat object. I chose to use the version that inputs the pca matrix since that aligns with how we were previously handling things. 

The only major change was dealing with their `lambda` argument and matching that with how we use covariates. Basically, the number of covariates specified must match up with the number of `lambda` values provided. At a minimum, we will have one covariate, the `batch_column`, and that can use the default `lambda` value of 1 that's set in the function itself. 

However, if we add an additional covariate column to use, then the number of covariates is now 2 (batch + additional column), and we haven't required changing the `lambda`, so it's still just 1. That leads to a mismatch and an error when using `RunHarmony`. To combat this, I created two new arguments, `batch_lambda` and `covariate_lambda`. I thought it would be helpful to separate them since we will always need some sort of lambda vector whether or not we have covariates. 

Another option would be to have one lambda as an argument and then automatically set all the lambdas to be equal if covariates are present, but I don't know if we want to do that. That removes the flexibility of setting lambdas for batch vs other covariates separately, which is supported in the actual `RunHarmony` function. 

This passed my tests locally, so fingers crossed it works here too 🤞

### DIFF
--- a/R/integrate_sces.R
+++ b/R/integrate_sces.R
@@ -14,6 +14,11 @@
 #' @param covariate_cols A vector of additional columns in the merged SCE to
 #'   consider as covariates during integration. Currently, this is used only by
 #'   `harmony`.
+#' @param batch_lambda The ridge regression penalty to use when correcting using the `batch_column`.
+#'   This is only used by `harmony`. Default is 1.
+#' @param covariate_lambda A vector of ridge regression penalties to use with each covariate.
+#'   A lambda value must be provided for each covariate column.
+#'   This is only used by `harmony`.
 #' @param return_corrected_expression A boolean indicating whether corrected expression
 #'   values determined by the given integration method should be included in the
 #'   integrated SCE object. Note that `harmony` does not calculate corrected expression,
@@ -32,6 +37,8 @@ integrate_sces <- function(merged_sce,
                            integration_method = c("fastMNN", "harmony"),
                            batch_column = "sample",
                            covariate_cols = c(),
+                           batch_lambda = 1,
+                           covariate_lambda = c(),
                            return_corrected_expression = FALSE,
                            seed = NULL,
                            ...) {
@@ -90,6 +97,8 @@ integrate_sces <- function(merged_sce,
       merged_sce,
       batch_column,
       covariate_cols,
+      batch_lambda,
+      covariate_lambda,
       ...
     )
   }
@@ -148,7 +157,11 @@ integrate_fastmnn <- function(merged_sce,
 #'   being integrated.
 #' @param covariate_cols A vector of other columns to consider as
 #'   covariates during integration.
-#' @param ... Additional arguments to pass into `harmony::HarmonyMatrix()`
+#' @param batch_lambda The ridge regression penalty to use when correcting using the `batch_column`.
+#'   Default is 1.
+#' @param covariate_lambda A vector of ridge regression penalties to use with each covariate.
+#'   A lambda value must be provided for each covariate column.
+#' @param ... Additional arguments to pass into `harmony::RunHarmony()`
 #'
 #' @return Integrated PCs as calculated by `harmony`
 #'
@@ -156,6 +169,8 @@ integrate_fastmnn <- function(merged_sce,
 integrate_harmony <- function(merged_sce,
                               batch_column,
                               covariate_cols = c(),
+                              batch_lambda = 1,
+                              covariate_lambda = c(),
                               ...) {
   # Check that harmony is installed
   if (!requireNamespace("harmony", quietly = TRUE)) {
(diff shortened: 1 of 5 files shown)


R/integrate_sces.R
```diff
@@ -172,17 +187,28 @@ integrate_harmony <- function(merged_sce,
     stop("The provided covariate columns are not all present in the `merged_sce` colData.")
   }
 
+  # check that the lambda values match up
+  if(length(covariate_cols) != length(covariate_lambda)){
+    stop("The number of covariate columns must be equal to the number of covariate lambda values.")
+  }
```
jashapiro, line 193: I like having the option here, but maybe we can set it automatically to 1 if not set at all? Which is to say if `length(covariate_lambda) == 0` we could set `covariate_lambda = rep(1, length(covariate_cols))`? Maybe with a warning? I also don't know how sensitive things are to this parameter. Maybe there is another default that would be better?

If all 1s is acceptable, I would  actually characterize this as their bug, but that is neither here nor there.

R/integrate_sces.R
```diff
@@ -172,17 +187,28 @@ integrate_harmony <- function(merged_sce,
...
     )
 
+  # get a vector of all lambdas
+  lambda_vec <- c(batch_lambda, covariate_lambda)
+
   # Perform integration
-  harmony_results <- harmony::HarmonyMatrix(reducedDim(merged_sce, "PCA"),
+  harmony_results <- harmony::RunHarmony(
+    data_mat = reducedDim(merged_sce, "PCA"),
     meta_data = harmony_metadata,
```
jashapiro, line 209: Is there any reason not to directly use `merged_sce` here?

If we do that we can remove the `harmony_metadata` definition as well, as it will get those automatically.
```suggestion
    merged_sce,
```
```

**Review by jashapiro — COMMENTED** [review]
I think this change is a good chance to use the built-in top level functions as built, so I am proposing that we do that rather than dropping into the more "implementation" level version. 

To do this, we would have to change the places that use this to expect the modified SCE object, which is a bit of extra work later, but it does make things more consistent between different functions.

**Comment by jashapiro** [comment]
> Haven't reviewed yet, just poking around at first. From the harmony docs on the `lambda` argument:
> 
> > Ridge regression penalty. Default lambda=1. Bigger values protect against over correction. If several covariates are specified, then lambda can also be a vector which needs to be equal length with the number of variables to be corrected. In this scenario, each covariate level group will be assigned the scalars specified by the user. If set to NULL, harmony will determine lambdas automatically and try to minimize overcorrection (beta).
> 
> A different option could be to set `lambda` to NULL, effectively ignoring it on our end, and let `harmony` figure out the values. Since we probably don't have a principled way to pick `lambda` values either, maybe this is a good middle ground.

The fact that auto-setting is considered beta makes me a bit nervous, but I do think this is something to keep in mind for the future. Which is to say we should do some mild testing to see how it looks before implementing it, as a separate issue.

**Comment by jaclyn-taroni** [comment]
+1 to not using the auto-setting feature in beta right now.

File: R/integrate_sces.R (hunk 1 above)
Comment (allyhawkins) [reply R/integrate_sces.R:193]: Given we haven't done much testing of this parameter at all, I would favor using the recommended default of 1, which is what we were previously doing by not setting it. I do like the idea of setting it, if it isn't provided though, so I'll add that check in.

**Comment by allyhawkins** [comment]
> I think this change is a good chance to use the built-in top level functions as built, so I am proposing that we do that rather than dropping into the more "implementation" level version.

@jashapiro I updated this to just use the regular `RunHarmony` with the SCE object as input and made the associated changes. I'm assuming that's what you meant when you said this, but let me know if I interpreted this incorrectly. 

Regarding the lambda, I kept everything mostly as is, but I added a check for if there is no `covariate_lambda` set, then we use a default of 1 for each covariate column.

File: R/integrate_sces.R [hunk 3]
```diff
@@ -172,20 +187,35 @@ integrate_harmony <- function(merged_sce,
...
+  lambda_vec <- c(batch_lambda, covariate_lambda)
 
   # Perform integration
-  harmony_results <- harmony::HarmonyMatrix(reducedDim(merged_sce, "PCA"),
-    meta_data = harmony_metadata,
-    vars_use = covariate_cols,
+  harmony_object <- harmony::RunHarmony(
+    merged_sce,
+    group.by.vars = covariate_cols,
+    reduction.use = "PCA",
```
jashapiro, line 210: This should be automatic, but having it explicit seems fine.

jashapiro (approved):
LGTM!

PR #232, opened 2023-10-18
Rename renv workflow for better badging

jashapiro:
This is an attempt at getting two status badges on the readme, which I hadn't expected to work so I didn't try last time. But I think maybe it will, so I'm trying a version of it here. If it doesn't work, I can revert, but I think I have to merge this to test it?


jashapiro:
The renaming in this branch means that the required test won't run here, actually. So I think I am going to change which test is required and then merge? This affects #233, so I will wait until after that is in to make the change.

PR #233, opened 2023-10-18
Remove objects and data frames from SCE metadata

### LINKED ISSUE #231: Remove any non-vector items in SCE metadata prior to converting to AnnData
> So I was testing this with the new updates to `scpcaTools`, and I'm getting an error converting SCE to AnnData objects that have the SingleR object stored in the metadata, similar to the issue we previously had with the miQC object. I think we have two options, and before I implement anything, I'm curious what others think.
> 
> 1. We can update the `sce_to_anndata` function in `scpcaTools` to check for the `singler_results` in the metadata before conversion. Then we remove the object and print out a warning that the SingleR results were removed. This is exactly how we handle the miQC model.
> 2. We can update the script we use in `scpca-nf` to prep objects prior to converting to AnnData to remove the SingleR results prior to running `sce_to_anndata`.
> 
> I tend to favor option 1 since it will catch any objects getting converted both in and out of the workflow. Tagging @jashapiro and @sjspielman for any thoughts or opinions on how to proceed.

I tend to favor option 1 as well. I wonder if we can do a more general check to remove anything that isn't a standard R vector from the metadata and print the name of the thing that was removed?

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/issues/510#issuecomment-1769057795_

allyhawkins:
Closes #231 

Here I make some modifications to the `metadata` of an SCE object before conversion. Specifically, I'm removing any objects or `data.frame`s. We've seen issues with complex objects, but I've also seen issues with `data.frame`s, depending on the contents. After removing items from metadata, a warning displays the removed items. For our processed objects, that's `sample_metadata`, `miQC_model`, `singler_results`, and `cellassign_predictions`. 
I also updated the tests slightly to ensure the proper items are removed.

### DIFF
--- a/R/sce_to_anndata.R
+++ b/R/sce_to_anndata.R
@@ -46,11 +46,22 @@ sce_to_anndata <- function(sce, anndata_file, x_assay_name = "counts") {
   # assign SCE to new variable to avoid modifying input SCE
   sce_to_convert <- sce
 
-  # remove miQC model from metadata
-  if (!is.null(metadata(sce_to_convert)$miQC_model)) {
-    metadata(sce_to_convert)$miQC_model <- NULL
-    message("miQC model cannot be converted between SCE and AnnData.")
+  # grab existing metadata
+  metadata_list <- metadata(sce_to_convert)
+
+  # remove any objects or dataframes
+  metadata_to_keep <- metadata_list |>
+    purrr::discard(is.object) |>
+    purrr::discard(is.data.frame)
+
+  # print out warning that removed objects won't be converted
+  removed_metadata <- names(metadata_list)[!names(metadata_list) %in% names(metadata_to_keep)]
+  if(length(removed_metadata) > 0){
+    glue::glue("{removed_metadata} cannot be converted between SCE and AnnData.")
   }
+
+  # reset metadata
+  metadata(sce_to_convert) <- metadata_to_keep
 
   # zellkonverter (or pandas) does not like NA values to be stored in character vectors
   colData(sce_to_convert) <- colData(sce_to_convert) |>
--- a/tests/testthat/test-sce_to_anndata.R
+++ b/tests/testthat/test-sce_to_anndata.R
@@ -12,6 +12,15 @@ colData(sce) <- DataFrame("test_column" = sample(0:10, 100, rep = TRUE),
                           row.names = barcodes)
 rowData(sce) <- DataFrame("test_row" = sample(0:10, 200, rep = TRUE))
 
+# add some metadata
+metadata(sce) <- list(
+  library_id = "library1",
+  sample_metadata = data.frame(
+    library_id = "library1",
+    sample_id = "sample1"
+  )
+)
+
 # define anndata output
 tempdir <- tempdir()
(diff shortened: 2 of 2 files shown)


R/sce_to_anndata.R
```diff
@@ -46,12 +46,23 @@ sce_to_anndata <- function(sce, anndata_file, x_assay_name = "counts") {
...
-  if (!is.null(metadata(sce_to_convert)$miQC_model)) {
-    metadata(sce_to_convert)$miQC_model <- NULL
-    message("miQC model cannot be converted between SCE and AnnData.")
+  # grab existing metadata
+  metadata_list <- metadata(sce_to_convert)
+
+  # remove any objects or dataframes
+  metadata_to_keep <- metadata_list |>
+    purrr::discard(is.object) |>
+    purrr::discard(is.data.frame)
```
jashapiro, line 55: I worry about lists too, and data frames are lists, so this captures both.
```suggestion
    purrr::discard(is.list)
```

R/sce_to_anndata.R
```diff
@@ -46,12 +46,23 @@ sce_to_anndata <- function(sce, anndata_file, x_assay_name = "counts") {
...
 
-  # remove miQC model from metadata
-  if (!is.null(metadata(sce_to_convert)$miQC_model)) {
-    metadata(sce_to_convert)$miQC_model <- NULL
-    message("miQC model cannot be converted between SCE and AnnData.")
+  # grab existing metadata
+  metadata_list <- metadata(sce_to_convert)
+
+  # remove any objects or dataframes
+  metadata_to_keep <- metadata_list |>
```
jashapiro, line 53: I think we can avoid creating extra objects?
```suggestion
  # remove any objects or dataframes
  metadata_to_keep <- metadata(sce_to_convert) |>
```

R/sce_to_anndata.R
```diff
@@ -46,12 +46,23 @@ sce_to_anndata <- function(sce, anndata_file, x_assay_name = "counts") {
...
+    purrr::discard(is.object) |>
+    purrr::discard(is.data.frame)
+
+  # print out warning that removed objects won't be converted
+  removed_metadata <- names(metadata_list)[!names(metadata_list) %in% names(metadata_to_keep)]
```
jashapiro, line 58: Slightly faster implementation, getting rid of `metadata_list`
```suggestion
  removed_metadata <- setdiff(names(metadata(sce)),  names(metadata_to_keep))
```

R/sce_to_anndata.R
```diff
@@ -46,12 +46,23 @@ sce_to_anndata <- function(sce, anndata_file, x_assay_name = "counts") {
...
+  if(length(removed_metadata) > 0){
+    glue::glue("{removed_metadata} cannot be converted between SCE and AnnData.")
```
jashapiro, line 60: I think we want to explicitly print here?
```suggestion
    glue::glue("{removed_metadata} cannot be converted between SCE and AnnData.") |>
      purrr::walk(message)
```

jashapiro (approved):
Looks good, just a few little cleaning comments.

PR #234, opened 2023-10-19
Rename R-CMD-check-current.yml to R-CMD-check-current.yaml

jashapiro:
That's it. Makes the workflow file names consistent, and fixes a link on the readme that assumed they were.


allyhawkins (approved):
👍

PR #235, opened 2023-11-06
Don't convert any single gene single-cell objects

allyhawkins:
Here, I'm adding a check to the function for converting SCE to AnnData to make sure the SCE object has > 1 gene or feature. 
This was originally mentioned in https://github.com/AlexsLemonade/scpca-nf/pull/540#issuecomment-1795572243.


jashapiro (approved):
LGTM!

Issue #239, opened 2023-11-22
Accomodate Seurat 5


jashapiro:
Seurat 5 changed some of the object formats, so our conversion script in `sce_to_seurat.R` does not work properly when using more recent versions of Seurat. (See failure https://github.com/AlexsLemonade/scpcaTools/actions/runs/6959753992/job/18937716162)

Specifically, the code fails when we try to use:
```
seurat_obj[[assay_name]]@var.features <- rowdata
```

We should do one or more of the following: 
* specify that the package requires Seurat < 5,
* modify our code to do the conversion and include `rowdata` without diving into the object structure, which might work with both versions
* check the Seurat version and use the appropriate slot name for Seurat >= 5

We also might consider updating Seurat to the latest version in the `renv` as part of this update (and specifying that version is required), but I don't know what else might be affected. (It might also remove the dependency on `sp` & `maptools`, which we are getting warnings about being defunct).

PR #240, opened 2023-11-22
Add precommit and specify Seurat version

### LINKED ISSUE #238: Add pre-commit to this repo
We use `pre-commit` in `scpca-nf`, so we may as well add it here too.

I expect we want linting and formatting in this one, and we could even use the spell checking that is part of the R [`precommit` package](https://github.com/lorenzwalthert/precommit), though I don't think we need to use that package explicitly.

jashapiro:
Closes #238
Here I added precommit to do automatic formatting and linting.

I made the linting a bit more lenient for things we probably don't care much about. The one thing that we were violating but I left was `cyclocomp_linter`, which tries to prevent too much complexity in functions. But we only violated that in one place, so I just used `# nolint` flags to disable it in that one place. 

I also did an initial run at #239 in the most simple way I could, just setting that `Seurat` has to be < v5.0.0.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -18,3 +18,7 @@
 
 # Docker folder
 ^docker$
+
+# development
+^\.pre-commit-config\.yaml$
+^\.lintr$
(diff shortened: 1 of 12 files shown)


allyhawkins (approved):
LGTM, just one typo to fix.

PR #241, opened 2023-11-22
Use AddMetaData

### LINKED ISSUE #239: Accomodate Seurat 5

jashapiro:
Closes #239 

This ended up a pretty small fix: I used the `AddMetaData` Seurat function rather than adding directly to the slot. I think we actually were adding the wrong slot anyway: we had been using `var.features`, whereas I think the correct slot was `meta.features` all along. Which can be accessed with just `object[[assay]][[]]` as I do in the tests now. 

I think I could have used `object[[assay]][[]] <- rowdata`, but this seemed safer.

I tested with both Seurat 4.4 and 5.0, and both seemed to work on my machine.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -62,7 +62,7 @@ Suggests:
     scran,
     scpcaData,
     sessioninfo,
-    Seurat (< 5.0.0),
+    Seurat,
     testthat (>= 3.1.9),
     zellkonverter
 Remotes:
--- a/R/sce_to_seurat.R
+++ b/R/sce_to_seurat.R
@@ -63,7 +63,7 @@ sce_to_seurat <- function(sce,
   )
 
   # add rowdata and metadata separately adding it while creating leaves the slots empty without warning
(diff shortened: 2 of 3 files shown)


allyhawkins (approved):
Thank you for doing this, LGTM!

Issue #150, opened 2022-11-23
Account for altExps when merging SCEs


sjspielman:
From https://github.com/AlexsLemonade/scpcaTools/pull/148#issuecomment-1325276505 (and other comments in this PR):

While implementing updates for #146, there were some challenges encountered with altExps when merging SCEs. Therefore, we'll need to circle back and include the ability for this function to handle altExps, in particular any CITEseq data which we expect in some ScPCA libraries.

allyhawkins:
We would like to go ahead and make these changes to account for merging objects with `altExp` objects. When we are working on this we should be sure to include the following: 

- A flag for whether or not we want to include the `altExp` in the returned merged object (keep/ don't keep)
- For any features that are not found in all libraries, set the value to NA for libraries in which that feature is missing. We do not want to use 0.

sjspielman:
We have discussed an updated plan for merging altExps, which will involve some associated changes with the main experiment merging:

- If `include_altexps` is `TRUE`,  add a check for altExps: All altExps of the same name must have the same features. If they do not, then error out that merging has failed.
	- Later, we will circle back to adding functionality to handle feature mis-match across altExps
- Let the current `do.call(cbind, sce_list)` code merge everything all at once, including altExps
	- This means that before merging, we need to make sure all altExps are compatible. Given that we are, for now, enforcing the same features, that should "just work."
- In main experiment side, rather than _subsetting_ to the intersection of shared features, we should expand the SCE to include the union of features. This will involve some code changes to `prepare_sce_for merge` - we'll want to create a new SCE entirely with all features as expected.
	- That said, we should probably still keep the check that the vector of intersecting features is not length 0. If there are no features are in common, that's a 🚩.

@jashapiro, please weigh in with any details I missed! 

One detail I know I'm missing is -  for the altExps check, do we want to enforce that all altExps are _present_ at this stage as well, or fill in with NA-matrices any that are missing? If we go with enforcing for now, we'll circle back and add dummies when we return to handle altexp feature mismatches.

jashapiro:
> * If `include_altexps` is `TRUE`,  add a check for altExps: All altExps of the same name must have the same features. If they do not, then error out that merging has failed.
>   
>   * Later, we will circle back to adding functionality to handle feature mis-match across altExps
> * Let the current `do.call(cbind, sce_list)` code merge everything all at once, including altExps
>   
[...]

PR #243, opened 2023-12-06
Add functionality to merge altExps

sjspielman:
Towards #150

This PR is _the first PR_ at updating the our SCE merging to consider alternative experiments. This process has been much more of a 🎢 than we maybe had imagined, so you can expect more PRs from here. At a minimum, there will be at least 1-2 more PR here for tests which are not yet implemented.

I also want to be cautious that this PR doesn't get too long, since the code changes here are also fairly extensive.  I think a good way to handle big reviews is to stack on this branch with separate PRs. Requesting a look from everybody since we may have to have some interesting discussions here! _In an effort to keep this PR from getting out of hand too quickly, please focus on leaving **high-level feedback** about the approach for your first round of review!_ Thanks!

## Ch-ch-ch-ch-changes

- Bonus spacing updates throughout for style, as well as relevant comment updates
- I added an argument `include_altexp` to the crux function `merge_sce_list()`.
  - This currently has a default of `TRUE`. Eh?
- Tests are only passing right now because I set `include_altexp=FALSE` in all relevant spots in the unit tests. I've updated some other parts in there to lay the groundwork for circling back and adding more tests, but we'll need 1-2 more PRs to get both i) current tests working with `include_altexp=TRUE`, and ii) tests specifically for the new altexp functions (read on...).
- The overall strategy I've taken is to handle main and alternative experiments _separately_ - we split out the altExps at the beginning. Then, do all main exp processing and merging as before. Finally at the end, we process and merge the altExps, and then pop it into the merged object's altExp.
  - Originally when working on this, I had the altExp processing happening before the SCE merge, but I kept hitting cryptic bugs, so I adopted this strategy of separate processing which seemed to be less mysteriously buggy!
- Altexp processing proceeds as follows, via two new functions: `prepare_altexps_for_merge()` and `update_altexp_assay()`.
  - I gave up kind of quickly tbh trying manipulate sparse matrices directly, so I ended up creating new SCEs with `NA` values added in for rowData and counts/logcounts assays (note some `TODO`s in the code about handling assays). The existing altExps are therefore replaced with these new ones. 
  - These fresh altexps are then passed through the existing function `prepare_sce_for_merge()` to keep things identifiable in the final merge


R/merge_sce_list.R
```diff
@@ -263,3 +302,131 @@ prepare_sce_for_merge <- function(sce,
...
+    #  either is missing. Any reviewer preferences?
+    new_counts <- update_altexp_assay(
+      counts(sce_altexp),
+      missing_features
+    )
+
+    new_logcounts <- update_altexp_assay(
+      logcounts(sce_altexp),
+      missing_features
+    )
```
allyhawkins, line 368: How about we just loop through the existing assays? First, get a list of all `assayNames` and then you can use `purrr::map` to apply the function to each assay. That should account for when we don't have both assays or if there's anything extra we aren't accounting for.

R/merge_sce_list.R
```diff
@@ -263,3 +302,131 @@ prepare_sce_for_merge <- function(sce,
...
+    altExp(sce) <- prepare_sce_for_merge(
+      new_altexp,
+      sce_name,
+      batch_column = "library_id",
+      cell_id_column = "cell_id",
+      shared_features = altexp_features,
+      retain_coldata_cols = colnames(colData(new_altexp)),
+      preserve_rowdata_cols = NA # rename them all
+    )
+  }
```
allyhawkins, line 397: this doesn't quite work. You haven't accounted for preparing the SCE if there are no missing features.

allyhawkins (commented):
I took a quick look, and I think I would approach this a little bit differently: 

- make a list of mainExp
- make a list of altExp 
- run both lists through the existing `prepare_sce_for_merge` 
  - within this function I would add an argument for `altExp = TRUE` (set to `FALSE` by default)
  - Then inside the existing `prepare_sce_for_merge` function you first check that all features are present in all SCE objects. If they are not shared and `altExp = TRUE` then you would run a separate function that prepares and modifies the assays specifically.  Something like:  `if (missing_features & altExp){ merge_assays }`. 
  - Before running `merge_assays` you could get a list of `assayNames` and then apply the `merge_assays` function to each assay. The last step would be creating the "new" altExp SCE before continuing with the rest of the `prepare_sce_for_merge` function (what's currently on line 372 - 385). 

Let me know what you think of that

R/merge_sce_list.R
```diff
@@ -80,6 +85,16 @@ merge_sce_list <- function(sce_list = list(),
      Please check that `retain_coldata_cols` was correctly specified.")
   }
 
+  # If we are including altExps, extract them now for separate merging, to be
+  #  added back into the final merged SCE at the end
+  if (include_altexp) {
+    altexp_list <- sce_list |>
+      purrr::map(altExp)
```
jashapiro, line 92: This will get the first `altExp`, so we are assuming that all the altExps across the merge are the same, which seems dangerous. We need probably want to use `altExpNames()` first to get the full list of altExps we can expect. Then we have to go through them one at a time. 

Alternatively, you could use `AltExps()` to get lists of altExps with names, but I think separating the tasks will be worth the effort.

R/merge_sce_list.R
```diff
@@ -263,3 +302,131 @@ prepare_sce_for_merge <- function(sce,
...
+        names_from = var,
+        values_from = val
+      ) |>
+      # repeat rows n_missing times
+      dplyr::slice(
+        rep(
+          1:dplyr::n(),
+          each = n_missing
+        )
+      )
```
jashapiro, line 349: I find this really opaque. Why not make a matrix with all `NA` and the expected row and column names/dimensions, then convert it to a data frame?

R/merge_sce_list.R
```diff
@@ -263,3 +302,131 @@ prepare_sce_for_merge <- function(sce,
...
+#' @return An updated SCE object with all altExp features present
+prepare_altexps_for_merge <- function(
+    sce,
+    sce_name,
+    altexp_features) {
+
+  sce_altexp <- altExp(sce)
+
+  # Determine which features are missing
+  missing_features <- altexp_features[!(altexp_features %in% rownames(sce_altexp))]
```
jashapiro, line 327: This common pattern can be shortened
```suggestion
  missing_features <- setdiff(altexp_features, rownames(sce_altexp))
```

R/merge_sce_list.R
```diff
@@ -263,3 +302,131 @@ prepare_sce_for_merge <- function(sce,
...
+  rownames(new_matrix) <- missing_features
+
+  # Now rbind it to the existing matrix
+  new_assay <- assay_sparse_matrix |>
+    as.matrix() |>
+    rbind(new_matrix) |>
+    # make it sparse again
+    as("CsparseMatrix")
+
+  return(new_assay)
```
jashapiro, line 430: I think we want to do this a bit differently, to be sure all of our matrices have the same row order, and it may simplify things too.

I think the below should work pretty directly if you pass in the full feature list.

```
# define the new full matrix
new_matrix <- matrix(nrow = length(all_features), ncol = ncol(assay_sparse_matrix), dimnames = list(all_features, rownames(assay_sparse_matrix))
# fill in existing features
new_matrix[rownames(assay_sparse_matrix),] <- matrix(assay_sparse_matrix)
```

jashapiro:
> because we only allow for 1 altexp at a time in a given SCE

I'll take minor issue with this statement: `scpca-nf` only supports 1 altExp at the moment, but `scpcaTools` has no such limitation! We can use `merge_altExp()` repeatedly to add as many altExps as we like.

Otherwise, I think your overall plan should be fine... To state it as I would see it:

1. Get a vector of all altexps from all libraries with `altExpNames() |> union()`
2. loop/walk through that list to make merged altExp objects and
3. join them them into the merged SCE.

One thing we haven't discussed, but in the merged SCEs, we probably want to update the `colnames` such that they are a combination of barcode and library_id if we are not doing that already, in order to keep them unique.

R/merge_sce_list.R
sjspielman: You never like my Rube-Goldberg machines 😞

sjspielman: 😉

sjspielman:
This is now ready for another look after a pretty decent overhaul. In the crux function `merge_sce_list()`, everything for altExps happens at or after the comment `## Handle altExps ---------`.  Here's the overall strategy:

1. Check for which altExpNames are present, and loop over each _name_ to prepare a merged altExp for each. These get saved into a list `merged_altexps`, which get added into the `merged_sce`. 
2. Within the loop, use the function `create_merged_altexp()` to create the merged altExp:
    - Determine the full set of features to include, as well as the full number of cells in the merged object.
    - Create an `NA` matrix with these dimensions and with the final row names (features) and column names (`{sce_name}-{barcode}`).
    - Determine which assays are present across all individual altexps. Using the function `build_new_altexp_assay()`,  create a new matrix that swaps in any existing matrices into the `NA` matrix. For any assay that wasn't present, it will be covered by the `NA` baseline in that matrix.


I also take care to ensure column names in the merged_sce and the merged_altexps _definitely match_. We do end up passing around an altExp list across some functions, but this is much smaller than the full SCE list at least and I didn't exactly see a way around it. Similarly, there are some for loops that I can't see a different way to handle. Any suggestions are welcome!


There are a couple remaining steps for subsequent PRs:
- Update tests (they are all passing locally for me now with `include_altexp = FALSE`)
- There is one `TODO` I left about getting rowData and colData into the 

Again, let's be mindful of this PR not getting too crazy, so when reviewing please suggest spots where you think stacking might be better than making changes directly in this PR! Hopefully we'll like the current overall strategy though 🤞

R/merge_sce_list.R
```diff
@@ -156,12 +166,71 @@ merge_sce_list <- function(sce_list = list(),
...
+        \(sce) altExpNames(sce)
+      ) |>
+      purrr::reduce(union)
+
+    for (altexp_name in altexp_names) {
+      # Determine which SCEs contain this altExp, and create list of those altExps
+      has_altexp_name <- sce_list |>
+        purrr::map_lgl(
+          \(sce, altexp_name) altexp_name %in% altExpNames(sce),
+          altexp_name
```
allyhawkins, line 196:
```suggestion
        purrr::map_lgl(
          \(sce) altexp_name %in% altExpNames(sce))
```
Maybe I'm just missing something, but I think you want to just return a vector of logicals with whether or not the `altexp_name` is present? So you don't want to return the `altexp_name`, since what you had will return it regardless of if it's present or not.

R/merge_sce_list.R
```diff
@@ -156,12 +166,71 @@ merge_sce_list <- function(sce_list = list(),
...
-  # replace existing metadata list with merged metadata
+  # Replace existing metadata list with merged metadata
   metadata(merged_sce) <- metadata_list
 
+  # Add the merged altE into the main merged_sce
+  if (include_altexp) {
+
+    # Ensure compatible column names
+    # (this is probably not necessary but doesn't hurt...)
+    merged_sce <- merged_sce[,merged_colnames]
```
allyhawkins, line 226: Do we want to subset here... I think if we are already grabbing all columns directly from the merged SCE, we don't need this or want this.

R/merge_sce_list.R
```diff
@@ -263,3 +335,99 @@ prepare_sce_for_merge <- function(sce,
...
+#'   set of features for this altExp
+#' @param merged_col_names Vector of matrix column names, corresponding to all cells
+#'   which will be in the final merged altExp
+#'
+#' @return Sparse matrix
+build_new_altexp_assay <- function(
+    assay_name,
+    altexp_list,
+    merged_row_names,
+    merged_col_names) {
```
allyhawkins, line 403:
```suggestion
    merged_rownames,
    merged_colnames) {
```
This is minor, but everywhere else you use `colnames`

R/merge_sce_list.R
```diff
@@ -156,12 +166,71 @@ merge_sce_list <- function(sce_list = list(),
...
+  # If we are including altExps, process them and save to list to add to merged SCE
+  merged_altexps <- list()
+  if (include_altexp) {
+
+    # First we need to determine the final column names of the merged_sce (not yet made)
+    #  for use in altExp code. Later we'll apply this order to the merged_sce itself.
+    merged_colnames <- sce_list |>
+      purrr::map(colnames) |>
+      unlist() |>
+      unname()
```
allyhawkins, line 181:
```suggestion
    all_merged_barcodes <- sce_list |>
      purrr::map(colnames) |>
      unlist() |>
      unname()
```
This really confused me when reviewing this. I thought you were grabbing the `colnames` to use to fill in the `colData`. Can we make this variable name clear that you are grabbing all the barcodes? I know that the columns have the batch ID and barcode, so maybe we just make that clear that's what your grabbing in the comment. Maybe also `all_merged_colnames` instead of `all_merged_barcodes`? This also could just be a me problem.

R/merge_sce_list.R
```diff
@@ -263,3 +335,99 @@ prepare_sce_for_merge <- function(sce,
...
+
+
+
+#' Build a new sparse matrix for merging altExps
+#'
+#' @param assay_name Name of assay of interest (e.g., "counts")
+#' @param altexp_list List of altExps which should be included in the new matrix
+#' @param merged_row_names Vector of matrix row names, corresponding to the full
+#'   set of features for this altExp
+#' @param merged_col_names Vector of matrix column names, corresponding to all cells
```
allyhawkins, line 395: Something like this comment but when you actually create the `merged_colnames` would be helpful!

R/merge_sce_list.R
```diff
@@ -263,3 +335,99 @@ prepare_sce_for_merge <- function(sce,
...
+      altexp_features,
+      merged_colnames
+    )
+  names(new_assays) <- altexp_assay_names
+
+
+  # Create merged altExp
+  merged_altexp <- SingleCellExperiment(assays = new_assays)
+
+  # TODO: Add rowData (and colData, if we need it) to merged_altexp
```
allyhawkins, line 380: Just noting that we should definitely do this, but I'm okay with a later PR.

allyhawkins (commented):
I didn't look at the changes to the tests you made yet, but I made a few smaller comments throughout the function changes. Generally, I'm good with this approach, I just need some helpful comments so that I remember that colnames of the merged SCE object correspond to cells 🤦‍♀️

R/merge_sce_list.R
```diff
@@ -156,12 +166,71 @@ merge_sce_list <- function(sce_list = list(),
...
+
+    for (altexp_name in altexp_names) {
+      # Determine which SCEs contain this altExp, and create list of those altExps
+      has_altexp_name <- sce_list |>
+        purrr::map_lgl(
+          \(sce, altexp_name) altexp_name %in% altExpNames(sce),
+          altexp_name
+        )
+      altexp_list <- sce_list[has_altexp_name] |>
+        purrr::map(altExp)
```
jashapiro, line 199: While my previous comment is still worth leaving, we can simplify!
```suggestion
      altexp_list <- sce_list |>
        purrr::keep(\(sce) altexp_name %in% altExpNames(sce))
        purrr::map(altExp)
```

R/merge_sce_list.R
sjspielman: This definitely is returning a logical vector... `altexp_name` on line 196 is the second argument to the "function".
How about I add braces like `\(sce, altexp_name) {altexp_name %in% altExpNames(sce)},` for clarity?
```
> has_altexp_name <- sce_list |>
+     purrr::map_lgl(
+         \(sce, altexp_name) altexp_name %in% altExpNames(sce),
+         altexp_name
+     )
> has_altexp_name
 sce1  sce2  sce3 
FALSE  TRUE  TRUE 
```

R/merge_sce_list.R
```diff
@@ -263,3 +335,99 @@ prepare_sce_for_merge <- function(sce,
...
+    dimnames = list(
+      merged_row_names,
+      merged_col_names
+    )
+  )
+
+  # Substitute existing assays into the matrix
+  # Note we need `sce_name` to reform column names
+  for (sce_name in names(altexp_list)) {
+    altexp <- altexp_list[[sce_name]]
```
jashapiro, line 419: I don't think you use `sce_name` in here? So maybe just:
```suggestion
  for (altexp in altexp_list) {
```

R/merge_sce_list.R
```diff
@@ -263,3 +335,99 @@ prepare_sce_for_merge <- function(sce,
...
+
+    # Add assay into matrix if it exists
+    # Note that column names were already formatted as `{sce_name}-{barcode}` by
+    #  the main SCE merging code
+    if (assay_name %in% assayNames(altexp)) {
+      new_matrix[rownames(altexp), colnames(altexp)] <- as.matrix( assay(altexp, assay_name) )
```
jashapiro, line 425: Is this `as.matrix()` required? I might expect implicit conversion to work.

R/merge_sce_list.R
```diff
@@ -237,24 +308,25 @@ prepare_sce_for_merge <- function(sce,
...
   sample_metadata <- metadata_list$sample_metadata
 
   # combine into one list
   metadata_list <- list(
     library_id = metadata(sce)[["library_id"]],
     sample_id = metadata(sce)[["sample_id"]],
-    library_metadata = library_metadata,
-    sample_metadata = sample_metadata
+    library_metadata = library_metadata, # this will be all previous metadata for the given library
+    sample_metadata = sample_metadata # this will be the same as the previous sample_metadata column
```
jashapiro, line 329: Not necessarily a column?
```suggestion
    sample_metadata = sample_metadata # this will be the same as the previous sample_metadata
```

jashapiro (commented):
A few small comments below, but I think the general plan looks good.

I see that in the meantime, Ally has put in some comments. 

I too want to go through the tests, but I will do that next...

R/merge_sce_list.R
jashapiro: see https://github.com/AlexsLemonade/scpcaTools/pull/243#discussion_r1423055684!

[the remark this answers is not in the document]
sjspielman: 😂 big day for me!!

R/merge_sce_list.R
sjspielman: I actually didn't check, but will!

R/merge_sce_list.R
sjspielman: > I don't think you use `sce_name` in here? So maybe just:

Ah, thanks. This was left over from when I still had to set `{sce_name}-{barcode}`. I added the comment for future us to this end on line 422, but didn't clean this part up!

R/merge_sce_list.R
sjspielman: > This also could just be a me problem.

Your confusion makes sense, will take suggestions!

R/merge_sce_list.R
sjspielman: This shouldn't be subsetting, although the syntax is the same. It's just ensuring it's all in the same _order_ as the altexp names. This is the same `all_merged_barcodes` variable that was confusing.

R/merge_sce_list.R
sjspielman: > Is this `as.matrix()` required?

yup.

sjspielman:
> I too want to go through the tests, but I will do that next...

FYI there isn't much new there. I only added some code to _set up for_ testing altexps, and I added `include_altexp=FALSE` to everything else! No new actual tests yet.

I expect my next steps will be 1) PR to add rowdata+coldata, and 2) PR for testing. Sound ok?

R/merge_sce_list.R
```diff
@@ -156,12 +166,67 @@ merge_sce_list <- function(sce_list = list(),
...
+
+  ## Handle altExps ------------------------------------------------------
+
+  # If we are including altExps, process them and save to list to add to merged SCE
+  merged_altexps <- list()
+  if (include_altexp) {
+
+    # First we need to determine the final column names of the merged_sce (not yet made)
+    #  for use in altExp code. Later we'll apply this order to the merged_sce itself.
+    # These values are cell ids: `{sce_name}_{barcode}`
```
allyhawkins, line 178: Thank you 😂

allyhawkins (approved):
LGTM!

R/merge_sce_list.R
sjspielman: 💯

jashapiro (commented):
So this looks good, but now I see that you only updated the tests that do not include altExps. 

I think it would also be useful to also include at least one test for what you expect the updated function to do. You can add edge cases in a subsequent PR, but there should be at least one test that shows that the code you included is working... You should also probably relabel the tests you updated to indicate that they are testing without altExps.

I might rearrange things to leave the tests as they were, starting with an sce_list with no altExp present (which should not require setting `include_altexp`), then add a separate section below to make an sce_list that _does_ have altExps and test with both `include_altexp = FALSE` and `include_altexp = TRUE`.

sjspielman:
I've rearrangement the tested code as suggested in review, and I added a couple tests - first a couple tests that `merge_sce_list()` works as expected when altExps are present, and another test that `include_altexp=FALSE` indeed doesn't return an altExp. Let me know if there are more tests you'd want to see at this stage!

tests/testthat/test-merge_sce_list.R
```diff
@@ -298,9 +303,124 @@ test_that("merging SCEs without names works as expected", {
...
+altexp_name <- "adt"
+sce_list_with_altexp <- sce_list_with_altexp |>
+  purrr::map(
+    \(sce) {
+      altExpNames(sce) <- altexp_name
+      sce # return the sce
+    }
+  )
+
+test_that("merging SCEs with matching features works as expected, with altexps", {
```
jashapiro, line 382: What do you mean by matching features here? There are missing features in 1, right?

tests/testthat/test-metadata_to_coldata.R
```diff
@@ -57,7 +57,8 @@ sce_list <- purrr::imap(
 # create merged sce
 merged_sce <- merge_sce_list(
   sce_list = sce_list,
-  retain_coldata_cols = "sum"
+  retain_coldata_cols = "sum",
+  include_altexp = FALSE
```
jashapiro, line 61: This should work without modification, right?
```suggestion
  retain_coldata_cols = "sum"
```

tests/testthat/test-merge_sce_list.R
```diff
@@ -298,9 +303,124 @@ test_that("merging SCEs without names works as expected", {
...
+    batch_column = batch_column,
+    # "total" should get removed
+    retain_coldata_cols = retain_coldata_cols,
+    # this row name should not be modified:
+    preserve_rowdata_cols = c("gene_names"),
+    include_altexp = FALSE
+  )
+
+  # this should fail if no altExp exists
+  expect_error( altExp(merged_sce) )
```
jashapiro, line 424: A slightly more specific test (other errors won't silently pass)
```suggestion
  # there should not be any altExps
  expect_length(altExpNames(merged_sce), 0)
```
```

File: tests/testthat/test-merge_sce_list.R [hunk 21]
```diff
@@ -41,23 +41,26 @@ shared_features <- rownames(sce)[1:10]
 retain_coldata_cols <- c("sum", "detected")
 preserve_rowdata_cols <- "gene_names"
 expected_coldata_cols <- sort(c("sum", "detected", batch_column, cell_id_column))
+num_altexp_features <- 5
 
-# Generate some shared data for testing `merge_sce_list()` ------
-sce1 <- sim_sce(n_cells = total_cells / 3, n_genes = total_genes, n_empty = 0)
+sce1 <-  sim_sce(n_cells = total_cells / 3, n_genes = total_genes, n_empty = 0)
```
jashapiro, line 46:
```suggestion
sce1 <- sim_sce(n_cells = total_cells / 3, n_genes = total_genes, n_empty = 0)
```

jashapiro (approved):
This looks good. I might add an issue for adding a few more tests for situations where there are multiple altExps, but this tests the basic functionality, which is kind of the minimal expectation. I also like that the previous tests are left mostly alone, which is a good thing for making sure you aren't introducing too many breaking changes (except directly related to the functionality you are adding). 

My suggestion below are pretty minimal, with the biggest one just being that you might as well set the altExp name when you create it rather than requiring a whole separate set of code. Other than that, I think this should be good to go.

tests/testthat/test-merge_sce_list.R
```diff
@@ -298,9 +302,154 @@ test_that("merging SCEs without names works as expected", {
...
+  # add metadata
+  library_id <- glue::glue("library-{batch}")
+  sample_id <- glue::glue("sample-{batch}")
+
+  metadata(sce_alt)$library_id <- library_id
+  metadata(sce_alt)$sample_id <- sample_id
+  metadata(sce_alt)$mapped_reads <- 100
+
+  # add sce_alt as sce's altExp, named "adt"
+  altExp(sce, "adt") <- sce_alt
```
jashapiro, line 355: Since you added the arg, you probably wanted to use it!
```suggestion
  # add sce_alt as sce's altExp
  altExp(sce, altexp_name) <- sce_alt
```

sjspielman: 🙄

PR #245, opened 2023-12-13
Add rowData to merged altexps

sjspielman:
Part of #150 

This PR adds the rowData slot into the merged altexp, in a manner than looks about like what you'd expect if Mr. Toad's Wild Ride were an R function. It works, but...I expect Reviews™️ about strategy 😬 

I added a function `build_new_altexp_rowdata()` which does what it says. The main reason this function is such a Wild Ride is because of code to handle the `preserve_rowdata_cols` argument; there will be some columns (notably, `target_type`) in the rowData which I figured we will not want to include for _each_ altExp, since the values are the same. If we feel it's ok to have several columns of the `{sce_name}-target_type` flavor, _a lot_ of this code can get simplified right off the bat! 

Strategy, for now:
1. Pass in a list of `rowData`s converted to `data.frame()` into the function
2. First, grab out the `final.final` version of each `preserve_rowdata_cols` we'd like to keep. The code is written flexibly enough to extend this default array in the future. The idea is, each rowData has the potential to have one of those columns, and we want to keep the comprehensive information from all of them "full joined." The list `preserved_columns` gets populated with the full columns to keep, and then turned into a dataframe for later `cbind`ing.
   - Currently I rely on the argument to `build_new_altexp_rowdata()` to define these columns, but we could move this up the chain and pass into the crux `merge_sce_list()` function.
3. Then, I define `new_rowdata`, considering all columns except the ones I pulled out into `preserved_columns`.
4. Finally, add back in the preserved columns if they exist, and `DataFrame()` the rowdata to ultimately add it back into the merged altexp. 

Finally, I added a test to make sure the rowData has the expected names, and it is passing, so that's something at least!

Please enjoy the 🎢 and give me all your thoughts on how to make it less of a 🎢 ...!


jashapiro (commented):
There are a few places here where I think some unintended consequences can sneak in, so I think I would take a different approach, possibly with less tidyverse...

The first task is getting the full list of columns so you know what the final dimensions and order will be. Currently you do this with a full join, but that can change order, so your final renaming might have some funny consequences, and we don't want that. And you already have the full list of features being passed in, so there isn't really a need for any of that!

Then, I think it is worth looking at how we did this in `prepare_sce_for_merge()`: we use the `preserve_rowdata_cols` to decide whether to rename for each rowData dataframe, adjusting the column names using an `ifelse()`. 

The next task is expanding the data frame to the full table, filling in any missing features with NA, which is pretty straightforward.

Once you have a list of tables with the same rows and columns that are specific to each altexp, you can pull off and merge the preserved columns, then cbind on all the others. 


Looking back, part of me thinks we might be able to get most of this to work with only a slight modification to `prepare_sce_for_merge()` and `merge_sce_list()`, or even using those functions directly (with a wrapper). Once we have a list of altExps of the same type, they are just SCEs with fewer features, so I don't expect that there should be any real differences except that we then need to add some rows for samples without a given altExp. But I wonder if we can do _that_ by making SCE objects for each missing altExp with no actual assay data in them, just `rowData` and `colData` corresponding to the cells and features we would see in the other altExps.

R/merge_sce_list.R
```diff
@@ -424,3 +429,138 @@ build_new_altexp_assay <- function(
...
+        # add this number of new rows with all NA values
+        if (length(missing_features) > 0) {
+          rowdata[nrow(rowdata) + length(missing_features),] <- NA
+        }
+
+        # name those rows
+        rownames(rowdata) <- c(present_features, missing_features)
+
+        # return updated rowdata
+        rowdata
```
jashapiro, line 544: Note that this will not be in the same order as `all_merged_features`, which is a problem.

sjspielman:
After some offline discussion, we're going to back up a few steps and come back to rowData soon. Closing this PR in favor of the different incoming strategies.

PR #246, opened 2023-12-14
Add an option to compress HDF5 files

### LINKED ISSUE #244: Support compression in `sce_to_anndata()`

We currently write out AnnData files without any compression. We should pass arguments for compression to `zellkonverter::writeH5AD()` 

We can add a `...` to the call that passes any additional arguments down to zellkonverter::writeH5AD

We could add a add a `compression` argument that we pass directly to the `compression` argument, rather than the catchall, but I  like the fact that this gives us free support for other `writeH5AD()` options.

allyhawkins:
Closes #244 

This PR adds new options to the `sce_to_anndata()` function to allow for compression of the exported HDF5 files. 

- I added a `compression` option and set the default to `gzip`. I like the idea of having a separate argument rather than just using `...`, partially as a reminder. I did not use the same default of `none` that the `zellkonverter::writeH5AD` function uses. Instead, I set it to `gzip` since I think we want to compress by default. I'm on the fence on whether that was a good decision. 
- I added `...` as an option to pass any additional arguments to the function. 
- I updated the tests to test that compression fails with an invalid option and a test to make sure using additional arguments works. Let me know if I should create any additional tests. 
- I also allowed for the `.h5ad` extension.

### DIFF
--- a/R/sce_to_anndata.R
+++ b/R/sce_to_anndata.R
@@ -5,7 +5,7 @@
 #' @param x_assay_name Name of assay in SCE object to save as X in AnnData. Default is "counts".
 #' @param compression Type of compression to use when exporting hdf5 files. Options are
 #'   "none", "gzip", or "lzf". Default is "gzip".
-#' @param ... Any arguments to be passed into zellkonverter::writeH5AD or zellkonverter::SCE2AnnData
+#' @param ... Any arguments to be passed into zellkonverter::writeH5AD
 #'
 #' @return original SingleCellExperiment object used as input (invisibly)
 #' **Note that any columns present in the `rowData` of an SCE object that contains
@@ -23,7 +23,13 @@
 #'   anndata_file = "test_anndata.h5"
 #' )
 #' }
-sce_to_anndata <- function(sce, anndata_file, x_assay_name = "counts", compression = "gzip", ...) {
+sce_to_anndata <- function(
+  sce, 
+  anndata_file, 
+  x_assay_name = "counts", 
+  compression = c("gzip", "none", "lzf"), 
+  ...
+) {
   if (!requireNamespace("zellkonverter", quietly = TRUE)) {
     stop("The zellkonverter package must be installed to convert objects to AnnData. No output file written.")
   }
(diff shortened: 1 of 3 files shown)


R/sce_to_anndata.R
```diff
@@ -1,8 +1,11 @@
 #' Convert SingleCellExperiment objects to AnnData file stored as HDF5 file
 #'
 #' @param sce SingleCellExperiment object to be converted to AnnData as an HDF5 file
-#' @param anndata_file Path to output AnnData file. Must be an `.h5` or `.hdf5`
+#' @param anndata_file Path to output AnnData file. Must end in `.h5`, `.hdf5`, or `.h5ad`
 #' @param x_assay_name Name of assay in SCE object to save as X in AnnData. Default is "counts".
+#' @param compression Type of compression to use when exporting hdf5 files. Options are
+#'   "none", "gzip", or "lzf". Default is "gzip".
+#' @param ... Any arguments to be passed into zellkonverter::writeH5AD or zellkonverter::SCE2AnnData
```
jashapiro, line 8: I know this gets passed internally to `SCE2AnnData`, but we don't use it, so it seems confusing to name it here.
```suggestion
#' @param ... Any arguments to be passed into zellkonverter::writeH5AD
```

R/sce_to_anndata.R
```diff
@@ -20,7 +23,7 @@
 #'   anndata_file = "test_anndata.h5"
 #' )
 #' }
-sce_to_anndata <- function(sce, anndata_file, x_assay_name = "counts") {
+sce_to_anndata <- function(sce, anndata_file, x_assay_name = "counts", compression = "gzip", ...) {
```
jashapiro, line 26: Updating spacing, and preparing for more formal argument checking.
```suggestion
sce_to_anndata <- function(
  sce, 
  anndata_file, 
  x_assay_name = "counts", 
  compression = c("gzip", "none", "lzf"), 
  ...
) {
```

R/sce_to_anndata.R
```diff
@@ -38,15 +41,19 @@ sce_to_anndata <- function(sce, anndata_file, x_assay_name = "counts") {
...
   }
 
   # make sure assay is found in sce object
   if (!x_assay_name %in% assayNames(sce)) {
     stop("`x_assay_name` is not an assay in `sce`")
   }
 
+  if(!compression %in% c("none", "gzip", "lzf")){
+    stop("`compression` must be one of `none`, `gzip`, or `lzf`.")
+  }
```
jashapiro, line 55: argument check
```suggestion
  compression <- match.arg(compression)
```

jashapiro (approved):
Just a couple little comments, including a reminder about `match.arg`, which is built for situations like this!

Issue #250, opened 2023-12-19
Fix merging metadata bug


sjspielman:
I'm pretty sure we have a bug in how we handle merging SCE metadata, such that we may have accidentally ended up with a lot of duplicate items. I have not dug into this much, but noticed that something looked funky. 

This issue therefore encompasses formally diagnosing the bug, and then fixing it.

allyhawkins:
What do you mean we have a lot of duplicate items?

sjspielman:
> What do you mean we have a lot of duplicate items?

I thought I was seeing too many `library_id` and/or `sample_id` fields that weren't properly nested inside the libraries from whence they came. But, I am jumping around quite a bit so I didn't really dig it, and I am also not 100% convinced I didn't just misread some output! So, this issue is for me to figure out whether I saw a real thing or not later... 🙃

PR #251, opened 2023-12-19
Update altexp merging

### LINKED ISSUE #150: Account for altExps when merging SCEs

sjspielman:
Closes #150 

We're getting close I think! Noting that I've set this to close out #150, since I think we should open further issues to tackle potential (likely?) next steps (e.g. #250 and/or doing more things with altExp rowdata/coldata/metadata).

First, some good news: It appears that we do not need to create dummy rowData or colData for any new altExps that get created, which is excellent. Notably, I've reached a point where I seriously need to refresh my eyeballs from this code, but I am 100% sure we would benefit from additional tests. Please let me know what test low-hanging-fruit I've missed! I know we still need a test for SCEs with multiple altexps, but wanted to get this out for review now since, as mentioned, my eyes need a break before writing more things!

- Note one recreational revision where I moved some metadata code into its own function because the script was getting a tad unwieldy. 
- I added a function `check_altexps()` to check for altExp compatibility and die otherwise. This function also returns a list which contains the expected features and assays that need to go into any new altExps that get created, and gets used down the line. Seemed prudent to just save this information while I was checking altExps anyways.
  - I left a `TODO` in there for a future-PR next step of subsetting assays rather than dying if any are missing.
- The function `prepare_altexp_for_merge()` does the altExp preparation, and it calls two other functions - 
  - ` build_na_matrix()` (called as needed for any missing altExps)
  -  `prepare_sce_for_merge()`. I added an argument to this function `is_altexp` with a default of `FALSE`. When we process altExps, this gets set to `TRUE`. This argument controls whether sce colnames should get a name prefix added; we don't want to do this if altExp is true since we'd end up with duplicate name prefixes in the end. I made this argument more general in case there are other aspects of this function we want to skip for altExps. 

Edit - one additional note for review. You probably want to use `split` (not `unified`) view, since git's detection of diffs is a yet another 🎢 .


tests/testthat/test-merge_sce_list.R
```diff
@@ -417,39 +498,61 @@ test_that("merging SCEs with different altExp features works as expected, with a
...
+  rows <- letters[1:5]
+  cols  <- letters[6:10]
+  sparse_mat <- build_na_matrix(
+    "name",
+    rows,
+    cols
+  )
+
+  expect_true(
+    class(sparse_mat) == "lgeMatrix"
```
jashapiro, line 519: This would be the wrong matrix type, as it is a matrix of logicals. We don't really want that! We can just check that it is sparse, which we probably do want, (earlier comment)
```suggestion
    class(sparse_mat) == "sparseMatrix"
```

tests/testthat/test-merge_sce_list.R
```diff
@@ -366,12 +383,58 @@ sce_list_with_altexp <- sce_list |>
...
+  )
+
+  # rowdata names should start with test-, but NOT the overall column names
+  expect_true(
+    all(stringr::str_starts(colnames(rowData(prepared_altexp)), "test-"))
+  )
+
+  expect_true(
+    !all(stringr::str_starts(colnames(prepared_altexp), "test-"))
+  )
```
jashapiro, line 412:
```suggestion
  # column names should be unchanged
  expect_equal(
    colnames(prepared_altexp), colnames(test_altexp)
  )
```

tests/testthat/test-merge_sce_list.R
```diff
@@ -366,12 +383,58 @@ sce_list_with_altexp <- sce_list |>
...
+    full_altexp_features,
+    retain_coldata_cols = NULL,
+    preserve_rowdata_cols = "target_type",
+    is_altexp = TRUE
+  )
+
+  # rowdata names should start with test-, but NOT the overall column names
+  expect_true(
+    all(stringr::str_starts(colnames(rowData(prepared_altexp)), "test-"))
+  )
```
jashapiro, line 408: Why not check the actual values here?
```suggestion
  expect_equal(
    colnames(rowData(prepared_altexp))
    paste0("test-", colnames(rowData(test_altexp))
  )
```

But this does not check whether we are correctly handling `preserve_rowdata_cols = "target_type"`, which we probably should... But this input doesn't have that rowdata column, so other things would have to change to check that too

tests/testthat/test-merge_sce_list.R
```diff
@@ -402,11 +465,29 @@ test_that("merging SCEs with same altExp features works as expected, with altexp
 
 
 
-test_that("merging SCEs with different altExp features works as expected, with altexps", {
+test_that("merging SCEs with 1 altexp but different features fails as expected, with altexps", {
 
-  # keep only the first 3 features from the first SCE
-  altExp(sce_list_with_altexp[[1]]) <- altExp(sce_list_with_altexp[[1]])[1:3]
+  # second list where 1 is missing all
```
jashapiro, line 470: I liked the previous comment better!
```suggestion
  # keep only the first 3 features from the first SCE
```

tests/testthat/test-merge_sce_list.R
```diff
@@ -417,39 +498,61 @@ test_that("merging SCEs with different altExp features works as expected, with a
     preserve_rowdata_cols = c("gene_names")
   )
 
-  merged_altexp <- altExp(merged_sce)
+  expect_true(altExpNames(merged_sce) == "adt")
```
jashapiro, line 501:
```suggestion
  expect_equal(altExpNames(merged_sce), "adt")
```

jashapiro, line 501: I think I would also add a few more tests here that the altExp matrices are correct: Are the values from `sce4` all NA? Are the values from other sces the same as what went in?
Maybe also check that the dimensions are what is expected?

tests/testthat/test-merge_sce_list.R
```diff
@@ -417,39 +498,61 @@ test_that("merging SCEs with different altExp features works as expected, with a
...
+
+  expect_equal(
+    colnames(sparse_mat), cols
+  )
 
 })
 
 
-test_that("merging SCEs with altExps works as expected when include_altexps = FALSE", {
+test_that("check_altexps passes when it should pass", {
```
jashapiro, line 533: Just noting that these should change with the function name change that I would like to see.

jashapiro (commented):
This looks pretty good!

I had some not huge suggestions, including a few extra tests. 

The one bigger thought is about how we should handle the `altexp_{name}_{stat}` columns in the merge. I think we want to retain them, but we need to add some kind of logic there to get the correct list. Building from `altexp_attributes` this should not be too bad... something like:
```
tidyr::expand_grid(
  name = names(altexp_attributes), 
  stat = c("sum", "detected", "percent")
) |> 
  glue::glue_data("altexp_{name}_{stat}")
```

I reserve the right to have more thoughts later, but I think we are on the right track.

sjspielman:
> I reserve the right to have more thoughts later, but I think we are on the right track.

🎉 I'll take it!

R/merge_sce_list.R
```diff
@@ -319,108 +377,114 @@ prepare_sce_for_merge <- function(
...
-  # TODO: Add rowData and colData to merged_altexp
+  # Attribute list to save for later use
+  altexp_attributes <- list()
 
-  return(merged_altexp)
+  # Find all altExp names present in the SCE objects.
+  altexp_names <- sce_list |>
+    purrr::map(
+      \(sce) altExpNames(sce)
+    ) |>
```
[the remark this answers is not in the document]
sjspielman: lol this was a good one 😂

R/merge_sce_list.R
```diff
@@ -301,11 +284,86 @@ prepare_sce_for_merge <- function(
...
+  }
+
+  # Now, prepare this altexp for merge
+  altExp(sce, altexp_name) <- prepare_sce_for_merge(
+    altExp(sce, altexp_name),
+    sce_name,
+    batch_column = batch_column,
+    cell_id_column = cell_id_column,
+    shared_features = expected_features,
+    retain_coldata_cols = NULL, # Is this right?
```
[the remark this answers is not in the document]
sjspielman: > However, I wonder if we should add in altexps_{name}_sum altexps_{name}_detected and altexps_{name}_percent for the full SCEs when we prepare those... if we are including altExps.

I think this is right right move, to keep them in the "top-level" SCE.

tests/testthat/test-merge_sce_list.R
```diff
@@ -417,39 +498,61 @@ test_that("merging SCEs with different altExp features works as expected, with a
...
+  cols  <- letters[6:10]
+  sparse_mat <- build_na_matrix(
+    "name",
+    rows,
+    cols
+  )
+
+  expect_true(
+    class(sparse_mat) == "dgCMatrix"
+  )
```
jashapiro, line 520: My bad here: in general it is better to use `is` for these tests, as we want to allow flexibility from parent classes. the `Matrix` constructors can choose among different classes, and we don't really care which specific class uses most of the time.

```suggestion
  expect_true(
    is(sparse_mat, "sparseMatrix")
  )
```

But even better: 
```suggestion
  expect_s4_class(sparse_mat, "sparseMatrix")
```

sjspielman: I seem to have landed on `expect_s4_class(sparse_mat, "dgCMatrix")`?

```
expect_s4_class(sparse_mat, "sparseMatrix") is not TRUE

`actual` is a double vector (NA, NA, NA, NA, NA, ...)
`expected` is a logical vector (TRUE)
```

sjspielman: Sticking with ` is(sparse_mat, "dgCMatrix")` for now.

jashapiro: Note my second suggestion _replaces_ the `expect_true()` and does not nest in it.

sjspielman: https://github.com/AlexsLemonade/scpcaTools/pull/251/commits/4f61572b84623b6ed19f7108fc44ec447360213f ✅

R/merge_sce_list.R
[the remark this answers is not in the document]
sjspielman: See here: https://github.com/AlexsLemonade/scpcaTools/pull/251/commits/955409e1f254497e01bf825399852e1a0ee4b3c6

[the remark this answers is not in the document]
sjspielman: ....and https://github.com/AlexsLemonade/scpcaTools/pull/251/commits/b1a0091d73e48901756dc33c4618769ef7a74bb2

sjspielman:
I am totally stuck on a failing test that I cannot figure out after https://github.com/AlexsLemonade/scpcaTools/pull/251/commits/bc6f82e1a2ec036f2c4c17d2fbb096fb9fda831c - 

```
── Error (test-merge_sce_list.R:563:3): merging SCEs where 1 altExp is missing works as expected, with altexps ──
Error in `value[[3L]](cond)`: failed to combine 'int_colData' in 'cbind(<SingleCellExperiment>)':
  failed to rbind column 'altExps' across DataFrame objects: failed to
  rbind column 'adt' across DataFrame objects: subscript contains invalid
```

I've tracked down the specific error back to this part of the `combine` code - 
```
[tryCatch](https://rdrr.io/r/base/conditions.html)({
        int_cd <- [do.call](https://rdrr.io/r/base/do.call.html)([rbind](https://rdrr.io/r/base/cbind.html), [lapply](https://rdrr.io/r/base/lapply.html)([args](https://rdrr.io/r/base/args.html), [int_colData](https://rdrr.io/bioc/SingleCellExperiment/man/internals.html)))
    }, error=[function](https://rdrr.io/r/base/function.html)(err) {
        [stop](https://rdrr.io/r/base/stop.html)(
            "failed to combine 'int_colData' in 'cbind(<", [class](https://rdrr.io/r/base/class.html)([args](https://rdrr.io/r/base/args.html)[[](https://rdrr.io/r/base/Extract.html)[1]]), ">)':\n  ",
            [conditionMessage](https://rdrr.io/r/base/conditions.html)(err))
    })
```
But I can't figure out why this error is being thrown. I _am_ able to run `rbind()` directly on the `colData` slots, and there aren't any columns here called `"adt"` or `"altExps"`... Time for me to take a little break from this code and circle back fresh. Any help spelunking to track down why this is being thrown is appreciated too.

sjspielman:
> Obviously we are missing something... But that is why we have the tests! I can look at it tomorrow if you need a break.
> Since this is one of the main use cases we expect to see in real data (some libraries missing altExps), this is important to get right.

💯 all around. I tried more and the "edit" I added in that comment above blew my mind, so I do need to actually take that break today...!

tests/testthat/test-merge_sce_list.R
```diff
@@ -397,59 +469,82 @@ test_that("merging SCEs with same altExp features works as expected, with altexp
...
+    "altexps_adt_percent",
+    batch_column,
+    "cell_id"
+  )
 
+  expect_true(
+    setequal(
+      colnames(colData(merged_sce)),
+      expected_coldata
+    )
```
jashapiro, line 487:
```suggestion
  expect_setequal(
    colnames(colData(merged_sce)),
    expected_coldata
```

tests/testthat/test-merge_sce_list.R
```diff
@@ -397,59 +469,82 @@ test_that("merging SCEs with same altExp features works as expected, with altexp
...
 
-})
 
+test_that("merging SCEs where 1 altExp is missing works as expected, with altexps", {
+  sce_list_with_altexp[["sce4"]] <- removeAltExps(sce_list_with_altexp[[1]])
 
-test_that("merging SCEs with altExps works as expected when include_altexps = FALSE", {
+  # from cbind docs:
+  # The colnames in colData(SummarizedExperiment) must match or an error is thrown.
+  # Duplicate columns of rowData(SummarizedExperiment) must contain the same data.
```
jashapiro, line 519: These comments should probably be updated.

tests/testthat/test-merge_sce_list.R
```diff
@@ -397,59 +469,82 @@ test_that("merging SCEs with same altExp features works as expected, with altexp
...
     retain_coldata_cols = retain_coldata_cols,
     # this row name should not be modified:
-    preserve_rowdata_cols = c("gene_names"),
-    include_altexp = FALSE
+    preserve_rowdata_cols = c("gene_names")
   )
 
-  # there should not be any altExps
-  expect_length(altExpNames(merged_sce), 0)
+  expect_equal(altExpNames(merged_sce), "adt")
```
jashapiro, line 530: Let's add in a few more tests here:
Check that the number of columns is what we expect, for the main and altexps
Check that the colData columns are still as expected?
(same checks as when all were present, basically)

Check that the missing values is all `NA`, perhaps? (Pull out the values where the library is "sce4"?)

jashapiro (commented):
Okay, assuming this is passing tests, because I am confident, I think we still want a few more tests for the missing altExp case, basically to match the complete altExp case.

We can go one step further to check that the filled in values are what we expected (NA) too.

tests/testthat/test-merge_sce_list.R
```diff
@@ -397,59 +469,82 @@ test_that("merging SCEs with same altExp features works as expected, with altexp
...
-      \(sce, sce_name) glue::glue("{sce_name}-{colnames(sce)}")
-    ) |>
-    unlist() |>
-    unname()
-  expect_equal(colnames(merged_altexp), expected_colnames)
 
-})
 
+test_that("merging SCEs where 1 altExp is missing works as expected, with altexps", {
+  sce_list_with_altexp[["sce4"]] <- removeAltExps(sce_list_with_altexp[[1]])
```
jashapiro, line 515: Oh, I meant to suggest a different (even easier!) approach here, because this leaves in the altExp stats in the colData. 
```suggestion
  sce_list_with_altexp[["sce4"]] <- sce_list[[1]]
```

sjspielman:
Tests added here! https://github.com/AlexsLemonade/scpcaTools/pull/251/commits/2f12c6cc7f652410d44178bb0ad99d7119821498

tests/testthat/test-merge_sce_list.R
```diff
@@ -526,6 +523,39 @@ test_that("merging SCEs where 1 altExp is missing works as expected, with altexp
...
+    expected_coldata
+  )
+
+  # check that the NAs are as expected
+  counts_mat <- as.matrix(counts(merged_altexp))
+  sce4_counts <- counts_mat[, stringr::str_starts(colnames(counts_mat), "sce4-")]
+  expect_true(
+    all(is.na(sce4_counts))
+  )
+  numeric_counts <- counts_mat[, !stringr::str_starts(colnames(counts_mat), "sce4-")]
```
jashapiro, line 555: Similar to the above
```suggestion
  numeric_counts <- counts_mat[, merged_sce$library_id != "sce4"]
```

tests/testthat/test-merge_sce_list.R
```diff
@@ -526,6 +523,39 @@ test_that("merging SCEs where 1 altExp is missing works as expected, with altexp
...
+  expect_true(
+    all(is.numeric(numeric_counts))
```
jashapiro, line 557: `is.numeric(NA_real_)` is TRUE, so this is a bit dangerous for the test you want here!
`is.finite()` is usually safer.
```suggestion
    all(is.finite(numeric_counts))
```

tests/testthat/test-merge_sce_list.R
```diff
@@ -526,6 +523,39 @@ test_that("merging SCEs where 1 altExp is missing works as expected, with altexp
...
+    batch_column,
+    cell_id_column
+  )
+  expect_setequal(
+    colnames(colData(merged_sce)),
+    expected_coldata
+  )
+
+  # check that the NAs are as expected
+  counts_mat <- as.matrix(counts(merged_altexp))
```
jashapiro, line 550: The conversion isn't hurting anything, but it shouldn't be needed
```suggestion
  counts_mat <- counts(merged_altexp)
```

jashapiro (approved):
LGTM... just a few suggestions on the added tests.

sjspielman:
Noting https://github.com/AlexsLemonade/scpcaTools/pull/251/commits/e4a7399ea4c2d686e13b0fc02e3b8371c008b8ff is passing locally!

PR #252, opened 2023-12-20
Make sure all sample metadata columns are compatible before merging

allyhawkins:
I was noticing a recurring error when trying to run the merged workflow with the latest version of scpcaTools. Specifically I was getting the following error about the `age` columns: 

```
Error in (function (cond)  : 
  error in evaluating the argument 'x' in selecting a method for function 'unique': Can't combine `..1$age` <double> and `..2$age` <character>.
```

In digging into this a little more, I was working with two processed objects that had differing types of `age` columns. One had `character` and the other had `double`. This is probably because we updated how we deal with the `age` column in https://github.com/AlexsLemonade/ScPCA-admin/pull/722. To avoid any future conflicts between the types of columns, I am converting them all to characters before creating the combined sample metadata. 

I checked with what's going on in #251 and it looks like this section hasn't been touched, so it shouldn't cause any conflicts.

### DIFF
--- a/R/merge_sce_list.R
+++ b/R/merge_sce_list.R
@@ -154,6 +154,11 @@ merge_sce_list <- function(
   # if object has sample metadata then combine into a single data frame
   if ("sample_metadata" %in% names(metadata_list)) {
     sample_metadata <- metadata_list$sample_metadata |>
+      purrr::map(\(df) {
+        # make sure all column types are compatible first
+        df <- df |>
+          dplyr::mutate(across(everything(), as.character))
+        }) |>
       dplyr::bind_rows() |>
       unique()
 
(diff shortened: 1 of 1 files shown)


sjspielman:
FYI I think (but am not sure precisely) something is might funky with metadata that I was planning to circle back to after dealing with merging altexps. The funkinesss I was seeing is not what you found here, though! https://github.com/AlexsLemonade/scpcaTools/issues/250

allyhawkins:
I'll go ahead and switch to request @jashapiro for review. But for some added context, all the other columns in the sample metadata are already characters. The only time we will have incompatibility is when we are merging SCE's that were processed before a few weeks ago with recently processed SCE objects. In that scenario one will have an age column that's a character and the other will have a double.

jashapiro:
The changes in https://github.com/AlexsLemonade/ScPCA-admin/pull/722 shouldn't affect this, as that code is just to create the text files that are getting read in by scpca-nf. So something must have changed in `scpca-nf` or `scpcaTools` to generate this mismatch. My guess is it is in a place where we read in the metadata with `readr` and I probably added a default column type to quiet some messages. Since we never perform any math on the metadata, leaving it as characters is probably the safe thing to do.

jashapiro (approved):
lgtm

Issue #248, opened 2023-12-18
Allow for altExps with different features 


sjspielman:
As part of #150, we are enforcing that altExps of the same name (e.g., `"adt"`) all have the same features during merging. In the future, we would like to allow for feature mismatch to merge altExps in cases where not all features will necessarily present in all altExps of that given name.

sjspielman:
Something else to think about when the time comes, from https://github.com/AlexsLemonade/scpcaTools/pull/253#issue-2052615164:

> I kept failures for different features, as I think that we may want to discuss what to do in the case of altExps with the same names but different feature sets: do we combine those or turn them into separately named altExps? I don't know yet, so failure seems the best option.

PR #253, opened 2023-12-21
Use combineCols for merging SCEs

jashapiro:
Stacked on #251, to hopefully fix the errors there.

So after fighting a bunch to try to see what we were missing, including some diving into the `cbind` code for SCEs, I decided to try a different approach, and let the developers of SCE handle our problem.

Turns out they did a pretty good job. Using there combineCols function made things work and removed the need for us to create our own NA matrices!

This also removes the need for us to test that we have all the assays, as those get combined correctly. I kept failures for different features, as I think that we may want to discuss what to do in the case of altExps with the same names but different feature sets: do we combine those or turn them into separately named altExps? I don't know yet, so failure seems the best option.

So this code does that, along with some linting/reformatting that precommit wanted.

Locally, this is passing all tests, but we will get a final test when merging into #251!


sjspielman (approved):
This all looks fine to me, let's see how it goes!

>  I kept failures for different features, as I think that we may want to discuss what to do in the case of altExps with the same names but different feature sets: do we combine those or turn them into separately named altExps? I don't know yet, so failure seems the best option.

FYI  -https://github.com/AlexsLemonade/scpcaTools/issues/248#issuecomment-1866515304

> So this code does that, along with some linting/reformatting that precommit wanted.

Thank you for the reminder to locally set up my precommit hook in _this_ repo!

PR #256, opened 2023-12-22
remove python from renv tests

jashapiro:
We don't use python in our scpcaTools tests; it is only there for other uses in the dockerfile, so I think we can safely remove it before restoring with renv.

Should hopefully solve the hash mismatch problems we were seeing. And speed up testing a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,6 @@
 
   <!-- badges: start -->
   [![R-CMD-check](https://github.com/AlexsLemonade/scpcaTools/workflows/R-CMD-check/badge.svg)](https://github.com/AlexsLemonade/scpcaTools/actions/workflows/R-CMD-check-current.yaml)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -3,7 +3,6 @@
 
   <!-- badges: start -->
   [![R-CMD-check](https://github.com/AlexsLemonade/scpcaTools/workflows/R-CMD-check/badge.svg)](https://github.com/AlexsLemonade/scpcaTools/actions/workflows/R-CMD-check-current.yaml)
-  [![R-CMD-check-renv](https://github.com/AlexsLemonade/scpcaTools/workflows/R-CMD-check-renv/badge.svg)](https://github.com/AlexsLemonade/scpcaTools/actions/workflows/R-CMD-check-renv.yaml)
```
sjspielman, line 6: 🎢

sjspielman (approved):
👍

PR #259, opened 2024-01-17
Keep celltype annotation columns when merging SCEs

### LINKED ISSUE #258: Ensure cell type labels are retained when merging SCEs
When merging SCEs in `scpcaTools::merge_sce_list()`, we use a vector of colData columns to retain. Missing from this list are cell type annotation columns, but we would definitely like to retain them. We'll need to update the function, and its docs, accordingly.

sjspielman:
Closes #258 

This PR updates SCE merging to ensure that any cell type annotation columns are retained in the `colData`. 

In our current implementation, we pass in a vector `retain_coldata_cols` to keep any colData columns we want kept. But, we also _create these columns_ if they are missing and populate them with `NA`s. Therefore, one does not simply add the cell type columns into this vector. If there are no cell type columns, we do not want to end up creating them!

So, I created a new argument in `merge_sce_list` `celltype_coldata_cols` with a default value of all the possible cell type columns that one might encounter in the `colData` slot. This argument gets used in the the subfunction `prepare_merge_sce`, which has a default of `NULL` for this argument. Then:
- If it's `NULL`, just ignore it
- Otherwise, _after filling in the missing `retain_coldata_cols` with `NA`_, tack on the celltype columns which actually exist into `retain_coldata_cols`. Use this full vector to subset `colData`.

I updated some docs along the way, and I added some tests ✅ 

A handful of recreational revisions in the tests as a 🍬 too - stop using `sort()` to compare vectors and start using `test_setequal`.


sjspielman:
I'll also note - the other option here, rather than adding this argument, is to change how we handle `retain_coldata_cols`. Rather than creating & filling in missing columns with `NA`, we could just ignore them. If we update in that approach, we can then make this parallel change in `scpca-nf` as described here: https://github.com/AlexsLemonade/scpcaTools/issues/258#issuecomment-1895922396

allyhawkins:
> In our current implementation, we pass in a vector retain_coldata_cols to keep any colData columns we want kept. But, we also create these columns if they are missing and populate them with NAs. Therefore, one does not simply add the cell type columns into this vector. If there are no cell type columns, we do not want to end up creating them!

I still don't think we need to change anything in the function. I would change the vector that actually gets passed in to the function in `scpca-nf` to only include the cell type columns that are present. Otherwise you don't include them and they won't get created. That makes it so you don't have anything to do here.

sjspielman:
> I would change the vector that actually gets passed in to the function in scpca-nf to only include the cell type columns that are present. Otherwise you don't include them and they won't get created. That makes it so you don't have anything to do here.

I still don't think this is true. For example, we might have the situation where _some but not all_ of the libraries in a given project have cell types. If that were to happen, we'd end up with `NA` cell types. This has most potential to happen I think for submitters.

jashapiro:
> > In our current implementation, we pass in a vector retain_coldata_cols to keep any colData columns we want kept. But, we also create these columns if they are missing and populate them with NAs. Therefore, one does not simply add the cell type columns into this vector. If there are no cell type columns, we do not want to end up creating them!
> 
> I still don't think we need to change anything in the function. I would change the vector that actually gets passed in to the function in `scpca-nf` to only include the cell type columns that are present. Otherwise you don't include them and they won't get created. That makes it so you don't have anything to do here.

I agree with this. In general we should make changes to that add arguments (and complexity) with great caution.

allyhawkins:
> I still don't think this is true. For example, we might have the situation where some but not all of the libraries in a given project have cell types. If that were to happen, we'd end up with NA cell types. This has most potential to happen I think for submitters.

I think this is okay though. If they exist for any libraries, we want that column to be present and then NA for other libraries.

jashapiro:
> > I would change the vector that actually gets passed in to the function in scpca-nf to only include the cell type columns that are present. Otherwise you don't include them and they won't get created. That makes it so you don't have anything to do here.
> 
> I still don't think this is true. For example, we might have the situation where _some but not all_ of the libraries in a given project have cell types. If that were to happen, we'd end up with `NA` cell types. This has most potential to happen I think for submitters.

If some libraries have cell types and others do not, don't we want `NA`?
This would only affect the merged object, right?

sjspielman:
> If they exist for any libraries, we want that column to be present and then NA for other libraries.

I don't actually know... might we want them to be `"Submitter excluded"`?


>  This would only affect the merged object, right?

But yes, it would only affect the merged object!


Let's regroup over meet shortly!

allyhawkins:
> I don't actually know... might we want them to be "Submitter excluded"?

If that's the case, then we should still make that change in `scpca-nf`.

sjspielman:
Well, it was fun (for me!) while it lasted! Closing this as not happening in _this_ repo.

PR #261, opened 2024-01-19
Retain altexp coldata when specified

### LINKED ISSUE #260: Allow for altExp colData columns to be retained when merging
Currently, the `merge_sce_list()` function will _not_ retain any colData columns in the alternative experiments. 
We'd like to extend this function to have an option to keep altExp columns.

We could include another function argument, `retain_altexp_coldata_cols` with a default of `NULL`. This argument would take a named list, where the name matches the relevant alternative experiment name that should retain those colData columns.

sjspielman:
Closes #260 

This PR adds an argument `retain_altexp_coldata_cols` to `merge_sce_list()` and relevant friends to retain altExp colData columns. I added some tests for it.

I also treated myself to a recreational revision to use `expect_setequal()` and stop relying on `sort()`.

Bonus: This addressed an old comment that was left in the function, `# Is this right?`. It wasn't wrong, but it wasn't _totally_ right.
Bonus 2: Finally, my time to use `purrr::pluck()` (🐱`::`🐔) has come.


sjspielman:
Also meant to say FYI I am writing another test for this condition now, incoming soon.

R/merge_sce_list.R
```diff
@@ -421,10 +431,14 @@ get_altexp_attributes <- function(sce_list) {
...
-    purrr::map_chr("library_id")
+    purrr::map_chr("library_id") |>
+    unname()
 
   metadata_sample_ids <- metadata_list |>
-    purrr::map_chr("sample_id")
+    # can't use map_chr in case we have multiplexed
+    purrr::map("sample_id") |>
+    unlist() |>
+    unname()
```
sjspielman, line 441: cellhash bug squash

inst/WORDLIST
```diff
@@ -1,18 +1,27 @@
-
-altExp
 Alevin
+altExp
+altexp
```
jashapiro, line 3: I'm curious where both of these are sneaking in... I feel like it should just be `altExp`?

inst/WORDLIST
```diff
@@ -1,18 +1,27 @@
...
+barcode
 CMD
+colData
 commenter
 Dockerfile
+ensembl
```
jashapiro, line 9: Probably should be `Ensembl` if we can swing it?

inst/WORDLIST
```diff
@@ -1,18 +1,27 @@
...
 Kallisto
 nf
 pre
-SCE
-SCEs
 PRs
+rowdata
+rowData
```
jashapiro, line 15: same here, just `rowData`?

inst/WORDLIST
```diff
@@ -1,18 +1,27 @@
...
+rowname
 RStudio
+SCE
+sce
```
jashapiro, line 19: Do we need the lower case?

jashapiro (commented):
This looks good. My only substantive comment is about the description, which is a bit of a bear. I made a suggestion, but I am not sure it is better. I think the better thing to do is add some `@examples` lines (In a `\dontrun{}`, block) that show the structure expected for `retain_altexp_coldata_cols`.

Also, spelling questions

sjspielman:
Spelling all cleaned up! Let me know what you think of the example I added. I didn't specify any arguments for spots where we already have defaults, but I could? Or, would you want to see more examples?

R/merge_sce_list.R
```diff
@@ -38,11 +38,25 @@
...
 #'
+#' @examples
+#' \dontrun{
+#' merge_sce_list(
+#'   sce_list = list("sce1" = sce1, "sce2" = sce2),
+#'   retain_altexp_coldata_cols = list(
+#'     "adt" = c("discard", "high.controls"),
+#'     "other_altexp" = c("first_column", "second_column")
+#'   )
+#' )
```
jashapiro, line 57: I think I would have more than one example here. Maybe one with altExps and one without? You can use comments (within comments 🤯)here to describe what each example is meant to demonstrate.

See https://github.com/AlexsLemonade/scpcaTools/blob/main/R/read_alevin.R#L32C1-L51C5 for an example example

sjspielman: Eh? https://github.com/AlexsLemonade/scpcaTools/pull/261/commits/da703d830869fa12ef95f4db3d24d42e68cbe4b3

jashapiro: I would flip the order here, probably, but the main thing I would want to see is an explanation of what the structure of the SCE going into the altExp example is: Explicitly say in the comment that these are SCEs that may contain altExpName `adt` and that those are expected to have a couple of columns. 

While I am looking at it, it is probably worth noting in the argument explanation (and maybe again in the example comment) that any named values that are not present (i.e. if there is no `adt` altExp in the SCE) that these will be ignored.

sjspielman:
Next round in https://github.com/AlexsLemonade/scpcaTools/pull/261/commits/540e25640a01697e32720068cbc309e4e04033be, noting that the dictionary is very particular. It's catching even things in backticks in the `.Rd` files, and we're probably stuck with the changes it wants.

jashapiro:
I just looked back at the main description of the function, and I wonder if we might want to update that section to describe (briefly) what happens to altExps as well? And to make sure that what is described there is still accurate!

sjspielman:
> And to make sure that what is described there is still accurate!

Oh good catch - it is not currently accurate...!

sjspielman:
Also, while we're here - given how we're changing things up to specify altExps coldata columns to retain, I feel like we should also do the same for rowdata column names to preserve. Currently we hardcode `"target_type"`, but I am thinking this should also be an named-list argument like we're doing for colData. What do you think? We'd then just specify "target_type" in scpca-nf.

R/merge_sce_list.R
```diff
@@ -12,41 +14,93 @@
...
 #'
+#'  SCE altExp contents are modified or retained as follows:
+#'  - As with the main experiment, the additional columns `batch_column` and
+#'  `cell_id_column` will be added to the colData slot.
+#'  - Of the original altExp colData columns, only column names provided in
+#'    the argument `retain_altexp_coldata_cols`, as specified for each named
+#'    altExp will be retained.
+#'  - The resulting rowData slot column names will be appended with the given
+#'    SCE's name, as `{sce_name}-{column_name}` except for the column `target_type`
+#'    which is commonly present in CITE-seq alternative experiments.
```
jashapiro, line 34: I do think making `preserve_altexp_rowdata_cols` an input and a namex list makes sense. I'm not sure it is the only column that should be preserved for cellhash? I might be overthinking, but if we have things like the barcode sequence, we probably want to include those too.

R/merge_sce_list.R
```diff
@@ -12,41 +14,93 @@
 #'    its index in the provided `sce_list`.
 #'  - The resulting colData slot will include another new column `cell_id_column`
 #'    (default "cell_id") that will contain the SCE's original column names (i.e.
-#'    original colData rownames). Often, but not always, this rowname holds a
+#'    original colData row names). Often, but not always, this row name holds a
 #'    unique cell barcode.
+#'  - Of the original colData columns, only column names provided in the argument
+#'    `retain_coldata_cols` will be retained.
 #'  - The resulting colData rownames will be be prefixed with `{sce_name-}`.
```
jashapiro, line 21:
```suggestion
#'  - The resulting colData row names will be be prefixed with `{sce_name-}`.
```

inst/WORDLIST
```diff
@@ -1,18 +1,29 @@
...
+ensembl
 Kallisto
 nf
 pre
-SCE
-SCEs
 PRs
+rowData
+rowdata
+rowname
```
jashapiro, line 18: I think we can get rid of this with the edits you made? (I can't find any `rowname` in the docs, except for the one I removed)
```suggestion
```

jashapiro (commented):
This looks good to me, but I agree that we could make a `preserve_altexp_rowdata_cols` named list to match `retain_altexp_coldata_cols` for a nice bit of parallelism.

R/merge_sce_list.R
sjspielman: > I'm not sure it is the only column that should be preserved for cellhash? I might be overthinking

Do you mean for cite? 

Either way, here's what we have -

```
> rowData(altExp(sce_cite)) |> names()
[1] "target_type" "mean"        "detected"  

> rowData(altExp(sce_cellhash)) |> names()
[1] "barcode_id" "sample_id"  "mean"       "detected" 
```

jashapiro: I did mean that. I'm surprised we don't have a `target` or `target_id` as a column? I feel like we should, even though it will be in general the same as the row name. Just me not liking important data in the row names...

jashapiro (approved):
LGTM, assuming the tests all pass

R/merge_sce_list.R
sjspielman: > Just me not liking important data in the row names...

Gotcha, yep. We can add this into `scpca-nf` probably here - 
https://github.com/AlexsLemonade/scpca-nf/blob/25b599adea91c28e672963577d4f10d77a90b150/bin/generate_unfiltered_sce.R#L152-L155

and then sneak into https://github.com/AlexsLemonade/scpca-nf/pull/666, if you want.

PR #264, opened 2024-01-25
Update column handling in merge_sce_list

### LINKED ISSUE #263: merge_sce_list colData updates
From https://github.com/AlexsLemonade/scpca-nf/pull/666#discussion_r1465730488, we should:

- Toss a `unique()` onto the arguments like `retain_coldata_cols` before their usage
- Remove the code for saving altExp stats columns since we should be specifying that at runtime instead

sjspielman:
Closes #263 
As discussed in https://github.com/AlexsLemonade/scpca-nf/pull/666#discussion_r1465730488, I've removed the altExp columns from this function. I also ensure that any vector of retained/preserved columns is unique when it goes into its relevant crux function.

I've updated over in my draft PR in `scpca-nf` as well: https://github.com/AlexsLemonade/scpca-nf/pull/666/commits/061100671bbb2dc3f4aae2018dde392991747fc3

### DIFF
--- a/tests/testthat/test-merge_sce_list.R
+++ b/tests/testthat/test-merge_sce_list.R
@@ -513,7 +513,12 @@ test_that("merging SCEs with 1 altexp and same features works as expected, with 
     sce_list_with_altexp,
     batch_column = batch_column,
     # "total" should get removed
-    retain_coldata_cols = retain_coldata_cols,
+    retain_coldata_cols = c(
+      retain_coldata_cols,
+      "altexps_adt_sum",
+      "altexps_adt_detected",
+      "altexps_adt_percent"
+    ),
     # this row name should not be modified:
     preserve_rowdata_cols = c("gene_names")
   )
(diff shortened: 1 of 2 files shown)


tests/testthat/test-merge_sce_list.R
```diff
@@ -513,7 +513,12 @@ test_that("merging SCEs with 1 altexp and same features works as expected, with
     sce_list_with_altexp,
     batch_column = batch_column,
     # "total" should get removed
-    retain_coldata_cols = retain_coldata_cols,
+    retain_coldata_cols = c(
+      retain_coldata_cols,
+      "altexps_adt_sum",
+      "altexps_adt_detected",
+      "altexps_adt_percent"
+    ),
```
jashapiro, line 521: I think for the purposes of testing simplicity it probably would have been fine to change the expected values below rather than updating this parameter, but it is fine.

jashapiro (approved):
LGTM! Removing code ftw!

Issue #265, opened 2024-03-05
Create a slim scpcaTools docker image


jashapiro:
**Is your feature request related to a problem? Please describe.**

Right now our main `scpcaTools` docker image includes all renv packages we use for development and testing.
This makes for a very large docker image, which is not particuarly efficient!

**Describe the solution you'd like**

We should create a slimmed down version which only has the tools that we need for scpca-nf.

This would not include Seurat, I think, and we could also try to minimize any annotation packages that are included. 

We may also want to separate out python package installation, but this would require more analysis of the specific tools used there.

We can update the docker build scripts to do multiple builds with different tags for each version, for example: `v0.3.1` and `v0.3.1-full` or something like that. The default version and `latest` tags should probably point to the slimmer version.

**Additional context**

This might be a good time to also address #96?

jashapiro:
closed by #270 and #271

PR #269, opened 2024-04-16
Update to Bioc3.18 and reduce merge memory

jashapiro:
This PR started as a couple small changes to reduce the merge memory footprint, but then I had to update some packages, so that ended up sneaking in.

The main changes worth a full look are in `merge_sce_list.R` where I changed a few `purrr` functions to prevent having to create copies of the sce_list. Some of these resulted in slightly more aggressive refactoring.

The other changes are mostly linting updates, now that we have pre-commit in here and some things were being rebuilt. Many of these are changes from single to double `&`, as is now required more aggressively. The changes cascaded a bit, with the most significant change being a place where I converted a bunch of tests to use `stopifnot` rather than separate `if` statements because that reduced the "cyclomatic complexity" by a good degree.

A lot of this is really in prep for #265, but I didn't want to put that in here so as not to end up bloating too much!


jashapiro:
I realized that this update also updates zellkonverter, so there is a chance it would allow merged anndata export to work on bigger files.

allyhawkins (approved):
LGTM!

PR #273, opened 2024-05-03
Update to R 4.4 & renv 3.19

jashapiro:
This PR updates scpcaTools development environment  and docker to use R 4.4 and Bioconductor 3.19. 

When running checks, I found a couple of missing escape characters in documentation, so I added those as well. All checks passed locally, so hopefully there are not any breaking changes!

### DIFF
--- a/R/integrate_sces.R
+++ b/R/integrate_sces.R
@@ -27,7 +27,7 @@
 #' @param ... Any additional parameters to be passed to the given integration method
 #'
 #' @return An updated merged SCE object containing a new reduced dimension named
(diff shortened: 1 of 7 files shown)


jashapiro:
Looks like I forgot to update the sub-lockfiles, which is probably why the 1st docker build attempt failed. 

When testing this though, I wanted to try building straight from the `scpcaTools` version in the PR, rather than what gets pulled from github, so I moved the dockerfile up a level, and rearranged a few things to make sure everything kept working as expected.

allyhawkins (approved):
LGTM

PR #274, opened 2024-05-08
Add batchelor to the slim image

jashapiro:
When running the merge workflow, I found that I did need `batchelor` installed, which I think was the only thing beyond the `slim` packages that was needed. So I am adding it here, because the ability to do clustering does seem reasonable to include in all images.


allyhawkins (approved):
Assuming this builds and passes checks, LGTM

PR #276, opened 2024-05-23
Add rtracklayer and updates

jashapiro:
In testing For https://github.com/AlexsLemonade/scpca-nf/issues/760, I found that we needed `rtracklayer` for the initial imports, which seemed worth adding to the `slim` image. So here I added that.

I also went ahead and updated the rest of the requirements to latest versions as of today, which I might not normally do but since we are still in a testing phase of these images, it seemed reasonable.


allyhawkins (approved):
LGTM

PR #279, opened 2024-07-08
Prep for v0.4

jashapiro:
I was going to just bump the version numbers here, but I thought it would probably be a good idea to also pull everything up to the latest releases for best future compatibility. This does mean we will need a workflow check, but all of the changes seemed to be minor version bumps, so I don't expect any major issues.


allyhawkins (approved):
LGTM, I don't anticipate issues but I agree that we should do a run of the workflow with these updates before creating the release.

PR #284, opened 2024-10-28
283 update lightning to fix critical alert

### LINKED ISSUE #283: update lightning to fix critical alert
Dependabot alerted us to a critical fix: https://github.com/AlexsLemonade/scpcaTools/security/dependabot/41

We should fix this by updating the following files to include a minimum version of `lightning` >= 2.3.3 and then rerunning the [`make-requrements.sh` script](https://github.com/AlexsLemonade/scpcaTools/blob/main/docker/make-requirements.sh). 

https://github.com/AlexsLemonade/scpcaTools/blob/main/docker/requirements.in
https://github.com/AlexsLemonade/scpcaTools/blob/main/docker/requirements_scvi.in

It is possible that simply rerunning the script with the current input files would be sufficient, but including the requirement seems a good idea for security.

jashapiro:
Closes #283
The main goal here was to update `lightning` to include the security fix, but I also updated the requirements script to better deal with updating python packages (I would otherwise have to delete the files and recreate them every time).

I also added an extra bit to the version string for `~=` dependencies to freeze them at the minor version level; I had misunderstood the syntax to freeze at the given level, but it actually does the level above. A `~=`1.1` string will allow updates to version `1.1.x`, `1.2.x` or generally `1.y.x`, but if we specify `~=1.1.1`, it will only update to `1.1.x`, which is what we likely want. 

The package changes seem to be mostly low-level; I don't expect significant effects, but we will want to test these at least in part before a new release of `scpcaTools` and incorporation into `scpca-nf`.


jaclyn-taroni (approved):
👍🏻

PR #286, opened 2024-10-30
Add rounding to alevin-fry import

jashapiro:
Apparently when we switched to using `fishpond` for more efficient importing, we forgot to ensure that the counts were still rounded. I am adding a step to round the counts here, which should fix https://github.com/AlexsLemonade/scpca-nf/issues/790 once we update and reprocess all of the samples.


jaclyn-taroni (approved):
👍🏻

PR #291, opened 2025-04-08
Output merged SCE as dgCMatrix

### LINKED ISSUE #289: Ensure that merged object counts are stored as sparse dgCMatrix
When working with the merged objects, I noticed that the counts matrix wasn't quite the correct format. It looks we are storing a `DelayedMatrix` instead of a `dgCMatrix`. I think we want to match what's typically in a `SingleCellExperiment` object, which would be a `dgCMatrix`. I don't think this is super pressing, but we might consider adding a line to the `merge_sce_list()` function to convert to `dgCMatrix` before returning.

allyhawkins:
Closes #289 

Here I made a small change to the `merge_sce_list` function to make sure the output matrix is not a`delayedMatrix`. To do this I just added the `delayed=FALSE` option to the `combineCols` call. I tested the run time of this vs. the original function with a small set of SCEs and it took 2 seconds instead of 1 second so there is going to be some increased time. I also tested directly converting the counts assay  to a `CsparseMatrix` after creating the object and that took 3 seconds, so I think this is our best option.

### DIFF
--- a/R/merge_sce_list.R
+++ b/R/merge_sce_list.R
@@ -237,10 +237,30 @@ merge_sce_list <- function(
           preserve_rowdata_cols = unique(altexp_rowdata_cols)
         )
     }
-  }
 
-  # Create the merged SCE from the processed list
-  merged_sce <- do.call(combineCols, c(unname(sce_list), delayed = FALSE))
+    # if using alt exp, need to merge and then convert to dgCMatrix
+    merged_sce <- do.call(combineCols, unname(sce_list))
+
+    # first update the assays in the main exp
+    # purrr doesn't work for updating items in place in SCE so use for loops
+    for(name in assayNames(merged_sce)){
+      assay(merged_sce, name) <- as(assay(merged_sce, name), "CsparseMatrix")
+    }
+
+    # now for the alt exps
+    for(altexp_name in names(altexp_attributes)){
+      alt_merged_sce <- altExp(merged_sce, altexp_name)
+      for(name in assayNames(alt_merged_sce)){
+        assay(alt_merged_sce, name) <- as(assay(alt_merged_sce, name), "CsparseMatrix")
+      }
+      altExp(merged_sce, altexp_name) <- alt_merged_sce
+    }
+
+  } else {
+    # Create the merged SCE from the processed list
+    # only use delayed = FALSE if no altExps
+    merged_sce <- do.call(combineCols, c(unname(sce_list), delayed = FALSE))
+  }
 
   # Update metadata in merged objects, using the unmerged sce_list
   metadata(merged_sce) <- sce_list |>
--- a/R/merge_sce_list.R
+++ b/R/merge_sce_list.R
@@ -240,7 +240,7 @@ merge_sce_list <- function(
   }
 
   # Create the merged SCE from the processed list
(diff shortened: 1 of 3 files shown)


sjspielman (commented):
I'm assuming the CI failure here is unrelated to this code change, and just a fun time with basilisk (but didnt look too closely at it).

I do think we want to add a test for this in https://github.com/AlexsLemonade/scpcaTools/blob/main/tests/testthat/test-merge_sce_list.R to confirm the format is as expected.

R/merge_sce_list.R
```diff
@@ -237,10 +237,30 @@ merge_sce_list <- function(
           preserve_rowdata_cols = unique(altexp_rowdata_cols)
         )
     }
-  }
 
-  # Create the merged SCE from the processed list
-  merged_sce <- do.call(combineCols, unname(sce_list))
+    # if using alt exp, need to merge and then convert to dgCMatrix
```
jashapiro, line 241: For modularity, I think it would be better to have a separate block `if` block for the actual merging step. The way it is now, the merging of non-altexp is nested under a header of "Handle altExps", which seems odd. For my own preference of trying to have the short option in an `if...else` block first, I might reverse the order; testing `if(!include_altxp)`.  But that seems less important than the modularity part.

tests/testthat/test-merge_sce_list.R
```diff
@@ -682,7 +702,7 @@ test_that("merging SCEs where 1 altExp is missing works as expected, with altexp
   counts_mat <- counts(merged_altexp)
   sce4_counts <- counts_mat[, merged_sce[[batch_column]] == "sce4"]
   expect_true(
-    all(is.na(sce4_counts))
+    all(sce4_counts == 0)
   )
```
jashapiro, line 706: I still had this tab open, and thought that here (and below) we can now also use a more specific test:
```
expect_equal(sum(sce4_counts), 0)
```

jashapiro: I guess that theoretically that could accidentally pass if you had balanced positives and negatives, so maybe a bit safer:

```
expect_equal(sum(abs(sce4_counts)), 0)
```

allyhawkins:
> I'm assuming the CI failure here is unrelated to this code change, and just a fun time with basilisk (but didnt look too closely at it).

This was just a stochastic error that fixed when I re-ran the job. 

> I do think we want to add a test for this in https://github.com/AlexsLemonade/scpcaTools/blob/main/tests/testthat/test-merge_sce_list.R to confirm the format is as expected.

I added some tests to check that the format was correct for merging with/without an `altExp`. I also updated some tests to account for the new format that started failing. We don't have NA anymore in the matrices for `altExp`, just 0. 

Another thing to keep in mind is that tests were failing because when using `combineCols` and `delayed=FALSE` with an `altExp` there was some complaining about the formatting and `combineCols` can't handle doing that conversion when there is an `altExp`. To account for that, I have a separate block that merges and converts the assays if including the `altExp`. I didn't want to do it for all objects because it is not as efficient as just using `delayed=FALSE`, so I thought it was good to keep that for the majority of cases. 

As long as this pass tests (which I believe it should...), then this is ready for another look.

sjspielman (approved):
LGTM, yay tests!

PR #294, opened 2025-04-09
Account for empties in merged SCE to be either NA or 0

### LINKED ISSUE #293: Fix tests for merge SCE list to pass R-CMD-check-current
Right now, the package is passing the renv R-CMD check but not the check that doesn't use renv, see https://github.com/AlexsLemonade/scpcaTools/actions/runs/14363916829/job/40272075020. It looks like the merging tests are expecting `NA` values rather than 0s. In #291, I updated these tests to check for 0s since the `CsparseMatrix` format conversion caused them to be converted from `NA` to 0. However, it looks like in the check that doesn't use `renv`, the `NA`'s stay there so there must be some difference in packages that avoids converting `NA`. I'm sure this is a `Matrix` issue somewhere...

allyhawkins:
Closes #293 

Here I'm trying a solution for the fact that using the environment we have set up in `renv` replaces NA values with 0 when converting to CsparseMatrix, while our R-cmd-check-current keeps things as `NA`. I don't think it really matters which one it does, but the way we have tests set up it only allows all 0s in the output. So I updated the tests to first check for all NAs, and then if it's not all NAs, they should all be 0s. 

The other option is we could just convert all NAs to 0s explicitly? I guess technically it's going to happen with the conversion to sparse matrices anyways...?

### DIFF
--- a/tests/testthat/test-merge_sce_list.R
+++ b/tests/testthat/test-merge_sce_list.R
@@ -695,9 +695,14 @@ test_that("merging SCEs where 1 altExp is missing works as expected, with altexp
   # check that the 0s/NAs are as expected
   counts_mat <- counts(merged_altexp)
   sce4_counts <- counts_mat[, merged_sce[[batch_column]] == "sce4"]
-  expect_equal(
-    sum(abs(sce4_counts)), 0
-  )
+  # check that everything is either all NA or all 0
+  all_na <- expect_true(all(is.na(sce4_counts)))
+  if(!all_na){
+    # will error if not all NA and not all 0
+    expect_equal(
+      sum(abs(sce4_counts)), 0
+    )
+  }
   numeric_counts <- counts_mat[, merged_sce[[batch_column]] != "sce4"]
   expect_true(
     all(is.finite(numeric_counts))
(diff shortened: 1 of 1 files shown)


tests/testthat/test-merge_sce_list.R
```diff
@@ -747,9 +752,16 @@ test_that("merging SCEs with different altExps works as expected; each SCE has 1
...
-    sum(abs(counts(adt_merged)[, merged_sce[[batch_column]] == "sce2"])), 0
+  # check that everything is either all NA or all 0
+  all_na <- expect_true(
+    all(is.na(counts(adt_merged)[, merged_sce[[batch_column]] == "sce2"]))
   )
+  if(!all_na){
+    # will error if not all NA and not all 0
+    expect_equal(
+      sum(abs(counts(adt_merged)[, merged_sce[[batch_column]] == "sce2"])), 0
+    )
```
jashapiro, line 763: I think you probably want to do this more directly for what you are trying to test?
```suggestion
  adt_counts <- counts(adt_merged)[, merged_sce[[batch_column]] == "sce2"]
  expect_true(
    all(is.na(adt_counts) | adt_counts == 0)
  )
```

But if you leave this as a dense matrix, as I think you should, then these would be properly left as `NA` to reflect that the ADTs were not measured:

```suggestion
  adt_counts <- counts(adt_merged)[, merged_sce[[batch_column]] == "sce2"]
  expect_true(
    all(is.na(adt_counts))
  )
```

jashapiro (commented):
I left a couple suggestions below for how I would handle the tests: I think you want to keep them simply doing what you are describing, rather than trying to add in an `if`.

But in reality, I think you probably want to revert some of the changes you made earlier; as I said in https://github.com/AlexsLemonade/scpcaTools/issues/293#issuecomment-2791458518, I think we may want to leave the altExp as a generic `Matrix` object, which will hopefully not convert NA to 0, because I don't think we actually want to do that: we want to preserve the NA values to reflect that these ADTs were not measured.

In other words, I think the new version of the Matrix package is actually more correct, as it seems to be properly preserving the NA values in the sparse matrix too.

allyhawkins:
Closing this in favor of the approach in #295

PR #295, opened 2025-04-10
Leave altExps (potentially) dense to preserve NA values

jashapiro:
As an alternate solution to #293, here I am changing the conversion of the altExps to be to leave them as dense matrices, so as to preserve the `NA` values for missing columns.  I also updated the tests to match.

I tested this both with the current renv and after updating all packages to the latest versions, so I expect this to pass both the pre- and post-merge testing.

There is also a very small renv update, because renv refused to restore on a new machine without updating `BiocManager`, so I added that small update.

At some point in the future we might want change to sparse matrices for altExps as well, and it seems from #293 that this could preserve the NA values with newer versions of the `Matrix` package: we may want to investigate when that change happened and update requirements (and the `renv`) to specify that version. But for now, I think leaving these matrices dense should be fine; they are small (compared to the RNA expression matrices), and tend to be denser anyway.

### DIFF
--- a/R/merge_sce_list.R
+++ b/R/merge_sce_list.R
@@ -260,7 +260,7 @@ merge_sce_list <- function(
     for(altexp_name in names(altexp_attributes)){
       alt_merged_sce <- altExp(merged_sce, altexp_name)
       for(name in assayNames(alt_merged_sce)){
-        assay(alt_merged_sce, name) <- as(assay(alt_merged_sce, name), "CsparseMatrix")
+        assay(alt_merged_sce, name) <- as(assay(alt_merged_sce, name), "Matrix")
       }
       altExp(merged_sce, altexp_name) <- alt_merged_sce
     }
--- a/R/merge_sce_list.R
+++ b/R/merge_sce_list.R
@@ -256,7 +256,7 @@ merge_sce_list <- function(
       assay(merged_sce, name) <- as(assay(merged_sce, name), "CsparseMatrix")
     }
 
-    # now for the alt exps, but leave them dense
+    # now for the alt exps: make sure they are Matrix class (not Delayed)
     for(altexp_name in names(altexp_attributes)){
       alt_merged_sce <- altExp(merged_sce, altexp_name)
       for(name in assayNames(alt_merged_sce)){
--- a/R/merge_sce_list.R
+++ b/R/merge_sce_list.R
@@ -247,7 +247,7 @@ merge_sce_list <- function(
     # only use delayed = FALSE if no altExps
     merged_sce <- do.call(combineCols, c(unname(sce_list), delayed = FALSE))
   } else {
-    # if using alt exp, need to merge and then convert to dgCMatrix
+    # if using alt exp, need to merge and then convert to CsparseMatrix
     merged_sce <- do.call(combineCols, unname(sce_list))
 
     # first update the assays in the main exp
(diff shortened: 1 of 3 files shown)


jashapiro:
After a bit more curiosity-based testing: it turns out that converting to `Matrix` class _may_ result in a sparse matrix if the input matrix is sufficiently sparse, so this version will still work as expected if we do end up with sparse altExps.

R/merge_sce_list.R
```diff
@@ -256,11 +256,11 @@ merge_sce_list <- function(
       assay(merged_sce, name) <- as(assay(merged_sce, name), "CsparseMatrix")
     }
 
-    # now for the alt exps
+    # now for the alt exps, but leave them dense
     for(altexp_name in names(altexp_attributes)){
       alt_merged_sce <- altExp(merged_sce, altexp_name)
       for(name in assayNames(alt_merged_sce)){
-        assay(alt_merged_sce, name) <- as(assay(alt_merged_sce, name), "CsparseMatrix")
+        assay(alt_merged_sce, name) <- as(assay(alt_merged_sce, name), "Matrix")
```
allyhawkins: Should we be concerned about the amount of memory this might take up in a merged object if we leave them dense?

allyhawkins: I just saw your latest commit so ignore this comment.

allyhawkins:
> At some point in the future we might want change to sparse matrices for altExps as well, and it seems from https://github.com/AlexsLemonade/scpcaTools/issues/293 that this could preserve the NA values with newer versions of the Matrix package: we may want to investigate when that change happened and update requirements (and the renv) to specify that version. But for now, I think leaving these matrices dense should be fine; they are small (compared to the RNA expression matrices), and tend to be denser anyway.

Yesterday I was trying to figure out why we are having discrepancies and it looks like the change was made in Matrix 1.5 to convert to 0? I looked at the release notes and found this line: 
> is.na(<ndenseMatrix>) gave TRUE where the x slot had NA. Now the result is always a zero matrix.

But we are using 1.7 which is the most recent version, so I would assume that both the renv and current would be using 0s, but that does not seem to be the case.

allyhawkins (approved):
Approving this, since ultimately I agree that this is the right approach to leave the NA values as is. I still don't understand what's going on with `Matrix`, but that's okay. Thanks for doing this!

jashapiro:
> But we are using 1.7 which is the most recent version, so I would assume that both the renv and current would be using 0s, but that does not seem to be the case.

Looks like it was a bug in DelayedArray that is now fixed: https://bioconductor.org/packages/release/bioc/news/DelayedArray/NEWS

But using `Matrix` seems to do the correct thing, which is good, and I think we are pretty safe if the matrix happens to be sparse and Matrix decides to make it sparse. We could see if it is worth also including the DelayedArray update in the renv for safety/Docker though.

allyhawkins:
> > But we are using 1.7 which is the most recent version, so I would assume that both the renv and current would be using 0s, but that does not seem to be the case.
> 
> Looks like it was a bug in DelayedArray that is now fixed: https://bioconductor.org/packages/release/bioc/news/DelayedArray/NEWS
> 
> But using `Matrix` seems to do the correct thing, which is good, and I think we are pretty safe if the matrix happens to be sparse and Matrix decides to make it sparse. We could see if it is worth also including the DelayedArray update in the renv for safety/Docker though.

I think to do that we would also need to update to bioconductor 3.20 since the 0.32 version of DelayedArray is not available as part of 3.19 as far as I can tell. So probably not worth doing right now if this solution is working.

jashapiro:
> I think to do that we would also need to update to bioconductor 3.20 since the 0.32 version of DelayedArray is not available as part of 3.19 as far as I can tell. So probably not worth doing right now if this solution is working.

Yes, that seems to be the case, and we _definitely_ don't want to go to 3.20 right now.

PR #297, opened 2025-06-11
Update read cellranger to read in H5 file

### LINKED ISSUE #296: Update read_cellranger to take as input HDF5 file rather than quant_dir
The newer versions of Cell Ranger have slightly different organizations of where the output files live since we originally wrote the [`read_cellranger`](https://github.com/AlexsLemonade/scpcaTools/blob/main/R/read_cellranger.R) function. Previously we were just providing the directory to the outputs and then hard-coding the path to the H5AD file to read in. We should update this to directly provide the H5AD file instead.

allyhawkins:
Closes #296 

This PR modifies the `read_cellranger()` function to use the path to the H5 file directly as input rather than using a `quant_dir`. While I was here, I also updated the `import_quant_data()` function to use the H5 file when the tool is `cellranger`. 

The only other thing I did here was remove the `sample.names` argument since we remove the `colData` anyways. And then I noticed that the `metadata` had `Samples` as a list item, so I made that an empty list to match the formatting of other objects.

### DIFF
--- a/R/read_cellranger.R
+++ b/R/read_cellranger.R
@@ -1,6 +1,6 @@
 #' Read in counts data processed with Cell Ranger
 #'
-#' @param quant_dir Path to directory where output files are located.
+#' @param cellranger_h5_file Path to H5AD file output from Cell Ranger.
 #'
 #' @return SingleCellExperiment of gene x cell counts matrix
 #' @export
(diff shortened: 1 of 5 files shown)


R/read_cellranger.R
```diff
@@ -1,32 +1,33 @@
...
     col.names = TRUE
   )
 
   # for consistency with other quantifiers:
   # change the column names just the barcode value, which is the first part of the barcode name
-  # drop colData
+  # drop colData and metadata
   colnames(sce) <- str_extract(colnames(sce), "^([ACGT]+)")
   SummarizedExperiment::colData(sce) <- NULL
+  SummarizedExperiment::metadata(sce) <- list()
```
jashapiro, line 31: Is there any data in the metadata that we might want to keep?

jashapiro (commented):
I understand why you did this, but I'm actually wondering if we might just want to avoid using this particular function for Cell Ranger multi data. I don't like that the `import_quant_data()` function works so differently for Cell Ranger now, so I am not sure that there is as much value in the single function. I wonder if we might be better off just using `read10xCounts()` directly in our workflows. 

I think an alternative might be to mostly leave your changes to the `read_cellranger()` and move the path-finding to the `import_quant_data()` function so we can leave the inputs to that mostly alone. If we need changes to the pathfinding, we can make that function include a `cellranger-multi` option so that we can update the pathfinding as appropriate (maybe specifying the sample? as an added argument). In short, I'd rather not break any existing use of `import_quant_data()` when we update this.

For the `read_cellranger()` function itself, I think it is fine to make it more specific, but I would be less strict and allow reading from the matrix directory as well, since `read10xCounts()` supports that natively. I'm still not sure which we really would want to use in the pipeline, though I suspect it doesn't matter too much after writing out the rds and rereading it.

PR #299, opened 2025-07-11
Fix cell_id bug

### LINKED ISSUE #298: Investigate possible bug: cell_id column is not being properly defined
I am filing this issue because I have just noticed, while working with a merged SCE downloaded (yesterday) from the portal, the `cell_id` column in the colData seems to _only contain_ the library IDs, but it should be formatted as `{library_id}-{barcode}`. Note that the SCE colnames are correctly formatted this way. 

I am not sure where the bug is coming from, but when I search `cell_id` in the scpca-nf repo, only this spot comes up:
https://github.com/AlexsLemonade/scpca-nf/blob/4bb82aa635b572a62f2028dbec587fcfd2155e26/bin/merge_sces.R#L275

So, I am assuming the issue lies in the `merge_sce_list` function here. Note that in our tests for this function, we don't seem to ever check the _values_ in the `cell_id` column, so indeed it may have snuck through. 

The goal of this issue is therefore to identify if the bug is indeed here, and if so what's the fix.

sjspielman:
Closes #298 

I have found the bug, and it was small but mighty. We were correctly defining the `cell_id` column as the SCE column name, but sadly we did so before actually updating the SCE colnames for merge in the first place.... I updated the code order to assign `cell_id` after the colnames are updated. There was also a buggy test that was checking `cell_id` against the wrong true value - it should be checked against the _new_ SCE colnames!

When I went to commit this, precommit failed with this fun old friend:
```
/Users/sjspielman/ALSF/scpca/scpcaTools/R/merge_sce_list.R:107:1: style: [cyclocomp_linter] Functions should have cyclomatic complexity of less than 20, this has 22.
```

The code change I implemented would probably not have tipped this over, but it's only coming up now. This function is indeed a bit of a 🎢 that could probably be addressed, but it seemed easier to turn off that lintr bit for the function now. Let me know if you disagree with that choice!

### DIFF
--- a/R/merge_sce_list.R
+++ b/R/merge_sce_list.R
@@ -1,3 +1,4 @@
+# nolint start: cyclocomp_linter.
 #' Merge a list of SCEs into one SCE object
 #'
 #' This function takes an optionally-named (if named, ideally by a form of
@@ -237,12 +238,11 @@ merge_sce_list <- function(
           preserve_rowdata_cols = unique(altexp_rowdata_cols)
         )
     }
-
   }
 
   # Merge SCEs ------------------------------------------------------
 
-  if(!include_altexp){
+  if (!include_altexp) {
     # Create the merged SCE from the processed list
     # only use delayed = FALSE if no altExps
     merged_sce <- do.call(combineCols, c(unname(sce_list), delayed = FALSE))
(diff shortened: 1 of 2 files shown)


jashapiro (approved):
LGTM

PR #304, opened 2025-07-29
Add scDblFinder to environment

### LINKED ISSUE #303: Add scDblFinder to environment
This issue tracks adding `scDblFinder` as a dependency. We'll be performing doublet detection in `scpca-nf`, and we plan to use this Docker container for that step.

sjspielman:
Closes #303 and does what it says! I added this library to `docker/dependencies.R` and snapshotted it into renv.


sjspielman:
> As such, we probably want to add it to one of the supplemented images.

In terms of which other image to put this in, of the options we have I'd go with either the `reports` or `seurat` one but how to choose besides flipping a coin isn't terribly clear to me if there's a stronger opinion out there!

allyhawkins:
> > As such, we probably want to add it to one of the supplemented images.
> 
> In terms of which other image to put this in, of the options we have I'd go with either the `reports` or `seurat` one but how to choose besides flipping a coin isn't terribly clear to me if there's a stronger opinion out there!

I would just include it in the full image and not in any of the subsetted images.

sjspielman:
Ok once @jashapiro flips a coin I'll be able run to `make-requirements.sh`; for now I've canceled workflows.

jashapiro:
> > > As such, we probably want to add it to one of the supplemented images.
> > 
> > 
> > In terms of which other image to put this in, of the options we have I'd go with either the `reports` or `seurat` one but how to choose besides flipping a coin isn't terribly clear to me if there's a stronger opinion out there!
> 
> I would just include it in the full image and not in any of the subsetted images.

I'm probably fine with the base image, as it doesn't seem to be too large a package.

sjspielman:
Ok, so if it's indeed going in the full image, then I wouldn't expect any updates when running `make-requirements.sh`. I ran the script anyways to confirm, and I did indeed get some diffs...
- The `lock` files' `BiocManager` version got bumped. It looks like this was correct, since the version they are now at matches what is currently in the big`renv.lock`, so this was just out of date.
- Interestingly we now have `nvidia` dependencies in the `requirement` files? This was unexpected, so I wanted to push so folks could see. To confirm, I did run the script from the `scpcatools-dev` environment.

jashapiro:
Okay, looking back at this: the main `renv.lock` file (and `dependencies.R`) should not be changed to add `scDoubleFinder`. Instead it should be added to the `-p` list in `make-requirements.sh`

sjspielman:
> My guess is that you are running this on the server? I'm not sure I ever properly tested it in that situation. For now, just don't commit the changes to the requirements files.

Good guess! I couldn't get renv building locally (xgboost problems, so just went to server). I'll revert that commit.

sjspielman:
I'm gonna shut this PR down and move to a fresh one where I'm _just doing_ this:

> Okay, looking back at this: the main renv.lock file (and dependencies.R) should not be changed to add scDoubleFinder. Instead it should be added to the -p list in make-requirements.sh for the slim docker image.

And then re-running `make-requirements.sh` without bringing nvidia into the mix.

PR #306, opened 2025-07-30
Update how we handle basilisk installation

allyhawkins:
Related to https://github.com/AlexsLemonade/scpca-nf/issues/789, this fix attempts to solve the problem with installing and using basilisk on a read-only singularity image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,17 +65,11 @@ FROM slim AS anndata
 LABEL org.opencontainers.image.title "scpcatools-anndata"
 
 COPY docker/renv_zellkonverter.lock renv.lock
+ENV BASILISK_USE_SYSTEM_DIR=1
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -65,17 +65,12 @@ FROM slim AS anndata
 LABEL org.opencontainers.image.title "scpcatools-anndata"
 
 COPY docker/renv_zellkonverter.lock renv.lock
+ARG BASILISK_USE_SYSTEM_DIR=1
```
jashapiro, line 68: This location should be fine, but just for safety, let's define this in the first build stage.

allyhawkins: > This location should be fine, but just for safety, let's define this in the first build stage.

Do you mean when building the slim image?

PR #307, opened 2025-07-31
Clean up dockerfile

jashapiro:
This PR was originally just to change some syntax in the dockerfile to avoid warnings about old syntax. 

But when I went to build it locally, I got some errors related to the Rprofile looking for renv files, so I also updated the ignore files to exclude a few more things when building.

There are no real substantive changes that I am aware of, but as long as we were changing things, I thought I would sneak this in.


allyhawkins (approved):
LGTM

PR #309, opened 2025-08-01
Add ggmap and DT to reports docker

### LINKED ISSUE #308: Add ggmap and DT to reports image
We need to add `ggmap` to the reports image to be able to make the dot plot we are adding. While we are here we mine as well add in `DT` so we can fix the tables. These should be in the full image and then in the reports subset image.

allyhawkins:
Closes #308 

I added `ggmap` and `DT` has packages required for building the `reports` image. To do that, I just added them to the list of packages using for building the `reports` image and then ran the `make-requirements.sh` script.

I wasn't sure if I should also add them to `dependencies.R` and then update the main `renv.lock` file? That's what's implied in the [instructions](https://github.com/AlexsLemonade/scpcaTools/tree/main/docker#generating-r-and-python-dependency-files) but I noticed `scDblFinder` isn't in the main `renv.lock`, so I wasn't totally sure what the best move was?


jashapiro:
> I wasn't sure if I should also add them to `dependencies.R` and then update the main `renv.lock` file? That's what's implied in the [instructions](https://github.com/AlexsLemonade/scpcaTools/tree/main/docker#generating-r-and-python-dependency-files) but I noticed `scDblFinder` isn't in the main `renv.lock`, so I wasn't totally sure what the best move was?

Yeah, for completeness we probably should add everything to the main `renv.lock` file (I re-realized this after the scDblFinder review, but didn't think it was worth another PR). Can you please add all three to the `dependencies.R` file? Thanks!

PR #311, opened 2025-08-01
Potential fix for basilisk installation with ENV and TRUE

allyhawkins:
After still seeing some issues with basilisk installing the environment and not being able to properly find `anndata`, I'm updating this to use `ENV` and not `ARG`. Hopefully this will set the environment variable in the container so that when it's used on a new machine it doesn't try to install the environment. 

I also think maybe `1` is incorrect and instead it should be `TRUE`. I tested this by pulling the existing image that we have in singularity and then inside the image I set the environment variable and then tried to run `basilisk`: 

Try 1: 
```
Singularity> export BASILISK_USE_SYSTEM_DIR=1
Singularity> R

R version 4.4.1 (2024-06-14) -- "Race for Your Life"
Copyright (C) 2024 The R Foundation for Statistical Computing
Platform: x86_64-pc-linux-gnu

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.

  Natural language support but running in an English locale

R is a collaborative project with many contributors.
Type 'contributors()' for more information and
'citation()' on how to cite R or R packages in publications.

Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.

> proc <- basilisk::basiliskStart(env = zellkonverter::zellkonverterAnnDataEnv(), testload = 'anndata')
Registered S3 method overwritten by 'zellkonverter':
  method                                             from      
  py_to_r.pandas.core.arrays.categorical.Categorical reticulate
Error: Unable to find conda binary. Is Anaconda installed?
```

So then I tried to use `TRUE` and this is what happened, which is the correct behavior: 

```
Singularity> export BASILISK_USE_SYSTEM_DIR=TRUE
Singularity> R

R version 4.4.1 (2024-06-14) -- "Race for Your Life"
Copyright (C) 2024 The R Foundation for Statistical Computing
Platform: x86_64-pc-linux-gnu

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.

  Natural language support but running in an English locale

R is a collaborative project with many contributors.
Type 'contributors()' for more information and
'citation()' on how to cite R or R packages in publications.

Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.

> proc <- basilisk::basiliskStart(env = zellkonverter::zellkonverterAnnDataEnv(), testload = 'anndata')
Registered S3 method overwritten by 'zellkonverter':
  method                                             from      
  py_to_r.pandas.core.arrays.categorical.Categorical reticulate
```

So I think setting `ENV` should work and setting the correct argument. I can build it locally and test if we prefer to do that before merging this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ LABEL org.opencontainers.image.source="https://github.com/AlexsLemonade/scpcaToo
 
 #### Build settings for Renv and Basilisk
 ENV RENV_CONFIG_CACHE_ENABLED=FALSE
-ARG BASILISK_USE_SYSTEM_DIR=1
+ENV BASILISK_USE_SYSTEM_DIR=TRUE
 
 #### R packages
 # Use renv for R packages
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ LABEL org.opencontainers.image.source="https://github.com/AlexsLemonade/scpcaToo
 
 #### Build settings for Renv and Basilisk
 ENV RENV_CONFIG_CACHE_ENABLED=FALSE
-ENV BASILISK_USE_SYSTEM_DIR=TRUE
+ENV BASILISK_USE_SYSTEM_DIR=1
 
 #### R packages
 # Use renv for R packages


jashapiro:
The package definitely requires `BASILISK_USE_SYSTEM_DIR=1`: 

https://github.com/LTLA/basilisk.utils/blob/04f649cb529456dbce44cc8896471ea7ad24b01f/R/useSystemDir.R#L17C1-L19C2

So I would not make that particular change. I suspect that just changing ARG to ENV is not sufficient, but we can start with that and maybe see what other changes might be needed (such as modifying the bash environment?). I'd first see what you get with just `echo $BASILISK_USE_SYSTEM_DIR` after starting the singularity image.

jashapiro:
> > So I would not make that particular change. I suspect that just changing ARG to ENV is not sufficient, but we can start with that and maybe see what other changes might be needed (such as modifying the bash environment?). I'd first see what you get with just echo $BASILISK_USE_SYSTEM_DIR after starting the singularity image.
> 
> When using `ARG` if I run `echo $BASILISK_USE_SYSTEM_DIR` I get nothing in return. I agree that the package is expecting it to be `1`, but when I set the environment variable to 1 and then go into R I get that error as shown in my original comment. If I set it to `TRUE` and then go into R then I am able to run the command you mentioned with no error.

Right, but that is with the existing image, which may not be behaving correctly for other reasons. If we set the variable to `TRUE` it will not install the conda environment at build time at all.

allyhawkins:
> > > So I would not make that particular change. I suspect that just changing ARG to ENV is not sufficient, but we can start with that and maybe see what other changes might be needed (such as modifying the bash environment?). I'd first see what you get with just echo $BASILISK_USE_SYSTEM_DIR after starting the singularity image.
> > 
> > 
> > When using `ARG` if I run `echo $BASILISK_USE_SYSTEM_DIR` I get nothing in return. I agree that the package is expecting it to be `1`, but when I set the environment variable to 1 and then go into R I get that error as shown in my original comment. If I set it to `TRUE` and then go into R then I am able to run the command you mentioned with no error.
> 
> Right, but that is with the existing image, which may not be behaving correctly for other reasons. If we set the variable to `TRUE` it will not install the conda environment at build time at all.

Yea I was just looking at it again and you're right. I'll revert back to using 1 but just change it to ENV and see if that properly sets it.

jashapiro (approved):
I expect we will need more than this, but it is a start.

PR #322, opened 2025-09-24
Keep data frames in metadata when exporting to anndata

### LINKED ISSUE #321: Update sce_to_anndata to retain data frames
> Just revisiting the initial step: I think we would probably be fine with this simpler version:
> 
> ```
> metadata_to_keep <- metadata(sce_to_convert) |>
>   purrr::keep(\(x) {is.vector(x) || is.data.frame(x)})
> ``` 

 _Originally posted by @jashapiro in [#317](https://github.com/AlexsLemonade/scpcaTools/issues/317#issuecomment-3304555194)_

---

This issue tracks using the code as suggested above to update the `sce_to_anndata()` function, which will change the final metadata in the object to include data frames. Tests will also need to be updated.

sjspielman:
Closes #321 
Pretty much does what it says! 

Let me know if you think there are any other cases to catch here. 
I expanded tests sans Claude, so all complaints can be registered with me directly. I tried to cover a cople conditions without getting too redundant, again let me know if you think more/less/different is needed. 

Note that as part of developing this code, I did also directly looked at the exported test anndata and it looks good. I also ran this on a fresh processed object, and it the metadata content looks as we expect; we now have both `cellassign_predictions` and `infercnv_table` data frames. This output does show that it has `sample_metadata` in there  rather than in `obs`, but that's only because I ran this function directly on the processed object! `scpca-nf` code, during anndata conversion, rips that data frame out. But, seeing this, I kind of wondered if we want a default `remove_metadata fields = "sample_metadata"` default or so in the scpcaTools function. But, really only kind of, since it's not something specific about anndata conversion so much as how _we_ like to format our converted objects in the context of the full pipeline.

```
>>> adata
AnnData object with n_obs × n_vars = 3226 × 60319
    obs: 'barcodes', 'sum', 'detected', 'subsets_mito_sum', 'subsets_mito_detected', 'subsets_mito_percent', 'total', 'prob_compromised', 'miQC_pass', 'scDblFinder_class', 'scDblFinder_score', 'scpca_filter', 'sizeFactor', 'cluster', 'singler_celltype_ontology', 'singler_celltype_annotation', 'cellassign_celltype_annotation', 'cellassign_celltype_ontology', 'cellassign_max_prediction', 'consensus_celltype_annotation', 'consensus_celltype_ontology', 'is_infercnv_reference', 'infercnv_total_cnv'
    var: 'gene_ids', 'gene_symbol', 'mean', 'detected'
    uns: 'X_name', 'af_num_cells', 'af_permit_type', 'af_resolution', 'alevinfry_version', 'assay_ontology_term_id', 'cellassign_predictions', 'cellassign_reference', 'cellassign_reference_organs', 'cellassign_reference_source', 'cellassign_reference_version', 'celltype_methods', 'cluster_algorithm', 'cluster_nn', 'cluster_weighting', 'filtering_method', 'highly_variable_genes', 'include_unspliced', 'infercnv_reference_celltypes', 'infercnv_success', 'infercnv_table', 'library_id', 'mapped_reads', 'mapping_tool', 'min_gene_cutoff', 'normalization', 'prob_compromised_cutoff', 'project_id', 'reference_index', 'salmon_version', 'sample_id', 'sample_metadata', 'sample_type', 'scpca_filter_method', 'seq_unit', 'singler_date', 'singler_gene_set_version', 'singler_reference', 'singler_reference_label', 'singler_reference_source', 'singler_reference_version', 'tech_version', 'total_reads', 'transcript_type', 'usa_mode'
    obsm: 'PCA', 'UMAP'
    layers: 'logcounts', 'spliced'
```

### DIFF
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -21,6 +21,26 @@ jobs:
       contents: read
       packages: write
     steps:
+      - name: Clear space
+        run: |
+          echo "Disk space before cleanup:"
+          df -h
+          # Software to remove
+          # see https://mathio28.medium.com/squeezing-disk-space-from-github-actions-runners-an-engineers-guide-d5fbe4443692
+          sudo rm -rf \
+            /usr/lib/jvm \
+            /usr/share/dotnet \
+            /usr/share/swift \
+            /usr/local/.ghcup \
+            /usr/local/.rustup \
+            /usr/local/julia* \
+            /usr/local/lib/android \
+            /usr/local/share/powershell \
+            /opt/hostedtoolcache
+
+          echo "Disk space after cleanup:"
+          df -h
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
 
@@ -75,6 +95,21 @@ jobs:
             image_name: scpcatools-infercnv
 
     steps:
+      - name: Clear space
+        run: |
+          # Software to remove
+          # see https://mathio28.medium.com/squeezing-disk-space-from-github-actions-runners-an-engineers-guide-d5fbe4443692
+          sudo rm -rf \
+            /usr/lib/jvm \
+            /usr/share/dotnet \
+            /usr/share/swift \
+            /usr/local/.ghcup \
+            /usr/local/.rustup \
+            /usr/local/julia* \
+            /usr/local/lib/android \
+            /usr/local/share/powershell \
+            /opt/hostedtoolcache
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,6 +23,7 @@ repos:
           .*\.Rds|
           .*\.Rproj|
           .*\.sh|
+          (.*/|)\.gitattributes|
           (.*/|)\.gitignore|
           (.*/|)\.gitlab-ci\.yml|
           (.*/|)\.lintr|
@@ -34,6 +35,8 @@ repos:
           (.*/|)appveyor\.yml|
           (.*/|)DESCRIPTION|
           (.*/|)NAMESPACE|
+          (.*/|)pixi\.toml|
+          (.*/|)pixi\.lock|
           (.*/|)renv/settings\.dcf|
           (.*/|)renv.*\.lock|
           (.*/|)requirements.*\.in|
@@ -48,7 +51,7 @@ repos:
       - id: no-browser-statement
       - id: no-debug-statement
       - id: deps-in-desc
-        exclude: docker/.*
+        exclude: docker/.*|renv/.*
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.6.0
     hooks:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 ##########################
 # Build the slim image target first --------------------------------------------
-FROM bioconductor/r-ver:3.19 AS slim
+FROM ghcr.io/bioconductor/r-ver:3.19 AS slim
 LABEL maintainer="ccdl@alexslemonade.org"
 LABEL org.opencontainers.image.title="scpcatools-slim"
 LABEL org.opencontainers.image.source="https://github.com/AlexsLemonade/scpcaTools"
--- a/docker/.gitattributes
+++ b/docker/.gitattributes
@@ -1,0 +1,2 @@
+# SCM syntax highlighting & preventing 3-way merges
+pixi.lock merge=binary linguist-language=YAML linguist-generated=true
--- a/docker/.gitignore
+++ b/docker/.gitignore
@@ -1,0 +1,3 @@
+# pixi environments
+.pixi/*
+!.pixi/config.toml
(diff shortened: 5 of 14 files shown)


R/sce_to_anndata.R
```diff
@@ -60,10 +60,35 @@ sce_to_anndata <- function(
...
+            dplyr::select(dplyr::where(\(col) !is.list(col))) |>
+            dplyr::mutate(
+              dplyr::across(dplyr::where(is.character), \(col) tidyr::replace_na(col, "NA")),
+              dplyr::across(dplyr::where(\(col) all(is.na(col))), as.logical)
+            ) |>
+            # ensure it's not a tibble; zellkonverter drops these since no python type
+            as.data.frame()
+        } else {
+          meta
+        }
```
jashapiro, line 88: Style thing, but as long as you are redefining `meta` in the `if` we can just return it here (I really don't like distant `else` statements):
```suggestion
        } 
        # return potentially modified item
        meta
```

R/sce_to_anndata.R
```diff
@@ -60,10 +60,35 @@ sce_to_anndata <- function(
...
+  # - ensure any NA values in character columns are formatted as character
+  # - ensure any
+  metadata_to_keep <- metadata_to_keep |>
+    purrr::map(
+      \(meta) {
+        if (is.data.frame(meta)) {
+          meta <- meta |>
+            dplyr::select(dplyr::where(\(col) !is.list(col))) |>
+            dplyr::mutate(
+              dplyr::across(dplyr::where(is.character), \(col) tidyr::replace_na(col, "NA")),
```
jashapiro, line 81: Do we need to do this for character vectors too?

tests/testthat/test-sce_to_anndata.R
```diff
@@ -20,7 +20,16 @@ metadata(sce) <- list(
   sample_metadata = data.frame(
     library_id = "library1",
     sample_id = "sample1"
-  )
+  ),
+  test_dataframe = tibble::tibble(
+    "char_col" = c("name1", NA),
+    "num_col" = 1:2,
+    "na_col" = c(NA, NA),
+    "list_col" = list(a = 1, b = "2")
```
jashapiro, line 28: Nothing wrong with this, but it doesn't really fit how I think of list columns in practice...
```suggestion
    "list_col" = list(list(a = 1, b = "foo"), list(a = 2, b = "bar"))
```

tests/testthat/test-sce_to_anndata.R
```diff
@@ -20,7 +20,16 @@ metadata(sce) <- list(
...
+  ),
+  # these should get removed:
+  metadata_list = list(value = "value"),
+  metadata_DataFrame = DataFrame(head(iris))
```
jashapiro, line 32: We _could_ convert `DataFrames` to `data.frames`, so long as there are not repeated rownames (which there should not be for any of our usages. In the one case that we currently have (as far as I am aware), `singler_results`, this would result in expanding the internal matrix to multiple columns, which I'm not necessarily upset by. I'm also not upset by removing it, so maybe up to you.

R/sce_to_anndata.R
```diff
@@ -60,10 +60,35 @@ sce_to_anndata <- function(
...
+    })
+
+
+  # prepare data frames for anndata conversion:
+  # - remove any list columns
+  # - ensure any NA values in character columns are formatted as character
+  # - ensure any
+  metadata_to_keep <- metadata_to_keep |>
+    purrr::map(
+      \(meta) {
```
jashapiro, line 76: Super minor, but I find calling the variable here `meta` to be confusing, because the whole thing is the metadata, and this is just an item within that list. Using `x` as you do above (consistency!) seems fine to me.
```suggestion
      \(x) {
```

jashapiro (commented):
This looks mostly fine, but I had a few style comments and a question about whether maybe just maybe we want to keep DataFrames too (I'm totally fine with not though).

R/sce_to_anndata.R
sjspielman: Yes - if there are any NAs in a character column, the conversion can't handle multiple types and just fully fails.
Possibly we want this kind of code later in this function where we make sure coldata/rowdata are ok, too.

jashapiro: Sorry, I meant any vectors in the metadata list itself. I don't see any place where this will happen, but it could. 

I am not concerned about `coldata` and `rowdata`; I expect that zellkonverter is handling this internally.

tests/testthat/test-sce_to_anndata.R
sjspielman: >  To implement this, we should check if each DataFrame being expanded results in duplicate names, and don't include it if so.

Nope! Turns out `as.data.frame()` is running `make.unique()` under the hood, so this just works. I placed it as the first step in wrangling so if it got expanded to something problematic, that should get removed/fixed.

R/sce_to_anndata.R
sjspielman: Character vectors with `NA` work fine; this conversion "bug" appears to just affect data frames

sjspielman:
Building failed again, but something is very funky since there don't appear to be downloadable logs from it and the actions page looks like something is hanging? So, could be runner bump time.

PR #323, opened 2025-10-06
Update renv version

jashapiro:
This is a draft to test updating to the latest renv version to see if it affects builds at all, related to https://github.com/AlexsLemonade/scpcaTools/pull/322#issuecomment-3335026763

I don't really expect it to, but this may be something we want to do anyway. 

The other change is to make _all_ CRAN packages report `CRAN` as the source, which prevents a bunch of misguided queries to bioconductor repos for packages that are not housed there.

### DIFF
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -21,14 +21,9 @@ jobs:
       contents: read
       packages: write
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v4
-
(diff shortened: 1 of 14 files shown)


jashapiro:
I fixed the build errors by removing a bunch of unnecessary data from the runner in the GitHub action, so that should save us effort for a while!

sjspielman (approved):
thanks for doing this! 🚀

PR #335, opened 2026-03-31
Update release issue template 

sjspielman:
Before cutting a release here (v0.4.5), I went ahead and updated text in the release issue to match our updated infrastructure.

Since the workflow has been heavily tested with `edge` containers from this repo, I think this is really all we need to do before releasing (aka - please register your disagreements now!)


allyhawkins (approved):
LGTM, onward!

allyhawkins:
You will need to update the description though! So one more thing before actually releasing.

sjspielman:
> You will need to update the description though! So one more thing before actually releasing.

Noting I tossed this in here.

PR #336, opened 2026-04-02
Add separate anndata test file

sjspielman:
Turns out we may have a slight bug in the test suite: https://github.com/AlexsLemonade/scpcaTools/actions/runs/23902390270/job/69710478784#step:7:308

I'm fixing this with a new file name, which should get around this error which is intermittent I think because of caching. I'll delete and re-do the 0.4.5 release to include this, assuming it indeed fixes things!


allyhawkins (approved):
LGTM

Issue #337, opened 2026-04-02
Fix this snapshot bug and then fix the 0.4.5 release


sjspielman:
I had released 0.4.5 but then got a very surprising R CMD CHECK failure related to anndata snapshot checks which I tried to address here: https://github.com/AlexsLemonade/scpcaTools/pull/336

However, this did not do the trick, and it seems like this is still related to caching somehow but differently. https://github.com/AlexsLemonade/scpcaTools/actions/runs/23918657040

> 2026-04-02T20:20:48.6879068Z ══ Failed tests ════════════════════════════════════════════════════════════════
> 2026-04-02T20:20:48.6879737Z ── Failure ('test-sce_to_anndata.R:41:3'): Conversion of SCE to AnnData works as expected ──
> 2026-04-02T20:20:48.6880350Z Adding new file snapshot: 'tests/testthat/_snaps/sce_to_anndata/anndata.h5'
> 2026-04-02T20:20:48.6881021Z ── Failure ('test-sce_to_anndata.R:76:3'): Conversion of SCE to AnnData works as expected ──
> 2026-04-02T20:20:48.6881618Z Adding new file snapshot: 'tests/testthat/_snaps/sce_to_anndata/new_anndata.h5'
> 2026-04-02T20:20:48.6882580Z ── Failure ('test-sce_to_anndata.R:98:3'): Conversion of SCE to AnnData works with additional arguments ──
> 2026-04-02T20:20:48.6883267Z Adding new file snapshot: 'tests/testthat/_snaps/sce_to_anndata/no_verbose_anndata.h5'
> 2026-04-02T20:20:48.6883585Z 
> 2026-04-02T20:20:48.6883909Z ── Snapshots ───────────────────────────────────────────────────────────────────
> 2026-04-02T20:20:48.6884299Z To review and process snapshots locally:
> 2026-04-02T20:20:48.6886201Z * Run `testthat::snapshot_download_gh("AlexsLemonade/scpcaTools", "23918657040")` to download snapshots.
> 2026-04-02T20:20:48.6886838Z * Run `testthat::snapshot_accept()` to accept all changes.
> 2026-04-02T20:20:48.6887438Z * Run `testthat::snapshot_review()` to review
> 

I'm sort of surprised at this point that these snapshot tests ever passed in CI, since we don't store them in `_snaps` and the errors seem like it wants some interactivity. 

[...]

PR #339, opened 2026-04-09
Add snapshot files

jashapiro:
I got an error with tests failing when I merged #54, and it seems like for some reason we are not including snapshots in the output files. Seems like maybe the version of `testthat` we have in the renv was not throwing an error when the snapshot didn't exist and new versions do?

Either way, the snapshots are small enough that it seems like we should keep them in the repo, as it will keep us all honest (If we run tests locally as we should).


sjspielman:
> Seems like maybe the version of testthat we have in the renv was not throwing an error when the snapshot didn't exist and new versions do?

Yup! https://github.com/AlexsLemonade/scpcaTools/issues/337#issuecomment-4183719676

sjspielman (approved):
🎉

